PR #55, opened 2019-09-10
TrackletEngine update

ksung25:
TrackletEngine class is templated to be able to handle processing of any seeding and phi region. Complete set of cut tables are added to emData/TE directory. The module passes cosimulation.


pwittich:
I was not expecting to see 1600 files changed?

pwittich:
right now I think the emData/TE has very few files in it, maybe we don't want to have all the data files in github?

ksung25:
Shall I make a new pull request with all the emData/TE files removed (except the one that works with the test bench)? I can host these data files on some public space for now.

ksung25:
Data files removed and pushed to this PR

PR #59, opened 2020-01-09
Add integration test

djcranshaw:


TestBenches/FileReadUtility.hh
```diff
@@ -138,7 +138,7 @@ unsigned int compareMemWithFile(const MemType& memory, std::ifstream& fout,
   for (int i = 0; i < memory_ref.getEntries(ievt); ++i) {
 
     // Maximum processing steps per event is kMaxProc
-    if (i >= kMaxProc) {
+    if (i >= kMaxProc-10) {
```
pwittich: What is this magic number?

IntegrationTests/PR_Test/sourceFiles/Memory_PR_Test.v
```diff
@@ -0,0 +1,148 @@
+`timescale 1ns / 1ps
+//////////////////////////////////////////////////////////////////////////////////
```
pwittich: Why is there another memory file here? Should this not just be the regular memory file ?

pwittich:
@djcranshaw did you see my questions?

TestBenches/FileReadUtility.hh
djcranshaw: Whoops, this is a temporary hack from when I was having issues with truncation that I forgot to fix. I'll fix this now.

djcranshaw:
@pwittich Yep I got them yesterday. I'm working on them now.

IntegrationTests/PR_Test/sourceFiles/Memory_PR_Test.v
djcranshaw: I went back and forth in my mind about this when putting this new file in. My worry about pointing directly to the main Memory.v file in firmware-hls/TrackletAlgorithm was that any updates to Memory.v could break any number of existing standalone tests in the IntegrationTests folder. And if a particular test wasn't used for a while it might be hard to know which version of Memory.v in the git history to use. The downside is that you have multiple versions of Memory.v floating around, which I sought to mitigate by changing the name to Memory_PR_Test.v. But I'm not convinced that pointing directly to the main Memory.v file in the TrackletAlgorithm folder is the wrong thing to do, and that could easily be changed. Especially since I don't really anticipate lots of updates to Memory.v (although an important one is coming up in the next week or two from Andrew I think). What do you think?

djcranshaw:
About questions 1 & 2
the PR directory, which has the zip, I think contains (once unzipped) the files in the correct directory structure that Vivado needs to add the HLS-generated IP to the Vivado IP catalog. The PR_0 directory contains the xci file which is what Vivado generates when you make an instance of the hls-generated PR IP block (which is now in the IP catalog). If you try to delete the xci file, Vivado can't find a projection router module, which it expects to find based on the contents of the top-level verilog, which calls it. If you try to delete the PR directory and zip file, Vivado complains that the PR it has is 'locked', and the reason that it's locked is that Vivado can't find the IP definition in its catalog. It also won't let you run the simulation until it is unlocked. So at least with the generate_project.tcl in the state that it's in now, both of these files are needed.

IntegrationTests/PR_Test/sourceFiles/Memory_PR_Test.v
pwittich: This is the point of integration tests - finding out when things break ... we should not have a proliferation of different nearly identical codes. Eventually we should have unit tests that run automatically on a PR that will identify such problems and then the PR will be rejected until it's resolved.

pwittich:
> About questions 1 & 2
> the PR directory, which has the zip, I think contains (once unzipped) the files in the correct directory structure that Vivado needs to add the HLS-generated IP to the Vivado IP catalog. The PR_0 directory contains the xci file which is what Vivado generates when you make an instance of the hls-generated PR IP block (which is now in the IP catalog). If you try to delete the xci file, Vivado can't find a projection router module, which it expects to find based on the contents of the top-level verilog, which calls it. If you try to delete the PR directory and zip file, Vivado complains that the PR it has is 'locked', and the reason that it's locked is that Vivado can't find the IP definition in its catalog. It also won't let you run the simulation until it is unlocked. So at least with the generate_project.tcl in the state that it's in now, both of these files are needed.

Hmm we should think about this. I don't like binary blobs in the repo, nor files that depend too heavily on the internals of Vivado...

IntegrationTests/PR_Test/sourceFiles/Memory_PR_Test.v
djcranshaw: Okay sounds good, I'll make this change as well.

TestBenches/FileReadUtility.hh
djcranshaw: When I look at this file in the AddIntegrationTest branch and the master branch, this hack isn't present. Also Github says that the file from this comment is Outdated. So I think this is already fixed in the AddIntegrationTest branch.

djcranshaw:
> ```
> IntegrationTests/PR_Test/sourceFiles/SectorProcessor_mod.v
> IntegrationTests/PR_Test/sourceFiles/SectorProcessor_test_mod.v
> ```
> 
> Shall we change these files as follows:
> 
> 1. the top-level file is the testfixture
> 2. in the test fixture the PR should be called `uut` (unit under test)

Just to make sure that I understand. Are you suggesting ... 
- renaming the top-level function from SectorProcessor to uut
- renaming the file that declares the top-level module from SectorProcessor_mod.v to uut.v
- renaming the test bench from SectorProcessor_test to testfixture
- renaming the file that declares the test bench module from SectorProcessor_test_mod.v to testfixture.v

pwittich:
> Just to make sure that I understand. Are you suggesting ...
> * renaming the top-level function from SectorProcessor to uut
> * renaming the file that declares the top-level module from SectorProcessor_mod.v to uut.v
> * renaming the test bench from SectorProcessor_test to testfixture
> * renaming the file that declares the test bench module from SectorProcessor_test_mod.v to testfixture.v

I think that is right. If I understand correctly this is intended to be the basis for more general integration tests, and that way it is clear what's being tested (what is the uut) and what is the test bench or test fixture. What do you think, does that make sense?

TestBenches/FileReadUtility.hh
pwittich: ok, great. Not sure why it appeared in the diff earlier but maybe I did something odd.

djcranshaw:
> > Just to make sure that I understand. Are you suggesting ...
> > 
> > * renaming the top-level function from SectorProcessor to uut
> > * renaming the file that declares the top-level module from SectorProcessor_mod.v to uut.v
> > * renaming the test bench from SectorProcessor_test to testfixture
> > * renaming the file that declares the test bench module from SectorProcessor_test_mod.v to testfixture.v
> 
> I think that is right. If I understand correctly this is intended to be the basis for more general integration tests, and that way it is clear what's being tested (what is the uut) and what is the test bench or test fixture. What do you think, does that make sense?

Yes this makes sense to me, and yes I agree whatever structure we set here will I hope be the basis for the other integration tests as they're developed. The only remaining question is what the name of the particular instance of the uut that gets declared in the text fixture should be. But I think uut_inst is fine, unless you have another suggestion.

pwittich:
> Yes this makes sense to me, and yes I agree whatever structure we set here will I hope be the basis for the other integration tests as they're developed. The only remaining question is what the name of the particular instance of the uut that gets declared in the text fixture should be. But I think uut_inst is fine, unless you have another suggestion. Or maybe the instance name should reflect the section of the algorithm being tested. For this one, maybe PR_PR, and once we have a 3-module chain going, something like PR_MC, or something like that.

`uut` is an easy way to find the code; let's keep that name

PR #61, opened 2020-02-04
Update ProjectionRouter and MemoryTemplate.hh to work with HLS >= 2019.1

djcranshaw:


TrackletAlgorithm/MemoryTemplate.hh
```diff
@@ -54,7 +54,7 @@ public:
 	return nentries_[bx];
   }
 
-  DataType* get_mem() {return dataarray_;}
+  const DataType (&get_mem() const)[1<<NBIT_BX][1<<NBIT_ADDR] {return dataarray_;}
```
pwittich: was the change to a reference required or just the addition of the size of the arrays?

TrackletAlgorithm/ProjectionRouterTop.cpp
```diff
@@ -23,14 +23,14 @@ void ProjectionRouterTop(BXType bx,
...
  #pragma HLS interface register port=bx_o
- #pragma HLS resource variable=proj1in latency=2
- #pragma HLS resource variable=proj2in latency=2
- #pragma HLS resource variable=proj3in latency=2
- #pragma HLS resource variable=proj4in latency=2
- #pragma HLS resource variable=proj5in latency=2
- #pragma HLS resource variable=proj6in latency=2
- #pragma HLS resource variable=proj7in latency=2
- #pragma HLS resource variable=proj8in latency=2
+ #pragma HLS resource variable=proj1in->get_mem() latency=2
```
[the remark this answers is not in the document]
djcranshaw: It does, but the pragma wants a variable (lvalue) to apply to, which is why the get_mem method is changed from a pointer to a reference.

TrackletAlgorithm/MemoryTemplate.hh
djcranshaw: It's needed to allow the latency pragma to operate on the return value of the method. The pragma wants an lvalue, while a pointer is an rvalue.

PR #63, opened 2020-02-21
Input router modification

sseifeln:
I've modified the input router algorithm to use the modified memory template; it still doesn't do anything with the stubs though!


pwittich:
It looks like the entire input router was commented out in this PR. Is there a good reason to submit this now rather than waiting until the IR actually does something?

sseifeln:
Nope .. especially since I just noticed that this was a total git fail (the changes I *did* make are not in the commit). I'm going to close the request and try again later - when it works and when I actually commit my changes! Sorry about that.

pwittich:
you can just add new commits to the PR when you have them -- no need to close this PR!
thanks

pwittich:
superseded by #72

PR #65, opened 2020-03-05
Provide common settings for HLS and Vivado compilation

tomalin:
1) project/settings_hls.tcl provides recommended HLS compilation settings
2) project/script_*.tcl all changed, so all now have common style and all run settings_hls.tcl
3) IntegrationTests/compile_hdl.tcl does Vivado synthesis & implementation using compilation settings that optimise high frequency performance.

### DIFF
--- a/project/script_IR.tcl
+++ b/project/script_IR.tcl
@@ -1,20 +1,31 @@
-############################################################
-## This file is generated automatically by Vivado HLS.
-## Please DO NOT edit it.
-## Copyright (C) 1986-2017 Xilinx, Inc. All Rights Reserved.
-############################################################
-open_project -reset inputrouter
-set_top InputRouter
-add_files ../TrackletAlgorithm/InputRouter.cpp -cflags "-std=c++11"
-add_files -tb ../TestBenches/InputRouter_test.cpp -cflags "-I../TrackletAlgorithm -std=c++11"
-add_files -tb ../emData/IL
-open_solution -reset "solution1"
-source set_fpga.tcl
-create_clock -period 4 -name default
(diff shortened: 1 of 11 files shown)


project/script_IR.tcl
```diff
@@ -1,20 +1,31 @@
...
+add_files -tb ../emData/IL/
+
+# solutions
+open_solution "solution1"
+source settings_hls.tcl
 source set_fpga.tcl
-create_clock -period 4 -name default
-#source "./projrouter/solution1/directives.tcl"
-csim_design -compiler gcc
+create_clock -period 250MHz -name default
```
pwittich: Shall we put the clock frequency into one of the global TCL files too to ensure all projects use the same one?

tomalin: As we explore running at higher frequencies, it is possible that one or two of the projects will successfully run at higher frequencies, before all projects achieve this. That is the only argument I see against setting the clock freq. in the common file. I don't mind either way.

pwittich: I can imagine two clock speeds at most, maybe we have a common slow clock and a common fast clock?

project/script_IR.tcl
```diff
@@ -1,20 +1,31 @@
...
+
+csim_design -compiler gcc -mflags "-j8"
```
pwittich: is -j8 passed to make or to the compiler? If the latter then maybe it should be set to something like the output of the `nproc` command.

tomalin:
I moved the setting of the clock frequency (and also of the FPGA type) into settings_hls.tcl, as discussed in HLS meeting, and removed the corresponding lines from the individual algo .tcl scripts.

PR #67, opened 2020-03-11
Small modification to get PR working again after tcl synchronization

djcranshaw:


pwittich:
@tomalin can we check if others are affected too

djcranshaw:
> @tomalin can we check if others are affected too

The status of the modules on the master are:
IR: Hasn't yet been updated to use the new write_mem function in MemoryTemplate.hh, so not yet working. I think Sarah is working on this.
VMR: Hasn't yet been updated to use the new write_mem function in MemoryTemplate.hh, so not yet working. It looks like Mei has a version working on her fork; I'll email her about getting this change into master.
TE: Version in master not working because it's referencing a file that doesn't exist in the repo. I've emailed Kevin to get this updated.
TC: Hasn't yet been updated to use the new write_mem function in MemoryTemplate.hh, so not yet working. I've emailed Andrew to get this updated.
PR: Will work once this PR is approved
ME: Working
MC: Working

So I think this is the only tcl that needs this change.

tomalin:
Thanks. I think the PR script was the only one affected. Though as several don't run properly in the master, it's hard to be 100% sure the other scripts are all bug free.

PR #75, opened 2020-03-30
Feat actions

RobertGlein:
Add TE to CI

### DIFF
--- a/.github/workflows/hls.yml
+++ b/.github/workflows/hls.yml
@@ -55,9 +55,15 @@ jobs: # A workflow run is made up of one or more jobs that can run sequentially 
       with:
         name: TC
         path: project/trackletCalculator/solution1/syn/report/TrackletCalculator_L1L2G_csynth.rpt
-# Does not work right now
-#    - name: TE # Script name
-#      run: |
-#        cd project
-#        source $VIVADO_PATH_SH
-#        vivado_hls -f "script_TE.tcl"
+    - name: TE # Script name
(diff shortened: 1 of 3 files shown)


.github/workflows/hls.yml
```diff
@@ -55,9 +55,15 @@ jobs: # A workflow run is made up of one or more jobs that can run sequentially
...
-# Does not work right now
-#    - name: TE # Script name
-#      run: |
-#        cd project
-#        source $VIVADO_PATH_SH
-#        vivado_hls -f "script_TE.tcl"
+    - name: TE # Script name
+      run: |
+        cd emData/TE
+        ./download.sh
```
pwittich: do we need this? I think the goal is to have the repo and CI use the few data files that are already in the repo.

pwittich:
also it looks like the MC test fails at line 500 but the overall test passed?

.github/workflows/hls.yml
RobertGlein: The TE HLS script does not work without it.
@aehart Can we do the TE without the download script?

RobertGlein:
> also it looks like the MC test fails at line 500 but the overall test passed?

This is how the MC test was designed.

pwittich:
> > also it looks like the MC test fails at line 500 but the overall test passed?
> 
> This is how the MC test was designed.

hmm, we need to fix this ... do any of the tests fail on invalid matches?

PR #83, opened 2020-04-08
Fw synch update

djcranshaw:


pwittich:
@RobertGlein can we run the CI on this PR

pwittich:
@ksung25 @aehart @meisonlikesicecream please take a look at the changes to your modules

pwittich:
@djcranshaw is this the minimal set of data files?

meisonlikesicecream:
> @ksung25 @aehart @meisonlikesicecream please take a look at the changes to your modules

VMR is fine

TestBenches/VMRouter_test.cpp
```diff
@@ -7,7 +7,7 @@
 #include "FileReadUtility.hh"
 #include "Constants.hh"
 
-const int nevents = 2;  //number of events to run
+const int nevents = 50;  //number of events to run
```
aehart, line 10: Is there any reason not to run over all 100 events in the memory files? Especially for CI, I think it's best to use all the events we have.

TrackletAlgorithm/MatchCalculatorTop.cpp
```diff
@@ -18,7 +18,8 @@ void MatchCalculatorTop(BXType bx,
                         FullMatchMemory<BARREL>* fullmatch4,
                         FullMatchMemory<BARREL>* fullmatch5,
                         FullMatchMemory<BARREL>* fullmatch6,
-                        FullMatchMemory<BARREL>* fullmatch7
+                        FullMatchMemory<BARREL>* fullmatch7,
+                        FullMatchMemory<BARREL>* fullmatch8
                         )
 {
  #pragma HLS inline off
```
aehart, line 25: This hasn't changed, but I'm wondering why inlining is turned off here. I believe we found that in these wrapper functions, HLS wastes a clock cycle with the extra function call unless inlining is turned on.

TrackletAlgorithm/VMRouterTop.cc
```diff
@@ -27,12 +27,12 @@ void VMRouterTop(BXType bx,
 // Variables that are specified with regards to the test TestBenches
```
aehart, line 27: Same comment as in MatchCalculatorTop.cpp: I wonder if we should use the inline pragma here to save a clock cycle otherwise wasted with the extra function call.

aehart:
TC is fine, since there are no changes.

I left a couple minor comments at specific places, but I also think we should take the opportunity to standardize the extensions of at least the files that have already been modified in this PR, with an eye toward eventual CMSSW integration (and our HLS style guide):
- .h for all header files
- .cc for all C++ source files in TrackletAlgorithm/
- .cpp for all C++ test benches

djcranshaw:
> @djcranshaw is this the minimal set of data files?

In some cases no, these files just replace those that were already in the repo. Should we use this PR to remove all files except those needed by the example block that is run by default in the tcl? I think that would be the cleanest thing to do.

On a related note, the download.sh scripts that exist for some of the processing blocks will download test vectors that are now out of date. I'm not sure what to do about this. Any suggestions?

djcranshaw:
> TC is fine, since there are no changes.
> 
> I left a couple minor comments at specific places, but I also think we should take the opportunity to standardize the extensions of at least the files that have already been modified in this PR, with an eye toward eventual CMSSW integration (and our HLS style guide):
> 
> * .h for all header files
> * .cc for all C++ source files in TrackletAlgorithm/
> * .cpp for all C++ test benches

Okay I can make these changes. Might as well do them for all files now.

pwittich:
> > @djcranshaw is this the minimal set of data files?
> 
> In some cases no, these files just replace those that were already in the repo. Should we use this PR to remove all files except those needed by the example block that is run by default in the tcl? I think that would be the cleanest thing to do.
> 
> On a related note, the download.sh scripts that exist for some of the processing blocks will download test vectors that are now out of date. I'm not sure what to do about this. Any suggestions?

We should store all the test vectors in a central area (do we already do that?) and then update the download scripts. Removing obsolete test vectors also seems like a good idea.

TrackletAlgorithm/MatchCalculatorTop.cpp
djcranshaw: I found the same thing with projection router. We should explicitly inline the call to the templated function in the top-level function, otherwise we waste a clock. I'll make these changes.

TestBenches/VMRouter_test.cpp
djcranshaw: I agree, except at the moment, ME does not pass for all 100 events. Alexx is working on a fix. But I will update nevents for the other blocks in the meantime.

TrackletAlgorithm/MatchCalculatorTop.cpp
djcranshaw: Actually I misspoke (or miscommented). Inlining the Top-level wrapper function has no effect because the inline pragma applies upward to functions that call it (which is none, in the top-level). The inline pragma I think needs to be in the templated generic function (although I think by default they are inlined anyways). This is my experience, @aehart do you remember if that's what you saw as well?

aehart: Oh right. I think I had that backwards because I use the `recursive` option in the TC in order to satisfy pipelining requirements and avoid any latency from the function hierarchy there. So you're right that the pragma should be in the generic template function, and in this example, I guess turning inlining off has no effect here.

I don't remember seeing the inlining happen by default, but it should be easy to check if it happens without the pragma. Even if it does though, it might be worth adding the pragma just to be explicit.

djcranshaw: Agreed. I'll add the pragmas and push momentarily

djcranshaw:
@aehart So the reason that the PR is failing the CI tests is that the TC in this branch (and in master) relies on the existence of files in emData/TC/tables (for example, TC_L3L4_drinv.tab) which don't exist unless you run the download.sh in emData/TC/. This is very similar to what's happening with TE. We could reconfigure TC to not require these files but since we haven't really agreed on how best to do that, we could as a short-term kludge just do the same thing as is being done with the TE and run download.sh inside the TE tests (I don't exactly know how that's done but I know Robert does). What's confusing me now though is how this situation made it to the master branch in the first place. Shouldn't the CI tests have failed when you tried to pull in the new TrackletCalculator_calculate_LXLY.h? Am I missing something?

aehart:
I don't think that's why it failed. Indeed, the download script is now run in the test for the TC:
https://github.com/cms-tracklet/firmware-hls/pull/82

Looking at the output from the test, it looks like it failed because it tried to include `StubPairMemory.hh`, which is now `StubPairMemory.h`. I just fixed this, so let's see if that makes the tests happy.

djcranshaw:
> I don't think that's why it failed. Indeed, the download script is now run in the test for the TC:
> #82
> 
> Looking at the output from the test, it looks like it failed because it tried to include `StubPairMemory.hh`, which is now `StubPairMemory.h`. I just fixed this, so let's see if that makes the tests happy.

Oh right, my mistake. Thanks for that, looks like everything is passing now.

pwittich:
> > I don't think that's why it failed. Indeed, the download script is now run in the test for the TC:
> > #82
> > Looking at the output from the test, it looks like it failed because it tried to include `StubPairMemory.hh`, which is now `StubPairMemory.h`. I just fixed this, so let's see if that makes the tests happy.
> 
> Oh right, my mistake. Thanks for that, looks like everything is passing now.

shall we still wait for @ksung25 or @KristianHahn to check their files?

ksung25:
The TE changes look fine to me. However, reviewing the stub memory files reminded me of something I keep forgetting to ask:
https://github.com/cms-tracklet/firmware-hls/blob/2ae32808dd376b3aabc7cc05a0bf63d23d809bd4/TrackletAlgorithm/VMStubTEInnerMemory.h#L158
https://github.com/cms-tracklet/firmware-hls/blob/2ae32808dd376b3aabc7cc05a0bf63d23d809bd4/TrackletAlgorithm/VMStubTEOuterMemory.h#L143
For number bits allocated for bx, there is 1 bit for inner stubs but 2 bits for outer stubs. Is it intended to be different?

djcranshaw:
> The TE changes look fine to me. However, reviewing the stub memory files reminded me of something I keep forgetting to ask:
> https://github.com/cms-tracklet/firmware-hls/blob/2ae32808dd376b3aabc7cc05a0bf63d23d809bd4/TrackletAlgorithm/VMStubTEInnerMemory.h#L158
> 
> 
> https://github.com/cms-tracklet/firmware-hls/blob/2ae32808dd376b3aabc7cc05a0bf63d23d809bd4/TrackletAlgorithm/VMStubTEOuterMemory.h#L143
> 
> 
> For number bits allocated for bx, there is 1 bit for inner stubs but 2 bits for outer stubs. Is it intended to be different?

I don't think so. Both TEInner and TEOuter should just be 2 page memories and so should only need 1 bx bit. I'll change this.

pwittich:
@djcranshaw can you squash this commit again (and rebase if needed)

pwittich:
I think we are ready to merge once the CI finished -- @RobertGlein it seems to be stuck?

RobertGlein:
Yeah looks like the build server is not responding. I put the runner on another server for now. Should be finished soon

PR #84, opened 2020-04-14
A working version of the MatchEngine Code

aperloff:
This PR provides an update to the MatchEngine. This comes along with a significant reorganization of the ME code and some additional changes to make debugging easier. The code is fully templated, which allows us to easily configure it for different layers and and module types. The code has been tested and verified under the following conditions:

| Layer | csim | synth | cosim | NEvents |
|---|:---:|---|---|---|
| 1 | <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> | 100 |
| 2 | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | -- |
| 3 | <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> | 100 |
| 4 | <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> | <ul><li>[x] </li></ul> | 100 |
| 5 | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | -- |
| 6 | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | <ul><li>[ ] </li></ul> | -- |

<ul><li>[x] = Pass</li></ul>
<ul><li>[ ] = Untested</li></ul>

### DIFF
--- a/.github/workflows/hls.yml
+++ b/.github/workflows/hls.yml
@@ -4,7 +4,7 @@ env:
 
 on: # Controls when the action will run.
   push: # Please add your branch down here if wanted. A build will be triggerd for each push.
-    branches: [master, feat_actions, fw_synch_update] 
+    branches: [master, feat_actions, fw_synch_update, MEUpdate] 
 jobs: # A workflow run is made up of one or more jobs that can run sequentially or in parallel
   hls-build: # Job name
     runs-on: self-hosted # The type of runner that the job will run on
@@ -32,7 +32,7 @@ jobs: # A workflow run is made up of one or more jobs that can run sequentially 
       uses: actions/upload-artifact@v1
       with:
         name: ME
-        path: project/matchengine/solution1/syn/report/MatchEngineTopL3_csynth.rpt
+        path: project/matchengine/solution1/syn/report/MatchEngineTop_csynth.rpt
     - name: PR # Script name
       run: |
         cd project
--- a/TestBenches/FileReadUtility.h
+++ b/TestBenches/FileReadUtility.h
@@ -11,6 +11,7 @@
 #include <cerrno>
 #include <unistd.h>
 #include <vector>
+#include <bitset>
 
 #include "../TrackletAlgorithm/Constants.h"
 
@@ -102,17 +103,17 @@ void writeMemFromFile(MemType& memory, std::ifstream& fin, int ievt, int base=16
 }
 
 // TODO: FIXME or write a new one for binned memories
-template<class MemType, int base=16>
+template<class MemType, int InputBase=16, int OutputBase=16>
 unsigned int compareMemWithFile(const MemType& memory, std::ifstream& fout,
                                 int ievt, const std::string& label)
 {
   bool truncated = false;
   unsigned int err_count;
-  err_count = compareMemWithFile<MemType>(memory,fout,ievt,label,truncated);
+  err_count = compareMemWithFile<MemType,InputBase,OutputBase>(memory,fout,ievt,label,truncated);
   return err_count;
 }
 
-template<class MemType, int base=16>
+template<class MemType, int InputBase=16, int OutputBase=16>
 unsigned int compareMemWithFile(const MemType& memory, std::ifstream& fout,
                                 int ievt, const std::string& label,
                                 bool& truncated, int maxProc = kMaxProc)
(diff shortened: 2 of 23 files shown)


aperloff:
The reason the checks failed is because the enum which stores the CandidateMatchMemory and FullMatchMemory sizes are named differently; kCandidateMatchSize and kFullMatchSize respectively. This causes an issue with my changes to the FileReadUtility.h. Can I get your permission to either standardize the name or add an alias to both called 'kMatchSize'?

djcranshaw:
> The reason the checks failed is because the enum which stores the CandidateMatchMemory and FullMatchMemory sizes are named differently; kCandidateMatchSize and kFullMatchSize respectively. This causes an issue with my changes to the FileReadUtility.h. Can I get your permission to either standardize the name or add an alias to both called 'kMatchSize'?

I'm not sure what you mean. They correspond to different memory types right? Also, ME should never have to interact with FullMatch I don't think? Why does kFullMatchSize cause problems but others (say, kStubPairSize) don't?

aperloff:
Really the problem is that the function `compareMemWithFile` in FileReadUtility.h uses `MemType::BitWidths::kCandidateMatchSize` to get the size of the word it's trying to print. `MemType` is the type of memory passed to the function as a template parameter (i.e. `CandidateMatchMemory`). As far as I can tell, all of the memories have an enum type named BitWidths. What they don't all share is the name of the enum which stores the size of the memory. Since the name of this value is not consistent among the memory types, there isn't an easy way to program this. Is there a different/better way of getting this value that will work for all of the memories?

aehart:
Maybe the cleanest thing is to add a static method to each of the memory classes that simply returns the bit width of the memory. Then you could call `MemType::width()`, or whatever you choose to name the method.

djcranshaw:
> Maybe the cleanest thing is to add a static method to each of the memory classes that simply returns the bit width of the memory. Then you could call `MemType::width()`, or whatever you choose to name the method.

This seems fine to me. Standardizing the names of the word size across all memories would also work, except that would require all the processing blocks to change to the new name. I think we should avoid anything that implies a special relationship between different data types where one doesn't exist. Personally I like the MemType::width() solution best.

aperloff:
@djcranshaw @aehart @pwittich @rzouCERN 
The CI checks have passed. I think everything is ready for your review.

pwittich:
@aperloff can you squash this commit and rebase if necessary (I don't think it's necessary.)

rzouCERN:
@aperloff can you make the ME correctly update the bx_o signal by adding the following line between line 208 and line 209 of MatchEngine.cpp?
`if (istep==kMaxProc-1) bx_o = bx;`

aperloff:
> @aperloff can you rebase if necessary (I don't think it's necessary) and squash this PR ?

I rebased last night, so I think it should be okay.

aperloff:
@rzouCERN , why not put this between 209 and 210 and avoid the if statement all together? Simply put `bx_o = bx;`?

rzouCERN:
@aperloff we only want to update the bx_o when we are done writing out data of the current bx. The same thing is done in PR and MC (e.g. https://github.com/cms-tracklet/firmware-hls/blob/master/TrackletAlgorithm/ProjectionRouter.h#L367).

aperloff:
@rzouCERN @pwittich 
1) The commits have been squashed
2) Rui and I talked and tested both of our methods. They produce identical results. The first image show's the waveform with the conditional statement at the end of the processing loop. The second shows the waveform if the statement is outside the processing loop, at the end of the module. I like not having conditional statements, so we ended up going with the version without the conditional.

[MatchEngineWaveForm_Conditional.pdf](https://github.com/cms-tracklet/firmware-hls/files/4478149/MatchEngineWaveForm_Rui.pdf)
[MatchEngineWaveform_AtEndOfModule.pdf](https://github.com/cms-tracklet/firmware-hls/files/4478150/MatchEngineWaveform_Alexx.pdf)

rzouCERN:
Just to confirm that is indeed the conclusion of our discussion. I learned something new about HLS today thanks to Alexx.

> 2. Rui and I talked and tested both of our methods. They produce identical results. The first image show's the waveform with the conditional statement at the end of the processing loop. The second shows the waveform if the statement is outside the processing loop, at the end of the module. I like not having conditional statements, so we ended up going with the version without the conditional.

PR #85, opened 2020-04-15
Feat actions

RobertGlein:
Mixed CI approach using GitHub actions on a public runner, which mirrors and triggers GitLab CI on a private runner. The secret for GitLab is only known to the GitHub master repo.
Works with the public GitHub repo.
It does not work with forks.


pwittich:
Can you squash this into one commit please thanks

PR #88, opened 2020-04-21
Synchronized TC with HEAD of master branch of emulation.

aehart:
TC is now in synch with the HEAD of the master branch of the emulation. This mostly involved tweaking bit widths, and everything (C-simulation, pipelining, C/RTL cosimulation) still works.

Also removed TC files in emData/ in favor of keeping them on CERNBox. Those on CERNBox now come from the HEAD of the emulation.


TrackletAlgorithm/TrackletCalculator.cc
```diff
@@ -28,7 +28,7 @@ void TrackletCalculator_L1L2E(
 #pragma HLS array_partition variable=projout_barrel_2s complete
 #pragma HLS array_partition variable=projout_disk complete
 
-  TC_L1L2E: TrackletCalculator<TC::L1L2, TC::E, BARRELPS, BARRELPS, 2, 1, 13, 0x1FE0, 0x0000, 0x7772000, 0x7777, kMaxProc>(
```
pwittich, line 31: these magic numbers make me nervous....

TrackletAlgorithm/Constants.h
```diff
@@ -26,12 +26,12 @@ constexpr int TEBinsBits = 3;
 
 // stub digitization constants
 constexpr double kr = 0.0292969;
-constexpr double kphi = 7.8283e-06;
+constexpr double kphi = 7.71297e-06;
 constexpr double kz = 0.0585938;
 
 // tracklet digitization constants
-constexpr double krinv = 1.04377e-06;
```
pwittich, line 33: why did these change?

TrackletAlgorithm/TrackletCalculator.h
```diff
@@ -78,18 +79,18 @@ namespace TC {
   static const ap_uint<9> z0cut_L1L2 = 15.0 / kz0 + 1.0e-1;
   static const ap_uint<9> z0cut = 1.5 * 15.0 / kz0 + 1.0e-1;
   static const Types::rmean RMEAN[6] = {
-    25.137 / kr + 1.0e-1,
```
pwittich, line 81: do these belong in the constants file?

aehart: Yes, I think so. I've moved them to Constants.h.

TrackletAlgorithm/Constants.h
aehart: I don't actually know. This changed in the emulation, so maybe @djcranshaw or @aryd know?

TrackletAlgorithm/TrackletCalculator.cc
aehart: These numbers come directly from the wiring, and I think ultimately, we want a script to generate these top functions automatically from the wires file. I can start working on that, but beyond that, I'm not sure what to do about these numbers.

pwittich: should they at least become named constants so people have a hope of understanding them?

aehart: OK, I moved the magic numbers to a new file called `TrackletCalculator_parameters.h`, which, in the future, can be generated by a script from the wiring file. I decided to organize the numbers into constexpr template functions to reflect the fact that they are functions of the seed and iTC letter.

I also realized that the masks didn't need to be parameters of the top template function anymore, so I removed them.

PR #93, opened 2020-05-14
Unified tarball of updated emulation data.

aehart:
This PR includes various changes in preparation for a unified tarball of updated emulation data. The `emData/` directory now contains just two scripts:
- `download.sh`: Downloads a tarball containing all emulation data for one sector from 100 ttbar events with PU=200. For those processing modules with test benches, also copies corresponding files to dedicated directories for user convenience.
- `clean.sh`: Removes everything from `emData/` except `download.sh` and `clean.sh`.
The `download.sh` script is also executed automatically whenever any of the Tcl scripts in `project/` are run.

A few obsolete files associated with the MatchEngine were also removed, and the inclusion of the external lookup tables in the TrackletEngine test bench was simplified to reduce memory usage. A few other miscellaneous changes were also made in order to be consistent with the current HEAD of the emulation.

**N.B.**: The VMRouter currently does _not_ pass C-simulation. We are temporarily allowing the CI test for this module to fail so that this unified tarball scheme can get on its feet and we can tag the emulation. The HLS will be debugged in the near future.


pwittich:
is it obvious that the `download.sh` script should be run automatically? Is this versioned? I worry that someone might have their data files change out from under them if they need to rebuild their project.

djcranshaw:
> is it obvious that the `download.sh` script should be run automatically? Is this versioned? I worry that someone might have their data files change out from under them if they need to rebuild their project.

The download script doesn't run if MemPrints already exists in the emData directory, so if there are already data files locally they shouldn't get replaced. As far as versioning goes (@aehart correct me if I'm wrong here), the scheme I think is to periodically make a new tarball, and update the download script to point to the new tarball, and then the PR that replaces the download script will only be allowed into master if it is coincident with all the HLS changes that are needed to make the HLS work with the new data (CI ensures this). In this way we ensure that the HLS version in a commit always works for the tarball being pointed to in that commit. And then if people want to look at the older data for some reason, they need only locally change the download script to point to an older tarball (no guarantee that the newer HLS will work with the older tarball though). I think the possibility of having too many tarballs to store isn't really a problem, since they aren't that big, we don't update them that often, and eventually if we have to we can delete very old ones.

aehart:
@djcranshaw is right, regarding how we envision it working. Keeping the tarball outside of Gitlab means we lose any kind of automatic versioning. So we just have to be very deliberate when we update the tarball with data from a new emulation version, and be careful not to overwrite existing tarballs from previous versions of the emulation.

In the future, I'm imagining part of the download script will look something like
```bash
# Download and unpack the tarball.
# fw_synch_200515
#wget -O MemPrints.tar.gz --verbose "https://cernbox.cern.ch/index.php/s/CipX7CfTXIj1lcK/download"
# fw_synch_200802
#wget -O MemPrints.tar.gz --verbose "https://cernbox.cern.ch/index.php/s/AAAAAAAAAAA/download"
# fw_synch_201114
wget -O MemPrints.tar.gz --verbose "https://cernbox.cern.ch/index.php/s/BBBBBBBBBBB/download"
tar -xzvf MemPrints.tar.gz
rm -fv MemPrints.tar.gz
```
That way, people can easily refer back to the data we used at any point in the past.

pwittich:
> @djcranshaw is right, regarding how we envision it working. Keeping the tarball outside of Gitlab means we lose any kind of automatic versioning. So we just have to be very deliberate when we update the tarball with data from a new emulation version, and be careful not to overwrite existing tarballs from previous versions of the emulation.
> 
> In the future, I'm imagining part of the download script will look something like
> 
> ```shell
> # Download and unpack the tarball.
> # fw_synch_200515
> #wget -O MemPrints.tar.gz --verbose "https://cernbox.cern.ch/index.php/s/CipX7CfTXIj1lcK/download"
> # fw_synch_200802
> #wget -O MemPrints.tar.gz --verbose "https://cernbox.cern.ch/index.php/s/AAAAAAAAAAA/download"
> # fw_synch_201114
> wget -O MemPrints.tar.gz --verbose "https://cernbox.cern.ch/index.php/s/BBBBBBBBBBB/download"
> tar -xzvf MemPrints.tar.gz
> rm -fv MemPrints.tar.gz
> ```
> 
> That way, people can easily refer back to the data we used at any point in the past.

ok, hopefully that is just a schematic and the thing that will be commented out is just a variable defining the version to download, not the whole download command...

pwittich:
@aehart mysterious branch name btw 🤣

aehart:
@pwittich well, we now have one tarball to rule them all…

PR #95, opened 2020-05-15
Symbolic links in download.sh

aehart:
Instead of copying files into dedicated directories for each processing module with a test bench, `download.sh` just creates symbolic links now.

I also moved the tarball URL to a variable at the top of the script.

### DIFF
--- a/emData/download.sh
+++ b/emData/download.sh
@@ -46,7 +46,7 @@ then
 fi
 
 # Download and unpack the tarball.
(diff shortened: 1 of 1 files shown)


emData/download.sh
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 
+# fw_synch_200515
+tarball_url="https://cernbox.cern.ch/index.php/s/CipX7CfTXIj1lcK/download"
```
pwittich, line 4: looks like the tarball_url variable isn't used in the `wget`

aehart: Doh! Just fixed it.

PR #96, opened 2020-05-19
Switch to nullptr from '0' to make clang-tidy happy

djcranshaw:


pwittich:
@djcranshaw did you confirm that this change has no impact on the ultimately generated HDL code?

djcranshaw:
> @djcranshaw did you confirm that this change has no impact on the ultimately generated HDL code?

Yes, and there is no impact on the HDL.

pwittich:
@aehart can we make the `download.sh` script be less verbose? both the `wget` and the `tar` spew a lot of info into the log files which makes it hard to see the actual information from the jobs.

pwittich:
@RobertGlein there appears to be an error in the CI script but it does not lead to a failure of the job; see [this log file location](https://gitlab.cern.ch/rglein/firmware-hls_actions/-/jobs/8446351#L8801)

aehart:
@pwittich I just removed the verbose options from `download.sh` so it produces less output.

BTW, I also turned off `allow_failure` for all modules except the VMRouter, so that modules that currently pass are required to continue passing in the future.

pwittich:
> @pwittich I just removed the verbose options from `download.sh` so it produces less output.

Thanks!

aehart:
> @RobertGlein there appears to be an error in the CI script but it does not lead to a failure of the job; see [this log file location](https://gitlab.cern.ch/rglein/firmware-hls_actions/-/jobs/8446351#L8801)

I just realized that `MatchEngine.cpp` changed to `MatchEngine.cc`, but we didn't update that in `.gitlab-ci.yml`. I just fixed it, so we'll see if that makes it happy.

pwittich:
> > @RobertGlein there appears to be an error in the CI script but it does not lead to a failure of the job; see [this log file location](https://gitlab.cern.ch/rglein/firmware-hls_actions/-/jobs/8446351#L8801)
> 
> I just realized that `MatchEngine.cpp` changed to `MatchEngine.cc`, but we didn't update that in `.gitlab-ci.yml`. I just fixed it, so we'll see if that makes it happy.

ok great. Regardless the CI should fail in this case.

aehart:
> > > @RobertGlein there appears to be an error in the CI script but it does not lead to a failure of the job; see [this log file location](https://gitlab.cern.ch/rglein/firmware-hls_actions/-/jobs/8446351#L8801)
> > 
> > 
> > I just realized that `MatchEngine.cpp` changed to `MatchEngine.cc`, but we didn't update that in `.gitlab-ci.yml`. I just fixed it, so we'll see if that makes it happy.
> 
> ok great. Regardless the CI should fail in this case.

Agreed.

PR #97, opened 2020-05-20
switch to new clang-tidy version

RobertGlein:
**New clang-tidy version based on LLVM 7 to get a non-zero return value when a diagnostic (compiler) error occurs according to: https://github.com/cms-tracklet/firmware-hls/pull/96#issuecomment-630928864**

#### Documentation (different version)
https://developers.redhat.com/blog/2018/07/07/yum-install-gcc7-clang/
#### Installation
```
sudo yum install centos-release-scl-rh
sudo yum --enablerepo=centos-sclo-rh-testing install llvm-toolset-7.0-clang-analyzer llvm-toolset-7.0-clang-tools-extra
source /opt/rh/llvm-toolset-7.0/enable
```
#### Test: Should return LLVM version 7.0.1
`/opt/rh/llvm-toolset-7.0/root/bin/clang-tidy -version `

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,7 +1,7 @@
 variables:
   GIT_SUBMODULE_STRATEGY: recursive
   VIVADO_PATH_SH: '/nfs/data41/software/Xilinx/Vivado/${VIVADO_VERSION}/settings64.sh'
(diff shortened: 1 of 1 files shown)


pwittich:
Sorry this seems to have generated more work than I expected it to ....

Issue #58, opened 2019-11-20
Memory interfaces


aehart:
I sort of mentioned this in passing during the meeting today, but the changes that @djcranshaw and @bryates made to the memory interfaces got me thinking about other ports that are actually redundant now that we are plugging everything into a Verilog top module:

- `nentries` ports for output memories: We don't really need these because we can just count the number of write enables in the top module. These can be removed by removing the `nentries_` data member from the memory classes for output memories. It was easy to make this change locally, and it has the desired effect. I can submit a pull request soon if we decide this is OK.
- `nentries` ports for each page of the input memories: We can make the processing modules "BX agnostic" such that they have no BX input port and have a single `nentries` port for each input memory, instead of one for each memory page. Then it will be up to the top module to read from and write to the correct page without unnecessarily routing that information through all the processing modules. I haven't attempted this change yet.

Any thoughts?

djcranshaw:
- On the first point. Would this require a change to Memory.v? I think it has to because the write-enables and data are wired directly into the memory module. This is still probably doable, just means logic internal to Memory.v Another thing is that we need the nentries_ array for input memories, so we can't just get rid of the data member without separating input memories from output memories in the HLS. We would probably need something like MemoryTemplateInputs.hh and MemoryTemplateOutputs.hh, where the former has nentries_ and the latter doesn't. That strikes me as a bit of a nuisance but it still would work. Another possibility is that if we don't ever use the nentries_ array in output memories, HLS will just optimize those ports away. Then we could still just have the one MemoryTemplate.hh file. I haven't tried this but it this should work, since HLS successfully optimized away the analogous input ports when we stopped using them.

- On the second point. So the way it is set up right now is that the processing blocks handle the bx signals, and the memories are bx-agnostic. I like this setup because after a proc block is finished with a processing step, the last thing it does is send out the bx_signal it was working on, coincident with its done signal. Then the next block knows to start at the same moment its getting the information about which bx its working on. If we make the proc blocks bx-agnostic, we will need some verilog logic in the top module to keep track of when each module is finished, so it can switch the bx signal (which would probably have to be a register array in the top-level or something) that is feeding the output memory module so that it knows which page to write to. Also, if we only have one nentries input per proc block, we will have to dynamically switch which page in the memory the proc block is reading from, which will also require some verilog logic in the top-level. My preference would be keeping as much logic in the HLS as possible.

gershtein:
Hi Derek,

[...]

PR #57, opened 2019-11-19
Prmemc chain

djcranshaw:

PR #100, opened 2020-06-17
Vmrouter mei

meisonlikesicecream:
Works for all memory types, including copies. Added testbench for disk if someone wants to try that. There's an ugly hack in the AllStubMemory.h to not make the cosim crash for the disk, neither me nor Ian understand why this happens. The binned memories still assumes 8 bins for the disk when emulation uses 16, however, it still works fine. Don't know if this should be changed or not?

Meets timing, but still uses A LOT of resources.


aehart:
@meisonlikesicecream Great work. We've seen in the past that HLS seems to not propagate the `regionType` enum definition to all the necessary places when it generates files for cosimulation, e.g. #38. I think you're seeing the same thing, and I think your solution is as good as we can do. I just added a `static_assert` to that specialization so that if `DISK` does not equal six, we'll get a compilation error.

I think we do want to use 16 bins for the disks. Right now, the 3-bit bin value is just overflowing, e.g., when you try to write to bin 8, it actual writes to bin 0; when you try to write to bin 9, it actual writes to bin 1, etc. The reason it *appears* to work is that the reference memory is of the same class as the actual memory that you are writing to, so it is subject to the exact same overflow. And you can see in the test bench output that nothing ever seems to be in bins 8-15, even though there are entries in these bins when you look in the test vector files.

I did also try running the test bench for VMR_D1PHIA, but I get a bizarre error during cosimulation:
```
…
source VMRouterTop.tcl
## run all
WARNING: file ../tv/rtldatafile/rtl.VMRouterTop.autotvout_teoMemories_2_3_nentries_0_V_1.dat could not be opened
Failed to open file "../tv/rtldatafile/rtl.VMRouterTop.autotvout_teoMemories_2_3_nentries_0_V_1.dat"!
ERROR: Simulation using HLS TB failed.
$finish called at time : 110 ns : File "/home/hart/Research/CMS/tracklets/firmware-hls/project/vmrouter/solution1/sim/verilog/VMRouterTop.autotb.v" Line 65854
## quit
```
I think this might be an HLS bug though. Can you confirm whether you see this as well?

Also, you may have noticed that the `nentries` ports for the binned output memories look different from the unbinned output memories because they have inputs:
```
|meMemories_0_nentries_0_V_0_i            |  in |    5|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
|meMemories_0_nentries_0_V_0_o            | out |    5|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
|meMemories_0_nentries_0_V_0_o_ap_vld     | out |    1|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
…
```
versus
```
|allStub_0_nentries_0_V                   | out |    8|   ap_vld   |     allStub_0_nentries_0_V     |    pointer   |
|allStub_0_nentries_0_V_ap_vld            | out |    1|   ap_vld   |     allStub_0_nentries_0_V     |    pointer   |
…
```
This is because we never added the write index as an input argument to `MemoryTemplateBinned::write_mem` like we did for `MemoryTemplate::write_mem` (see #57). I don't think this was intentional, so I tried adding this (see the [binned_memory_nentries](https://github.com/cms-tracklet/firmware-hls/tree/binned_memory_nentries) branch) but it caused the resource usage to explode. As you say, the resource usage is already pretty high, so we need to think more about that as well as the interface, especially for the binned output memories, but I don't think this should hold up the PR.

Finally, I modified `emData/download.sh` to set up symbolic links for all the lookup tables and test vectors needed for the two test benches that you created and modified the code to use these, just to be in line with what the other modules do.

meisonlikesicecream:
> @meisonlikesicecream 
> I think we do want to use 16 bins for the disks. Right now, the 3-bit bin value is just overflowing, e.g., when you try to write to bin 8, it actual writes to bin 0; when you try to write to bin 9, it actual writes to bin 1, etc. The reason it _appears_ to work is that the reference memory is of the same class as the actual memory that you are writing to, so it is subject to the exact same overflow. And you can see in the test bench output that nothing ever seems to be in bins 8-15, even though there are entries in these bins when you look in the test vector files.

I wrote the VMR to be able to handle 16 bins for the disk (I was just fortunate enough that the overflowing of the memories wasn't causing any issues), so one just needs to change the number of bits used for the bins in the memory definitions, but I'm not quite sure how to do this without ruining it for the barrels that only use 8 bins. I.e. change the number of bits on this line for example: https://github.com/cms-tracklet/firmware-hls/blob/vmrouter_mei/TrackletAlgorithm/VMStubMEMemory.h#L132

> I did also try running the test bench for VMR_D1PHIA, but I get a bizarre error during cosimulation:
> 
> ```
> …
> source VMRouterTop.tcl
> ## run all
> WARNING: file ../tv/rtldatafile/rtl.VMRouterTop.autotvout_teoMemories_2_3_nentries_0_V_1.dat could not be opened
> Failed to open file "../tv/rtldatafile/rtl.VMRouterTop.autotvout_teoMemories_2_3_nentries_0_V_1.dat"!
> ERROR: Simulation using HLS TB failed.
> $finish called at time : 110 ns : File "/home/hart/Research/CMS/tracklets/firmware-hls/project/vmrouter/solution1/sim/verilog/VMRouterTop.autotb.v" Line 65854
> ## quit
> ```
> 
> I think this might be an HLS bug though. Can you confirm whether you see this as well?

I was running on Vivado 2019.1 until recently... and it worked fine there, but yes, I see the same issue when I try 2019.2 or 2020.1. Couldn't figure what Vivado is doing, but the file it can't open is completely empty? Is this something that can be ignored (for now) or should I try to fix it?

> Also, you may have noticed that the `nentries` ports for the binned output memories look different from the unbinned output memories because they have inputs:
> 
> ```
> |meMemories_0_nentries_0_V_0_i            |  in |    5|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
> |meMemories_0_nentries_0_V_0_o            | out |    5|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
> |meMemories_0_nentries_0_V_0_o_ap_vld     | out |    1|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
> …
> ```
> 
> versus
> 
> ```
> |allStub_0_nentries_0_V                   | out |    8|   ap_vld   |     allStub_0_nentries_0_V     |    pointer   |
> |allStub_0_nentries_0_V_ap_vld            | out |    1|   ap_vld   |     allStub_0_nentries_0_V     |    pointer   |
> …
> ```
> 
> This is because we never added the write index as an input argument to `MemoryTemplateBinned::write_mem` like we did for `MemoryTemplate::write_mem` (see #57). I don't think this was intentional, so I tried adding this (see the [binned_memory_nentries](https://github.com/cms-tracklet/firmware-hls/tree/binned_memory_nentries) branch) but it caused the resource usage to explode. As you say, the resource usage is already pretty high, so we need to think more about that as well as the interface, especially for the binned output memories, but I don't think this should hold up the PR.

I've got like 2000 input/output ports to my VMR because of the nentries arrays, and they seem to be a major reason to why I'm using so many LUTs. Are the modules supposed to have "direct contact" with what's going on inside the memories?

I also noticed that my change to using strict inequality signs in the write_mem functions didn't go into this branch, so I added that now.

aehart:
> > @meisonlikesicecream
> > I think we do want to use 16 bins for the disks. Right now, the 3-bit bin value is just overflowing, e.g., when you try to write to bin 8, it actual writes to bin 0; when you try to write to bin 9, it actual writes to bin 1, etc. The reason it _appears_ to work is that the reference memory is of the same class as the actual memory that you are writing to, so it is subject to the exact same overflow. And you can see in the test bench output that nothing ever seems to be in bins 8-15, even though there are entries in these bins when you look in the test vector files.
> 
> I wrote the VMR to be able to handle 16 bins for the disk (I was just fortunate enough that the overflowing of the memories wasn't causing any issues), so one just needs to change the number of bits used for the bins in the memory definitions, but I'm not quite sure how to do this without ruining it for the barrels that only use 8 bins. I.e. change the number of bits on this line for example: https://github.com/cms-tracklet/firmware-hls/blob/vmrouter_mei/TrackletAlgorithm/VMStubMEMemory.h#L132

Well, the internet tells me that it isn't possible to specialize a alias template, so I think we have to add an extra template parameter to VMStubMEMemory, e.g.:
```c++
template<int VMSMEType, int NBIT_BIN> using VMStubMEMemory = MemoryTemplateBinned<VMStubME<VMSMEType>, 2, kNBits_MemAddr,NBIT_BIN>;
```
Then you just have to propagate that new parameter through the VMRouter code, setting it to four when VMSMEType is DISK.

> 
> > I did also try running the test bench for VMR_D1PHIA, but I get a bizarre error during cosimulation:
> > ```
> > …
> > source VMRouterTop.tcl
> > ## run all
> > WARNING: file ../tv/rtldatafile/rtl.VMRouterTop.autotvout_teoMemories_2_3_nentries_0_V_1.dat could not be opened
> > Failed to open file "../tv/rtldatafile/rtl.VMRouterTop.autotvout_teoMemories_2_3_nentries_0_V_1.dat"!
> > ERROR: Simulation using HLS TB failed.
> > $finish called at time : 110 ns : File "/home/hart/Research/CMS/tracklets/firmware-hls/project/vmrouter/solution1/sim/verilog/VMRouterTop.autotb.v" Line 65854
> > ## quit
> > ```
> > 
> > 
> > I think this might be an HLS bug though. Can you confirm whether you see this as well?
> 
> I was running on Vivado 2019.1 until recently... and it worked fine there, but yes, I see the same issue when I try 2019.2 or 2020.1. Couldn't figure what Vivado is doing, but the file it can't open is completely empty? Is this something that can be ignored (for now) or should I try to fix it?

Yeah definitely sounds like an HLS bug. I think we can ignore for now.

> 
> > Also, you may have noticed that the `nentries` ports for the binned output memories look different from the unbinned output memories because they have inputs:
> > ```
> > |meMemories_0_nentries_0_V_0_i            |  in |    5|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
> > |meMemories_0_nentries_0_V_0_o            | out |    5|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
> > |meMemories_0_nentries_0_V_0_o_ap_vld     | out |    1|   ap_ovld  |   meMemories_0_nentries_0_V_0  |    pointer   |
> > …
> > ```
> > 
> > 
> > versus
> > ```
> > |allStub_0_nentries_0_V                   | out |    8|   ap_vld   |     allStub_0_nentries_0_V     |    pointer   |
> > |allStub_0_nentries_0_V_ap_vld            | out |    1|   ap_vld   |     allStub_0_nentries_0_V     |    pointer   |
> > …
> > ```
> > 
> > 
> > This is because we never added the write index as an input argument to `MemoryTemplateBinned::write_mem` like we did for `MemoryTemplate::write_mem` (see #57). I don't think this was intentional, so I tried adding this (see the [binned_memory_nentries](https://github.com/cms-tracklet/firmware-hls/tree/binned_memory_nentries) branch) but it caused the resource usage to explode. As you say, the resource usage is already pretty high, so we need to think more about that as well as the interface, especially for the binned output memories, but I don't think this should hold up the PR.
> 
> I've got like 2000 input/output ports to my VMR because of the nentries arrays, and they seem to be a major reason to why I'm using so many LUTs. Are the modules supposed to have "direct contact" with what's going on inside the memories?

I'm not sure what you mean by "direct contact", but I think you're right that your embarrassment of ports is what's driving up the LUT usage. We've had a discussion in the past about how the `nentries` ports aren't strictly necessary for output memories (see #58), but at the time we just weren't able to get something that worked without these ports. It may be worth investing some more time in that endeavor though, for the sake of the VMRouter.

meisonlikesicecream:
> Well, the internet tells me that it isn't possible to specialize an alias template, so I think we have to add an extra template parameter to VMStubMEMemory, e.g.:
> 
> ```c++
> template<int VMSMEType, int NBIT_BIN> using VMStubMEMemory = MemoryTemplateBinned<VMStubME<VMSMEType>, 2, kNBits_MemAddr,NBIT_BIN>;
> ```

Stupid question, does this mean that you (or someone that is not me) will add it?

aehart:
> > Well, the internet tells me that it isn't possible to specialize an alias template, so I think we have to add an extra template parameter to VMStubMEMemory, e.g.:
> > ```c++
> > template<int VMSMEType, int NBIT_BIN> using VMStubMEMemory = MemoryTemplateBinned<VMStubME<VMSMEType>, 2, kNBits_MemAddr,NBIT_BIN>;
> > ```
> 
> Stupid question, does this mean that you (or someone that is not me) will add it?

Sorry I wasn't very clear. I was thinking that you would add it. It will require propagating the new template parameter through the VMRouter code and seeing that it successfully writes to all 16 bins in the disks.

djcranshaw:
Is the plan for the disks then to use a 16-bin, 16-entry per bin page = 256 entries per bx, or a 16-bin, 8-entry per bin = 128 entries per bx? If the former, this will drive up our BRAM usage which is already rather high, and if the latter this might result in significant truncation (although I don't know if that's been studied).

aehart:
@djcranshaw At least in the HLS, all our memories are assumed to have 128 entries per BX, so for a 16-bin memory, that would be 8 entries per bin. I don't know if the effect of this number of entries per bin on truncation has been studied, or even if limitations on the sizes of memories is implemented in any way in the emulation.

djcranshaw:
> @djcranshaw At least in the HLS, all our memories are assumed to have 128 entries per BX, so for a 16-bin memory, that would be 8 entries per bin. I don't know if the effect of this number of entries per bin on truncation has been studied, or even if limitations on the sizes of memories is implemented in any way in the emulation.

In the emulator on the master branch of gitlab truncation is implement just by fixing the number of iterations that a particular proc block can take, so for unbinned memories that effectively limits the memory size to 128 (actually 108). But for binned memories that means that we could still have a 16 page memory memory with 16 slots in each. In fact, on the gitlab emulator which allows a particular bin to overflow the 16 slots in the bin, and the HLS interprets the excess as being in the subsequent bin, which causes (very occasionally) a difference between HLS and emulator. This has been fixed in the github version that's going into CMSSW so once we resync and make a new tarball that will go away, but it is the state of affairs now. 

I suspect limiting each bin to 8 will cause significant enough truncation that we probably shouldn't ignore it, but it should be studied. Maybe we can talk about it this afternoon.

meisonlikesicecream:
> > > Well, the internet tells me that it isn't possible to specialize an alias template, so I think we have to add an extra template parameter to VMStubMEMemory, e.g.:
> > > ```c++
> > > template<int VMSMEType, int NBIT_BIN> using VMStubMEMemory = MemoryTemplateBinned<VMStubME<VMSMEType>, 2, kNBits_MemAddr,NBIT_BIN>;
> > > ```
> > 
> > 
> > Stupid question, does this mean that you (or someone that is not me) will add it?
> 
> Sorry I wasn't very clear. I was thinking that you would add it. It will require propagating the new template parameter through the VMRouter code and seeing that it successfully writes to all 16 bins in the disks.

No worries, I added it now! Don't know if I was supposed to fix it for the Match Engine as well, but I did it anyway...

PR #101, opened 2020-06-24
MatchEngine memory paging and latency

aperloff:
Switch from a 4 page to an 8 page stub memory. Also change the read latency of the memory to 2.

### DIFF
--- a/TrackletAlgorithm/MatchEngine.cc
+++ b/TrackletAlgorithm/MatchEngine.cc
@@ -217,8 +217,8 @@ void MatchEngineTop(const BXType bx, BXType& bx_o,
 					CandidateMatchMemory& outputCandidateMatch) {
 
 #pragma HLS interface register port=bx_o
-//#pragma HLS resource variable=inputStubData->get_mem() latency=2
-//#pragma HLS resource variable=inputProjectionData->get_mem() latency=2
+#pragma HLS resource variable=inputStubData->get_mem() latency=2
+#pragma HLS resource variable=inputProjectionData->get_mem() latency=2
 
 	MatchEngine<LAYER,MODULETYPE,PROJECTIONTYPE>(bx, bx_o, inputStubData, inputProjectionData, outputCandidateMatch); 
 }
--- a/TrackletAlgorithm/VMStubMEMemory.h
+++ b/TrackletAlgorithm/VMStubMEMemory.h
@@ -129,6 +129,6 @@ private:
 };
 
 // Memory definition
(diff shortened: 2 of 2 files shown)


aperloff:
> Is the post-implementation latency for the new ME less than 4ns with these pragmas turned on? I checked for the old ME and it was fine but I just want to make sure

Let me check. Just need a few minutes.

aperloff:
> Is the post-implementation latency for the new ME less than 4ns with these pragmas turned on? I checked for the old ME and it was fine but I just want to make sure

The post implementation timing goes from:
```
CP required:    4.000
CP achieved post-synthesis:    2.272
CP achieved post-implementation:    2.719
```
to
```
CP required:    4.000
CP achieved post-synthesis:    2.887
CP achieved post-implementation:    3.035
```

The latency goes from 115 clk to 116 clk (according to HLS). This makes sense given that added memory read clock cycle.

The resource utilization goes from:
```
CLB:            104
LUT:            410
FF:             478
DSP:              0
BRAM:             0
SRL:              0
URAM:             0
```
to
```
CLB:            153
LUT:            686
FF:             606
DSP:              0
BRAM:             0
SRL:              0
URAM:             0
```

djcranshaw:
> > Is the post-implementation latency for the new ME less than 4ns with these pragmas turned on? I checked for the old ME and it was fine but I just want to make sure
> 
> @djcranshaw, the post implementation timing goes from:
> 
> ```
> CP required:    4.000
> CP achieved post-synthesis:    2.272
> CP achieved post-implementation:    2.719
> ```
> 
> to
> 
> ```
> CP required:    4.000
> CP achieved post-synthesis:    2.887
> CP achieved post-implementation:    3.035
> ```
> 
> The latency goes from 115 clk to 116 clk (according to HLS). This makes sense given that added memory read clock cycle.
> 
> The resource utilization goes from:
> 
> ```
> CLB:            104
> LUT:            410
> FF:             478
> DSP:              0
> BRAM:             0
> SRL:              0
> URAM:             0
> ```
> 
> to
> 
> ```
> CLB:            153
> LUT:            686
> FF:             606
> DSP:              0
> BRAM:             0
> SRL:              0
> URAM:             0
> ```

Okay great. The timing remains okay and the extra clock makes sense. Thank for checking. Not sure what to make of the slightly higher resources.

aperloff:
@djcranshaw I think that's because of the memory latency going from 1 to 2. It has to store values in memory for longer and thus takes more resources. I think the memory paging increased the resources just slightly, but not by much.

djcranshaw:
> @djcranshaw I think that's because of the memory latency going from 1 to 2. It has to store values in memory for longer and thus takes more resources. I think the memory paging increased the resources just slightly, but not by much.

Okay, makes sense.

PR #102, opened 2020-07-01
Passing memories as C arrays (match and fullmatch)

bryates:


aehart:
@bryates Thanks for making this change. It looks much cleaner with the arrays.

One thing I'm just noticing though is that you pass in eight FullMatch memories, but only write to two.. And indeed, there are only two FullMatch memories attached to MC_L3PHIC in the wiring. Is there a reason for passing in eight? I see from the C-synthesis report that the unused ports are not removed, so this seems like a bit of a waste of resources at best.

bryates:
@aehart when I inherited the code, the others were off. I left it as is for debugging, but I suppose now is a good time to enable the other 6 outputs.

djcranshaw:
> > @aehart when I inherited the code, the others were off. I left it as is for debugging, but I suppose now is a good time to enable the other 6 outputs.
> 
> I think we actually want to _disable_ the other six outputs in a way that lets HLS remove those unused ports. You might have a look at the top-level function of the TrackletCalculator, where we pass arrays of TrackletProjection memories, and then use bit mask template parameters to inform the compiler which memories are written to. Something similar should work for the MatchCalculator.
> 
> In any case, I don't think this should hold up this PR. It's just something that I happened to notice here, which should be fixed when you have a chance.

I also noticed this when switching from top-level verilog to VHDL since the latter doesn't let you simply not connect unconnected ports in an IP, the VHDL wants you to specify 'open' on unused ports. But yes there is no reason this PR can't go through now, it will just need to be fixed eventually.

PR #103, opened 2020-07-06
Pr arrays

djcranshaw:
Change to using arrays of memory classes for inputs and outputs of Projection Router rather than individual pointers. In addition to simplifying the code on the interface (the synthesized interface is unchanged except port names), the logic for reading the input memories has been fairly substantially simplified which resulted in the new PR using about 60% of the resources of the old PR (FF and LUTs, mainly).

### DIFF
--- a/TestBenches/ProjectionRouter_test.cpp
+++ b/TestBenches/ProjectionRouter_test.cpp
@@ -18,25 +18,11 @@ int main()
 
   ///////////////////////////
   // input memories
-  static TrackletProjectionMemory<BARRELPS> tproj1;
-  static TrackletProjectionMemory<BARRELPS> tproj2;
-  static TrackletProjectionMemory<BARRELPS> tproj3;
-  static TrackletProjectionMemory<BARRELPS> tproj4;
-  static TrackletProjectionMemory<BARRELPS> tproj5;
-  static TrackletProjectionMemory<BARRELPS> tproj6;
-  static TrackletProjectionMemory<BARRELPS> tproj7;
-  static TrackletProjectionMemory<BARRELPS> tproj8;
+  static TrackletProjectionMemory<BARRELPS> tprojarray[8];
 
   // output memories
   static AllProjectionMemory<BARRELPS> allproj;
-  static VMProjectionMemory<BARREL> vmproj1;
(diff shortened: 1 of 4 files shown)


pwittich:
Sounds great! can you point to the changes that reduced the resource usage so dramatically? I can't really tell from looking at the 'files changed' tab.

djcranshaw:
> Sounds great! can you point to the changes that reduced the resource usage so dramatically? I can't really tell from looking at the 'files changed' tab.

So I did some digging and found out that about half of the improvement comes from removing all of the recursive function calls in lines 14-71 of ProjectionRouter.h https://github.com/cms-tracklet/firmware-hls/blob/9d1b291bd67150495319e6139f963a8044b82c33/TrackletAlgorithm/ProjectionRouter.h#L16-L70

djcranshaw:
> Sounds great! can you point to the changes that reduced the resource usage so dramatically? I can't really tell from looking at the 'files changed' tab.

So I did some digging and found that about half of the improvement comes from removing the recursive functions calls from lines 14-71 in ProjectionRouter.h. I'm not exactly sure what causes the improvement (or rather what about those function calls causes the extra usage) but getting rid of them is what's responsible for about half of the reduction in resources. The other half comes from changing the ints in nvmprojout (lines 217-225) to ap_uint<7>. I was surprised by this because I thought HLS was suppose to figure out how big these internal ints need to be and adjust their size accordingly, but maybe it can't figure out enough about the internal logic of the PR to make that determination.

pwittich:
Ah, removing recursion is a good one. And the `ap_uint<X>` is also a good thing to know...

pwittich:
I'm confused, the commit appears as coming from someone I don't know? @aehart ?
image

aehart:
> I'm confused, the commit appears as coming from someone I don't know? @aehart ?
> image

My bad. I pushed from a shared machine where somebody decided to set up the gitconfig with their name. I just fixed it.

pwittich:
> > I'm confused, the commit appears as coming from someone I don't know? @aehart ?
> > image
> 
> My bad. I pushed from a shared machine where somebody decided to set up the gitconfig with their name. I just fixed it.

mystery solved!

PR #109, opened 2020-09-23
IR update 

sseifeln:
Updating out of date input router with latest version from my fork. 
IR now makes use of the rewind pragma, and I have added the IR to the quality + has checks in the CI. 
I need a file to be included in the tarball (dtclinklayerdisk.dat) to run the test bench - since its not there right now and I wanted to verify that the changes I made would pass the CI - I have, for now, added it to git (emData) and modified the clean-up script in that directory so as not to automatically delete it.


sseifeln:
my mistake .. wrong branch for the sync.

Issue #110, opened 2020-09-25
Eliminate unnamed hard-wired constants in code


tomalin:
CMS coding rules forbid unnamed hard-wired constants in the code. Although we're well behaved here, a couple of dozen have survived. Please eliminate these. (P.S. I've not checked IR or VMR yet, since they are soon to be updated).

## LayerRouter

•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/LayerRouter.h#L9
•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/LayerRouter.cpp#L8
•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/LayerRouter.cpp#L19
•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/LayerRouter.cpp#L21
•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/LayerRouter.cpp#L37

## Match Calculator

•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/MatchCalculator.h#L266
•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/MatchCalculator.h#L267
•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/MatchCalculator.h#L286
•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/MatchCalculatorTop.cc#L17

## Match Engine

•       https://github.com/cms-tracklet/firmware-hls/blob/937145cf43bccc84825bd5c2ea60ce087cfcae02/TrackletAlgorithm/MatchEngine.cc#L171
[...]

PR #111, opened 2020-09-29
Removed hard-wired constants for LUTs and MC template

bryates:
modified:   ../TrackletAlgorithm/MatchCalculator.h
	modified:   ../TrackletAlgorithm/MatchCalculatorTop.cc


pwittich:
Thanks, Brent. Let's get everyone's commit into this single PR to address this issue

tomalin:
@bryates Much better, though 2 hard-wired consts remain at https://github.com/cms-tracklet/firmware-hls/blob/MCconst/TrackletAlgorithm/MatchCalculator.h#L291 & L292.

bryates:
@tomalin those two are now gone.

djcranshaw:
@aehart I notice you have enums for layer, disk, and seed, defined for the TC here 
https://github.com/cms-tracklet/firmware-hls/blob/e207950e7aa0c5edd90c75e7686639a310684a93/TrackletAlgorithm/TrackletCalculator.h#L35
These would be useful elsewhere in the project (at least in e.g. Projection Router), should we move these out of the TC namespace and into the constants file? I can do it while updating the Projection Router if you agree.

aehart:
@djcranshaw Yes, I think that's a good idea.

TrackletAlgorithm/Constants.h
```diff
@@ -3,8 +3,10 @@
 
 #include "ap_int.h"
 
-//#include <cmath>
-//#include "hls_math.h"
+// Macro to convert floating point values to integers, given a digitization
+// constant. The 1.0e-1 is a fudge factor needed to get the floating point
+// truncation to agree exactly with the emulation.
+#define INT(x,k) (static_cast<int>(x / k + 1.0e-1))
```
pwittich, line 9: this should be ((x)/((k)+1...) due to the usual issue with preprocessor macros.  or better yet since we are in C++ it should be an inline function.

aehart: Switched to an inline function.

TrackletAlgorithm/Constants.h
```diff
@@ -55,27 +64,54 @@ constexpr double kphi0 = 1.54259e-05;
...
+constexpr double z0cut = 15.0; // cm
+
+// various bit widths
+constexpr unsigned nbitsallstubs[N_LAYER + N_DISK] = {3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2};
+constexpr unsigned int nbits_maxvm = 5; // number of bits needed for max number of VMs per layer/disk (max number is 32)
 
 // List of regions for memory template parameters
 enum regionType {BARRELPS, BARREL2S, BARRELOL, BARREL, DISKPS, DISK2S, DISK};
 
+namespace TF {
```
djcranshaw, line 104: @aehart I left these enums in a TF namespace to avoid confusion with more generic uses of seed, layer, disk. An alternative might be to change these enum names to something less prone to be accidentally reused. Also, there are now a slew of trivial changes to TC which change many TC:: to TF::

aehart: I think this is fine for now. Our use of namespaces right now is a little ad-hoc and inconsistent, so at some point, we'll need to overhaul it. But we'll address that in a future issue.

tomalin:
Almost done! But none of the hard-wired constants of the ME, mentioned in https://github.com/cms-tracklet/firmware-hls/issues/110 have so far been eliminated. (The same is also true of MemoryBinned.v)..

aehart:
@tomalin I asked @aperloff if he could take a look at the ME, and he has promised to do so soon. And I will take a look at MemoryBinned.v today.

pwittich:
> @tomalin I asked @aperloff if he could take a look at the ME, and he has promised to do so soon. And I will take a look at MemoryBinned.v today.

@aehart @aperloff what's the status of this?

aehart:
@aperloff Have you had a chance to look at the ME?

aehart:
@tomalin I believe all the areas you flagged in issue #110 have now been addressed.

Regarding MemoryBinned.v, I added some comments to that file, but it will also likely require significant changes after the nentries ports are removed from the output memories. Since this is one of the PRs in our backlog, I would suggest delaying a more critical review of MemoryBinned.v until that time.

tomalin (commented):
Looking much better, though could we remove the few unnamed constants that remain?
1) 7 at https://github.com/cms-tracklet/firmware-hls/blob/MCconst/TrackletAlgorithm/MatchCalculator.h#L258
2) 10 at https://github.com/cms-tracklet/firmware-hls/blob/MCconst/TrackletAlgorithm/MatchCalculator.h#L259
3) 5 at https://github.com/cms-tracklet/firmware-hls/blob/MCconst/TrackletAlgorithm/MatchEngine.h#L90
4) 8 & 3 in https://github.com/cms-tracklet/firmware-hls/blob/MCconst/TrackletAlgorithm/ProjectionRouterTop.cc

Ian

bryates:
@tomalin I'll look into the three you flagged in the MC. These were put in place before I started working on, so I'll have to track down their meaning.

aperloff:
@tomalin I fixed the one remaining in `MatchEngine.h`.

tomalin:
I've removed the few in ProjectionRouterTop, to get this finished.

tomalin (approved):
All done

Issue #108, opened 2020-09-21
Split emData/download.sh


tomalin:
emData/download.sh should be split into two scripts, one to download the LUT data & one to download to the memory data input/output from each algo step. (Motivation: CMSSW when calling the HLS algos will only need the LUT data).

aehart:
@tomalin I've split the tarball in emData/download.sh into MemPrints.tar.gz (test vectors) and LUTs.tar.gz (lookup tables and other small miscellaneous files). I've also added an option for downloading and unpacking only the latter, which should be what's needed for the emulation.

These changes have been made to the fw_synch branch, but since the synchronization is taking some time to complete, I also opened a PR that applies just these changes to the master branch: #112.

PR #112, opened 2020-09-29
Download separate MemPrints and LUTs tarballs

aehart:
This PR splits the single tarball into two:
- MemPrints.tar.gz contains the test vectors that are the inputs and outputs used by the HLS test benches
- LUTs.tar.gz contains lookup tables, as well as other small files copied from L1Trigger/TrackFindingTracklet/data/, that may potentially be used in an HLS module

For the sake of the eventual HLS-based emulation, `download.sh` now also accepts an option for downloading and unpacking only LUTs.tar.gz, while preserving the default behavior of downloading and unpacking everything when no option is given:
```
download.sh [-h|--help] [-t|--tables]

Options:
  -h, --help    show this help message and exit
  -t, --tables  download and unpack only LUTs.tar.gz
```

This should address #108. These changes have already been made on the fw_synch branch, but since the synchronization is taking some time to complete, I wanted to go ahead and add them to the master branch now.

### DIFF
--- a/emData/download.sh
+++ b/emData/download.sh
@@ -66,11 +66,19 @@ do
   shift
 done
 
-# If either the MemPrints or the LUTs directory exists, assume the script has
-# already been run, and simply exit.
-if [ -d "MemPrints" ] || [ -d "LUTs" ]
+# If the MemPrints directory exists, assume the script has already been run,
+# and simply exit.
+if [ -d "MemPrints" ]
 then
   exit 0
+fi
+
+# If the LUTs directory exists, assume LUTs.tar.gz has already been downloaded
+# and unpacked, and only download and unpack MemPrints.tar.gz.
+memprints_only=0
+if [ -d "LUTs" ]
+then
+  memprints_only=1
 fi
 
 # Exit with an error message if run from a directory other than emData/.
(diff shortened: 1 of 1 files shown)


tomalin (commented):
Looks good. Two small suggestions:
1) To https://github.com/cms-tracklet/firmware-hls/blob/split_download/emData/download.sh#L41 , ensure when option "-h" used, it prints out that the code downloads both LUT & Memory if run without options.
2) At https://github.com/cms-tracklet/firmware-hls/blob/split_download/emData/download.sh#L67 , consider downloading the Memory files if download.sh is run without options, and only the LUT directory already exists.

QUESTION: Is creating lots of symbolic links, based on the incomplete list of processing modules in https://github.com/cms-tracklet/firmware-hls/blob/split_download/emData/download.sh#L67 the best strategy? Unless this lists is completed, one can test the entire nonant. And creating loads of symbolic links is messy. Would it be better to rename the output files created by the emulation, so no symbolic links needed?

aehart:
@tomalin Thank you for the suggestions. I have implemented both.

The reason for the symbolic links is because the output test vectors for one module are, in almost all cases, the input test vectors for another module. I wanted to provide module-specific directories of test vectors to aid with debugging, while avoiding a bunch of duplicate files.

It should be very easy to add an option for setting up these directories for every processing module in a nonant, and maybe that behavior should entirely replace the list of modules at some point. But since the HLS test benches are currently hand-written, i.e., not easy to produce en masse, I think it makes sense to have a restricted list of modules for now.

tomalin:
@aehart Thanks for clarification. Despite your reasoning, my personal preference would have been to keep download.sh simple, and avoid this. But I see your rational, and that it's a matter of taste.

tomalin (approved):
checked and approved

PR #113, opened 2020-09-30
IR update

sseifeln:
Updated IR after synchronization with master.
Had to manually add dtclinklayer.dat file to emData because I couldn't find it in the .tarball.


tomalin:
You say dtclinklayer.dat is missing. I assume you mean dtclinklayerdisk.dat? This is now downloaded by the new emData/download.sh script that was merged into "master" 7 hours ago by https://github.com/cms-tracklet/firmware-hls/pull/112 . (It no longer takes the files from https://cernbox.cern.ch/index.php/s/tsxTkilHDVhnbYF/download , but from new tar files). So if you update to latest master, you can delete dtclinklayerdisk.dat from your PR.

sseifeln:
Thans @tomalin. I didn't check whether it was now included - only modified the location of the LUTs in the test bench after syncing with master. Will make a commit now with it removed (and update the location in the test bench).

tomalin:
Is it possible to simplify the IR test-bench, & adopt a similar style to the test-benches of the other algo steps? For example:
1) The name of InputStubs.h seems misleading, since it actually seems to run the whole IR test-bench. None of the other TBs use a separate class like this. Consider moving all this code inside InputRouter_test.cpp.
2) Please ensure that any constants which differ depending on which of the 48 IR one is using are clearly labelled, and ideally all next to each other in the code. 
3) If you look at MatchEngine_test.cpp, as an example, (though all the others are similar), the whole test bench is less than 100 lines long. It manages this by using functions openDataFile(), writeMemFromFile()& compareMemWithFile()from FileReadUtility.h. Is there a good reason for InputRouter_test.cpp not to use these?

tomalin:
The VMR reads its input data from memories of type InputStubMemory. I would therefore expect the IR to write its output data to these same memories. But I see no use of InputStubMemory anywhere in the IR code. Does this make sense?

tomalin:
The variable names used in the top-level function are used to name the ports of the IP core, so they need to be meaningful. and for the data, ideally indicate which are input and which output arguments. Perhaps rename hStubs & hMemories to inputStubs & outputStubs, respectively?

sseifeln:
@tomalin  - thank you for your comments - I'm replying to all the messages above in one thread to make it simpler to read : 
1) Separating the utilities of the IR test bench: The reason I separated it this way is that when I started to look at the possibility of combining, in a simple test bench, the IR + VMR steps I realized that I couldn't do that 'easily' unless I already had all the utilities needed for the IR separated like this. I can move everything back into InputRouter_test.cpp though since that was just me trying to understand/work-out how the two modules connect together. 
2) Constants that depend on the flavour of the IR : spotted a couple of places in InputRouter.h where that was happening; now fixed. 
3) Length of the IR test bench : I think one reason that the test bench for InputRouter is so much longer is that I wrote it to  be completely generic; so I don't need to change the names of the files/memory prints when testing different instances of the IR. Instead I only need to prove the id of the link that I would like to use to test the IR. Saying that, there were a few functions that I was no longer using so I have removed them. I also had to add a function that compares the contents of the Link_* file with the MemPrint from the emulation - as there were inconsistencies between the two. I'll look through and check if there is anywhere that I can easily re-use the functionality already available in FileReadUtility (I think I do use them for filling the reference memories, but not to prepare the array of stubs used as input for the InputRouter) 
4) Variable names : will change the variable names in the top level function.

tomalin:
Chatted with Sarah.
1) She will delete dtclinklayerdisk.dat from PR, as she confirms it is created by download.sh
2) IR will continue to write DTCStubMemory (but Sarah will rename it to GenericMemory), which is a 36b wide memory, with no code assigning physical meaning to smaller bit ranges within the 36b. Using this, rather than writing to InputStubMemory simplifies the IR interface, since one can have array of DTCStubMemory, rather than array InputStubMemory<DISKPS> + <BARREL2S> etc. We must add constructor to InputStubMemory, so it can be created from DTCStubMemory.
3) IR interface always uses array of 20 DTCStubMemory, even if specific IR needs to talk to fewer than 20 DTCStubMemories. Sarah will check if this wastes resources. As InputRouterTop.h needs no template arguments, it can be renamed InputRouter.h
4) TB currently using compareMems() rather than compareMemWithFile(). This is a workaround for the bug in the emulation that results in incorrect txt files being made. Sarah will add comment to code to make this clear; and after emulator is fixed, TB will change to use compareMemWithFile()
5) TB uses tracker wiring map to get configuration of all the IRs. (i.e. List of txt file names etc.). Sarah will cleanly separate this code that gets the configuration from the code doing the testing. e.g. Put the code doing the testing in InputRouter_test.cpp (deleting InputStubs.h) and put the code getting the configuration (including array of file names) into InputRouter_test_getCfg.cc.
6) Hard-wired constants to be given names.

pwittich:
actually I'm trying to understand this a bit more. Are you suggesting to use the format of the output memory for the input memory just as a convenience to avoid having to have two sets of lists of interfaces on the input? I'm confused. 
> 
> 2. IR will continue to write DTCStubMemory (but Sarah will rename it to GenericMemory), which is a 36b wide memory, with no code assigning physical meaning to smaller bit ranges within the 36b. Using this, rather than writing to InputStubMemory simplifies the IR interface, since one can have array of DTCStubMemory, rather than array InputStubMemory +  etc. We must add constructor to InputStubMemory, so it can be created from DTCStubMemory.

sseifeln:
(Apologies for the very long comment .. please let me know if you'd rather comments like these go in an email thread/some other discussion forum) My understanding is that the input router needs to process stubs from DTC that readout either : 
Barrel PS modules/Barrel PS + Disk PS modules/Disk PS modules  
/Barrel 2S modules/Barrel 2S + Disk 2S modules/Disk 2S modules 
And so the interface to the top level function needs to somehow accommodate all these variants.. which I could have done in one of two ways : 
1) Use 4 arrays at the interface, and manually set those not being used for the specific instant of the IR to null pointers (similar to what is done by the VMR) 
2) Use a single array of 'generic' 36 bit wide memories at the interface. 
Given that the input router needs to have some knowledge of the cabling map in order to convert the 'encoded layer id' used in the DTC stub formation to 'physical layer/disk Id' used by the rest of the project, I decided that it would make more sense to go with option (2); the input router can make use of the same information used to retrieve the layer id to also figure out the position of the different stub coordinates + bends within the bits within the 36 bit stub word. Initially I did this by creating a new InputStubBase of type <Tracker> but not really making use of any of the fields in BitWidths. I discussed this with @djcranshaw and he suggested simply creating a new type of memory .. which I did.  I thought that this would be ok because : 
1) the HLS VMR reads from text files anyways so the type of memory I choose to use at the interface does not directly affect it 
2) I thought that the project generation scripts still needed to be modified to deal with having arrays at the interface of the InputRouter .. so I thought that now would not be a bad time to make such a change 
3) any conversion between interface types at the level of CMSSW can be done with a copy constructor of the sort mentioned by @tomalin  

Does that help at all or have I just confused everyone further?

pwittich:
Hi @sseifeln thanks, some comments here and  inline. Maybe @aehart and others (@aperloff maybe) can also weigh in.

If I understand your proposal correctly this reads to me like the moral equivalent of a pointer to void. the idea being that you just wand to pass in a generic set of memories and use some other knowledge (cabling map) to decode them elsewhere. Did I understand that correctly? Is this already in the code now? I have a hard time imagining how this would be done in a clean way.  Can this be done in such a way that the compiler can know already which of your arrays are empty or which code branches will be used at compile-time (rather than an `if` statement that gets evaluated at run-time)? In my view the advantage of having pointers that are null when unused is that it is very clear to the compiler that this code can be pruned away.  (I think we demonstrated that this works as expected.) 

can you point me at the code that shows how this will work?

> 3. the HLS VMR reads from text files anyways so the type of memory I choose to use at the interface does not directly affect it

I'm not sure I understand this comment. The VMR reads from text files when it is run in a test bench that only tests the VMR. Otherwise it reads from the output memories for the InputRouter, no? 

> 5. any conversion between interface types at the level of CMSSW can be done with a copy constructor of the sort mentioned by @tomalin

I realize we need to keep an eye on CMSSW too, but the main focus should be on writing something that can efficiently be translated into firmware.

sseifeln:
@pwittich  - sorry for the delay. I am going through your comments, thinking, and will reply early next week. Sorry - week has been a bit more hectic than usual.

sseifeln:
hi @tomalin  and @pwittich - I've updated the merge request to address the issues above. Hope I've caught them all!

pwittich:
> hi @tomalin and @pwittich - I've updated the merge request to address the issues above. Hope I've caught them all!

Hi @sseifeln great! any chance you can give us a bit of a guided tour as to what you changed? i.e. respond to some of the comments above?

sseifeln:
@pwittich - yeah sure .. going to start with these first as these were the top of my list : 

- @tomalin asked that I simplify the test bench - I have done so and now only use common functions from the FileReadUtility; only exception is for reading in the link data file where I need to fill an array and not a memory. To do that I added a new function to FileReadUtility.h (https://github.com/cms-tracklet/firmware-hls/blob/IR_Update/TestBenches/FileReadUtility.h#L52-L54) 
- @tomalin asked me to include a check of the valid bit and also modify the code to make sure that it is compatible with the stub format from the DTC ( valid bit is LSB not MSB). The last time I checked the data files in the tarball still didn't include that change (@aehart maybe I missed something)? But anyways - I modified InputRouter.h in preparation for that https://github.com/cms-tracklet/firmware-hls/blob/IR_Update/TrackletAlgorithm/InputRouter.h#L73-L80. Once the tar balls are updated I just need to change those constants 
- @pwittich you had asked that I make sure that each InputRouter uses exactly the correct number of output ports (== number of memories written to by that IR). I've taken Derek's suggestion and added a constant to the top level function (https://github.com/cms-tracklet/firmware-hls/blob/IR_Update/TrackletAlgorithm/InputRouterTop.h#L13-L14) that sets that. I guess that means that eventually a script can be written to modify the top level function for each flavour of IR.

More general comments :

- I think I have managed to remove all hard coded values from the code! 
- I make use of two arrays (which I guess eventually need to be LUTs) to tell the test bench the number of coarse phi bins for stubs from a layer/disk on a given link.

TrackletAlgorithm/InputRouterTop.h
```diff
@@ -0,0 +1,30 @@
...
+// is that really the only way to do this? 
+
+// only for PS10G_1_A for now 
+// other cases to be added 
+// when I have final link map
+constexpr unsigned int cNMemories = 16; 
+constexpr unsigned int cNEntriesLUT = kSizePhiCorrTablePS;
+
+void InputRouterTop( const BXType hBx
+	, const ap_uint<6> hInputLinkId // link id
```
aehart, line 17: `hInputLinkId` and `hLinkNMemories` don't seem to be used anywhere (except for being printed out during C-simulation). Can they be removed?

emData/download.sh
```diff
@@ -13,6 +13,7 @@ luts_url="https://cernbox.cern.ch/index.php/s/DuhCjcykSHZLRhM/download"
 declare -a processing_modules=(
   # VMRouter
   "VMR_L1PHIE"
+  "VMR_L2PHIB"
```
aehart, line 16: Is this change necessary for the InputRouter?

emData/dtclinklayerdisk.dat
```diff
@@ -0,0 +1,25 @@
+2S_1 4 5 -1
```
aehart, line 1: This file is already in the tarball downloaded by `download.sh`. Can we just use that, and not add a separate copy of it to the repo?

project/script_IR.tcl
```diff
@@ -8,23 +8,27 @@ open_project -reset inputrouter
...
-csim_design -compiler gcc -mflags "-j8"
+create_clock -period 240MHz -name slow_clock 
+create_clock -period 360MHz -name fast_clock
+
+set nProc [exec nproc]
+csim_design -compiler gcc -mflags "-j$nProc"
 csynth_design
 cosim_design 
-export_design -format ip_catalog
+export_design -format ip_catalog -flow syn
```
aehart, line 30: The other modules don't run the synthesis during export by default, i.e., during the CI tests. However, I actually like this idea. Implementation is probably much too slow for the CI tests, but synthesis still lets users get a probably-more-accurate look at the resource utilization and timing.

So I support the change you've made, and would like to see it replicated for the other modules, which we could do in a separate PR. But I wonder if @pwittich or @RobertGlein have thoughts on this?

aehart:
@sseifeln Sorry for the delay in finishing the review. I've added a few comments inline regarding some technical points, but otherwise, I think this is good to go.

project/script_IR.tcl
RobertGlein: @aehart The -flow syn synthesizes the design again to double-check it with the results of the csynth. I think if we want to go this path we should do -flow impl. It takes longer but gives you the best estimation based on an implementation in an empty FPGA. Time it takes for the ME:
export_design -format ip_catalog # ~1 min
export_design -format ip_catalog -flow syn # ~10 min
export_design -format ip_catalog -flow impl # ~17 min

sseifeln:
hi @aehart  - thanks for checking the super delayed PR :) Will go back and check your comments later in the week, and re-commit. First quick glance, I thinks it a yes to all, and the changes should be quick to implement.

TrackletAlgorithm/InputRouterTop.h
sseifeln: Thanks for spotting it ... legacy from older version of InputRouterTop. Will remove.

emData/download.sh
sseifeln: No - removing.

emData/dtclinklayerdisk.dat
sseifeln: Sorry, should have removed it.

sseifeln:
Hi @aehart  - the latest commit has the changes requested after the review. The only comment which was not addressed was the one re: changing export_design to use either synthesis or implementation flow because I'm not sure what the conclusion was for that one.

aehart:
> Hi @aehart - the latest commit has the changes requested after the review. The only comment which was not addressed was the one re: changing export_design to use either synthesis or implementation flow because I'm not sure what the conclusion was for that one.

@sseifeln Thanks. This looks good to me. I want to ask if there are opinions regarding turning on the `-flow` option by default in the HLS chat today. Whatever the decision, I'll merge your PR later today.

aehart (approved):

PR #117, opened 2020-10-16
Remove unnamed constants from ME

aperloff:
Remove unnamed constants from the MatchEngine and related code, especially the memory templates. This should solve the ME portion of #110.


aehart:
Just for organizational purposes, I have moved this commit to PR #111.

PR #118, opened 2020-10-16
Changes needed for Vitis HLS

aperloff:
The minimal changes needed to make the MatchEngine code compatible with Vitis HLS 2020.1. This mostly amounts to changes in the shared `settings_hls.tcl` file and the module specific `script_ME.tcl` file. The changes to the module specific file can be added to the other module specific files as needed.

This PR also demonstrates the necessary change to one of the HLS pragma statements, but does not actually implement the change. Other performance enhancing changes still need to be made before the code is efficient when compiled using Vitis HLS.


aperloff:
The check failed because it's not using Vivado/Vitis 2020.1. Should we just hold off on merging this until the group updates (eventually) or should I put in a check for the Xilinx version as well as the specific executable?

pwittich:
Probably it's cleaner for you and everyone to add an explicit check and see if you can add some #ifdef macros in the code rather than keeping code for Vitis commented out. (A cursory glance at UG902 did not suggest there was a version macro number, alas, so maybe this is something you'd need to pass into the code or hack in some other way...)

I would prefer to merge this now rather than have it linger for a long time.

aperloff:
@pwittich @aehart This PR is ready for review...finally.

TestBenches/FileReadUtility.h
```diff
@@ -4,6 +4,7 @@
 #ifndef TestBenches_FileReadUtility_h
 #define TestBenches_FileReadUtility_h
 
+#include <algorithm>
```
aehart, line 7: @aperloff Is this extra include necessary?

TrackletAlgorithm/MatchEngine.cc
```diff
@@ -220,6 +220,9 @@ void MatchEngineTop(const BXType bx, BXType& bx_o,
 #pragma HLS interface register port=bx_o
 #pragma HLS resource variable=inputStubData->get_mem() latency=2
 #pragma HLS resource variable=inputProjectionData->get_mem() latency=2
+// For use with Vitis >=2020.1
```
aehart, line 223: @aperloff Is it possible to automatically choose, using preprocessor macros, which set of pragmas is enabled? This document claims there is a predefined `__VITIS_HLS__` macro that might be useful:
https://www.xilinx.com/html_docs/xilinx2020_2/vitis_doc/migrating_to_vitis_hls.html

project/script_IR.tcl
```diff
@@ -12,7 +15,11 @@ set_top InputRouterTop
 add_files ../TrackletAlgorithm/InputRouterTop.cc -cflags "$CFLAGS"
 add_files -tb ../TestBenches/InputRouter_test.cpp -cflags "$CFLAGS"
 
-open_solution "solution1"
+if { ([string first "vitis" $exe] > -1) && ($year > 2019) } {
```
aehart, line 18: @aperloff Is it possible to define something like an alias of `open_solution` in env_hls.tcl that basically adds the `-flow_target` option when appropriate? It would be nice if we could avoid duplicating this if-else block in each script.

project/script_IR.tcl
```diff
@@ -24,7 +31,14 @@ create_clock -period 240MHz -name slow_clock
 create_clock -period 360MHz -name fast_clock
 
 set nProc [exec nproc]
-csim_design -compiler gcc -mflags "-j$nProc"
```
aehart, line 27: @aperloff Would it make sense to simply drop the `-compiler gcc` option altogether? In versions where the option is available, I think it defaults to gcc anyway, and in any case, I don't think we need a specific C++ compiler.

TestBenches/FileReadUtility.h
aperloff: It is. I was surprised that no one else caught it, but `std::remove` comes from the `algorithm` header. Some of the unit tests were failing because it wasn't included in the FileReadUtility. `std::remove` is used on L85.

project/script_IR.tcl
aperloff: I'd be happy just to from the `-compiler gcc` flag. I too don't see any downside. If you confirm, I will go ahead and remove the flag and the switch statement.

project/script_IR.tcl
aperloff: Based on https://wiki.tcl-lang.org/page/interp+alias I think this is possible. Let me look into it and get back to you.

project/script_IR.tcl
aehart: I say go ahead and remove it.

TestBenches/FileReadUtility.h
aehart: That's good enough for me. But I am curious what unit tests you're talking about. I don't see any warnings or errors in the output from the CI tests.

tomalin:
In README.md, after line "vivado_hls -f script_PR.tcl", add line "or: vitis_hls -f script_pr.tcl"

pwittich:
@aperloff any updates?

project/script_IR.tcl
aperloff: The `-compiler gcc` option is now removed.

TrackletAlgorithm/MatchEngine.cc
aperloff: Done

aperloff:
@pwittich @aehart I have rebased, made all of the changes requested in the previous review, and all of the CI/CD checks are passing. Please let me know if you have any other questions or concerns.

pwittich (approved):
This looks ready to merge from my POV

aehart (approved):
Looks good to me too. Thanks, @aperloff.

PR #119, opened 2020-10-23
Remove NEntries ports from output memories for each module

djcranshaw:
This PR removes the nentries ports for all the output memories for each of the HLS blocks. To test each module in HLS, the HLS test benches instead clear the full memory between each bunch crossing, and in the real FW system a signal in the top-level HDL will be sent to each memory at the end of each bx, telling the memory to set the relevant number of entries for the page to zero. More disussion can be found in issue #58, and in this talk https://indico.cern.ch/event/940922/contributions/3955291/attachments/2079826/3493207/HLSChat_nentries_2020-07-24.pdf

With this pull request (with the exception of MC, discussed below), the clock period and II achieved either stays the same or improves slightly, and the resource usage for each block goes down in every case (by a factor of ~2 for the VMR, modestly for the other blocks).

For the MC, before this PR we were sacrificing the first clock of the BX to zero the memories. Since we no longer have to do that, we now start the work of the MC on the first clock instead of the second. This change results in a change in the post-synthesis (post-implementation) timing from 2.84 (3.24) to 3.45 (3.56) ns. It's not clear why this is, but since the CP remains under our target even after the increase, I've left it as is.

### DIFF
--- a/TestBenches/FileReadUtility.h
+++ b/TestBenches/FileReadUtility.h
@@ -30,33 +30,6 @@ bool openDataFile(std::ifstream& file_in, const std::string& file_name)
   return success;
 }
 
-template<class DataType>
-void readEventFromFile(DataType& memarray, std::ifstream& fin, int ievt){
-
-  std::string line;
-
-  if (ievt==0) {
-    getline(fin, line);
-  }
-
-  memarray.clear(ievt);
-
-  while (getline(fin, line)) {
-    
-    if (!fin.good()) {
-      return;
-    }
-
-    if (line.find("Event") != std::string::npos) {
-	return;
-    }
-    else {
-      memarray.write_mem_line(ievt,line);
-    }
-    
-  }
-}
-
 std::vector<std::string> split(const std::string& s, char delimiter)
 {
   std::vector<std::string> tokens;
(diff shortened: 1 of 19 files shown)


aehart:
@djcranshaw Thanks for putting this together. Overall I think it looks really good. I just added the following minor changes:
- Added the rewind option to the TE and ME. Clearing the output memories was the last hurdle to achieving this. Now that we no longer have to do this in the HLS modules, I figured we might as well add this option.
- Moved the memory counter initialization out of the main processing loop in the TC so that we no longer have to sacrifice an iteration to this.
- Changed the type of the memory counters in the VMR from `int` to `ap_uint<kNBits_MemAddr>`. This reduces the FF and LUT utilization by ~2.5 and ~1.5, respectively.
- A few methods with `bool` return types were missing return statements. I'm surprised that the compiler didn't complain about these, but in any case, I added return statements to these methods.
- A couple functions took in parameters of type `const bool &`, which I changed to `const bool`. This probably doesn't matter, but it could be slightly more efficient, given that `bool` is usually represented with one byte, while a reference is usually represented with eight.

aehart (approved):
Everything looks good to me.

djcranshaw:
> @djcranshaw Thanks for putting this together. Overall I think it looks really good. I just added the following minor changes:
> 
> * Added the rewind option to the TE and ME. Clearing the output memories was the last hurdle to achieving this. Now that we no longer have to do this in the HLS modules, I figured we might as well add this option.
> * Moved the memory counter initialization out of the main processing loop in the TC so that we no longer have to sacrifice an iteration to this.
> * Changed the type of the memory counters in the VMR from `int` to `ap_uint<kNBits_MemAddr>`. This reduces the FF and LUT utilization by factors of ~2.5 and ~1.5, respectively.
> * A few methods with `bool` return types were missing return statements. I'm surprised that the compiler didn't complain about these, but in any case, I added return statements to these methods.
> * A couple functions took in parameters of type `const bool &`, which I changed to `const bool`. This probably doesn't matter, but it could be slightly more efficient, given that `bool` is usually represented with one byte, while a reference is usually represented with eight.

I just looked at the diff for all these changes and they all look good to me. I'm surprised about the missing returns not causing errors as well. Thanks for catching it in any case.

PR #121, opened 2020-10-28
Memory counter size

aehart:
I switched the few `int` memory counters remaining to `ap_uint<kNBits_MemAddr>`. This has a positive impact on the resource utilization reported by HLS for the affected modules, especially the TC and MC:
| Module | Change in FF utilization | Change in LUT utilization |
| --- | --- | --- |
| TE | -2% | -1% |
| TC | -7% | -17% |
| ME | -2% | -2% |
| MC | -13% | -7% |


djcranshaw (approved):
All straightforward changes, looks good to me. Vivado seems to indicate (e.g. note on P16 of UG902 v2019.2) that this shouldn't be necessary because any bits outside the address range of the memory being written to aren't used and should therefore be optimized away. Oh well.

Issue #120, opened 2020-10-24
Unused FM memories in MC


aehart:
@bryates I wanted to make note of this issue here so that we don't forget about it.

The current top function for the MC, which represents MC_L3PHIC, has ports for eight FM memories, while it is connected to only two in the wires file. We should use template parameters to "inform" HLS of this, so that it only generates ports for the two FM memories in use.

An example of how this might be done can be found in the TC, where bit masks are defined in TrackletCalculator_parameters.h that are then used to select which TPROJ memories are valid and can be written to in TrackletCalculator.h.

bryates:
This has been addressed in PR #123

PR #123, opened 2020-10-30
Added bit masks to remove unused FM ports in MC

bryates:
Addressing Issue #120

### DIFF
--- a/TestBenches/MatchCalculator_test.cpp
+++ b/TestBenches/MatchCalculator_test.cpp
@@ -75,7 +75,7 @@ int main() {
 
   // loop over events
   for (int ievt = 0; ievt < nevents; ++ievt) {
-    cout << "Event: " << dec << ievt << endl;
+    //cout << "Event: " << dec << ievt << endl;
 
     fullmatch[0].clear();
 //    fullmatch[1].clear();
--- a/TestBenches/MatchCalculator_test.cpp
+++ b/TestBenches/MatchCalculator_test.cpp
@@ -75,7 +75,7 @@ int main() {
 
   // loop over events
   for (int ievt = 0; ievt < nevents; ++ievt) {
-    //cout << "Event: " << dec << ievt << endl;
+    cout << "Event: " << dec << ievt << endl;
 
     fullmatch[0].clear();
 //    fullmatch[1].clear();
--- a/TrackletAlgorithm/MatchCalculator.h
+++ b/TrackletAlgorithm/MatchCalculator.h
@@ -148,32 +148,32 @@ ap_uint<width> iabs( ap_int<width> value )
 template<bool phi, TF::layer L, int width, int depth>
 void readTable_Cuts(ap_uint<width> table[depth]){
   if (phi){ // phi cuts
-    if (L==0){
+    if (L==TF::L1){
       ap_uint<width> tmp[depth] =
 #include "../emData/MC/tables/MC_L1PHIC_phicut.tab"
       for (int i = 0; i < depth; i++) table[i] = tmp[i];
     }
-    else if (L==1){
+    else if (L==TF::L2){
       ap_uint<width> tmp[depth] =
 #include "../emData/MC/tables/MC_L2PHIC_phicut.tab"
       for (int i = 0; i < depth; i++) table[i] = tmp[i];
     }
-    else if (L==2){
+    else if (L==TF::L3){
       ap_uint<width> tmp[depth] =
 #include "../emData/MC/tables/MC_L3PHIC_phicut.tab"
       for (int i = 0; i < depth; i++) table[i] = tmp[i];
     }
-    else if (L==3){
+    else if (L==TF::L4){
       ap_uint<width> tmp[depth] =
 #include "../emData/MC/tables/MC_L4PHIC_phicut.tab"
       for (int i = 0; i < depth; i++) table[i] = tmp[i];
     }
-    else if (L==4){
+    else if (L==TF::L4){
       ap_uint<width> tmp[depth] =
 #include "../emData/MC/tables/MC_L5PHIC_phicut.tab"
       for (int i = 0; i < depth; i++) table[i] = tmp[i];
     }
-    else if (L==5){
+    else if (L==TF::L6){
       ap_uint<width> tmp[depth] =
 #include "../emData/MC/tables/MC_L6PHIC_phicut.tab"
       for (int i = 0; i < depth; i++) table[i] = tmp[i];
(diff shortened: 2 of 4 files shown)


aehart:
@bryates Nice! Seems to work as expected. I have just a few suggestions:
- We want to avoid hard-coding the mask for a specific instance of the MC in the MatchCalculator template function, which you do here:
https://github.com/cms-tracklet/firmware-hls/blob/ead0f05406d0fead5a699dccb9f3e3847b445b11/TrackletAlgorithm/MatchCalculator.h#L332-L340
Instead, these should be determined from the LAYER, DISK, and PHISEC parameters. One way of doing this is by passing these to a `constexpr` function that returns the appropriate bit mask, e.g., for the TC:
https://github.com/cms-tracklet/firmware-hls/blob/ead0f05406d0fead5a699dccb9f3e3847b445b11/TrackletAlgorithm/TrackletCalculator_parameters.h#L40-L42
- Right now, you represent the bit mask using eight constants, but since each constant has a unique shift, you could just OR them all into a single mask. This is partially a matter of style, but having a single mask would probably make addressing the previous suggestion easier.
- This is a really minor thing, but the right-shift by `projseed` isn't really necessary on these lines:
https://github.com/cms-tracklet/firmware-hls/blob/ead0f05406d0fead5a699dccb9f3e3847b445b11/TrackletAlgorithm/MatchCalculator.h#L829-L871

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -798,36 +826,52 @@ void MatchCalculator(BXType bx,
     if(newtracklet && goodmatch==true) { // Write out only the best match, based on the seeding 
       switch (projseed) {
       case 0:
+      if((FML1L2 & mask_L1L2) >> projseed) { //projseed 0-7 equivalent to shift
```
pwittich, line 829: In this `switch` statement since the case label is the same as the value of `projseed` if you make the amount you shift by the same as the case label it will be more clear to the compiler that these are constant expressions (assuming the shift doesn't go away like @aehart says)

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -311,6 +311,34 @@ void MatchCalculator(BXType bx,
...
+  static const uint16_t mask_L3L4 = 1 << shift_L3L4;
+  static const uint16_t mask_L4L5 = 1 << shift_L4L5;
+  static const uint16_t mask_L5L6 = 1 << shift_L5L6;
+  static const uint16_t mask_D1D2 = 1 << shift_D1D2;
+  static const uint16_t mask_D3D4 = 1 << shift_D3D4;
+  static const uint16_t mask_L1D1 = 1 << shift_L1D1;
+ 
+  // MC_L3PHIC mask {1: on, 0: off}
+  static const uint16_t FML1L2 = 1 << shift_L1L2;
+  static const uint16_t FML2L3 = 0 << shift_L2L3;
```
pwittich, line 334: what does it mean to shift a zero? I am confused. And are these static const or constexpr?

TrackletAlgorithm/MatchCalculator.h
pwittich: also probably the switch statement argument should not be a magic number.

bryates:
@aehart thanks for the pointers. I've changed things over to look more similar to the tracklet calculator. The mask is now stored in a constexpr in MatchCalculator_parameters.h, and projseed was completely removed from the mask evaluation. This also adresses the comment from @pwittich about projseed.

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -314,30 +319,30 @@ void MatchCalculator(BXType bx,
   static const uint8_t shift_L1L2 = 0;
```
pwittich, line 319: these are still static const -- can't they be constexpr?

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -826,49 +831,49 @@ void MatchCalculator(BXType bx,
     if(newtracklet && goodmatch==true) { // Write out only the best match, based on the seeding 
       switch (projseed) {
       case 0:
-      if((FML1L2 & mask_L1L2) >> projseed) { //projseed 0-7 equivalent to shift
+      if(FMMask<LAYER, PHISEC>() & mask_L1L2) { //projseed 0-7 equivalent to shift
```
pwittich, line 834: lots of bad formatting here please fix

TestBenches/MatchCalculator_test.cpp
```diff
@@ -75,7 +75,7 @@ int main() {
 
   // loop over events
   for (int ievt = 0; ievt < nevents; ++ievt) {
-    //cout << "Event: " << dec << ievt << endl;
+    cout << "Event: " << dec << ievt << endl;
```
pwittich, line 78: probably should wrap this into an `ifdef __SYNTHESIS__`

bryates:
I've cleaned up the code a bit, and moved the shift and mask values to MatchCalculator_parameters.h. They should not have to bee constexpr, but @pwittich please let me know if you'd prefer they are.

TrackletAlgorithm/MatchCalculator.h
bryates: Moved to MatchCalculator_parameters.h

TrackletAlgorithm/MatchCalculator.h
bryates: Cleaned up

TestBenches/MatchCalculator_test.cpp
bryates: I forgot to comment that line out again after testing. It is now commented.

pwittich:
There are a lot of magic numbers in this code (What do you think @aehart @tomalin ) Can you get rid of some of them with `auto` and otherwise get rid of them with meaningful constants?

aehart (commented):
> There are a lot of magic numbers in this code (What do you think @aehart @tomalin ) Can you get rid of some of them with `auto` and otherwise get rid of them with meaningful constants?

I agree. Liberal use of `auto` in particular will go a long way toward cleaning up the magic bit widths.

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -145,35 +145,35 @@ ap_uint<width> iabs( ap_int<width> value )
 // Template to get look up tables
 
 // Table for phi or z cuts
-template<bool phi, int L, int width, int depth>
+template<bool phi, TF::layer L, int width, int depth>
 void readTable_Cuts(ap_uint<width> table[depth]){
   if (phi){ // phi cuts
-    if (L==1){
+    if (L==0){
```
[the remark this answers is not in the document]
bryates: Simple enough, I'll do that now.

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -225,8 +225,14 @@ void readTable_Cuts(ap_uint<width> table[depth]){
 //////////////////////////////////////////////////////////////
 
 // MatchCalculator
+namespace MC {
+  enum itc {UNDEF_ITC, A = 0, B = 1, C = 2, D = 3, E = 4, F = 5, G = 6, H = 7, I = 8, J = 9, K = 10, L = 11, M = 12, N = 13, O = 14};
```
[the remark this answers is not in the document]
bryates: Good point, I'll make that change.

bryates:
Several changes in the latest commit:
- itc -> imc
- shifts completely replaced with TF numbers 
https://github.com/cms-tracklet/firmware-hls/blob/aba7d82bf32bc440ee78482f285ef0b0af2871ed/TrackletAlgorithm/MatchCalculator_parameters.h#L7-L15
- table reads now use TF numbers
https://github.com/cms-tracklet/firmware-hls/blob/aba7d82bf32bc440ee78482f285ef0b0af2871ed/TrackletAlgorithm/MatchCalculator.h#L151

bryates:
One more change. I was able to completely remove the mask constants, and just use the seed values directly.
https://github.com/cms-tracklet/firmware-hls/blob/01c5f17e6748895fb8e156a0bc6f5488f5e16a86/TrackletAlgorithm/MatchCalculator_parameters.h#L12-L14

aehart (approved):
Looks good to me.

PR #126, opened 2021-01-07
MC now throws an error for missing FullMatches

bryates:
truncation = false


aehart:
FYI, I just rebased with respect to the master branch after merging @sseifeln's PR.

pwittich:
It looks like now the only thing preventing the merge is the missing agreement from the MC (or did I miss any modules).

aehart:
> It looks like now the only thing preventing the merge is the missing agreement from the MC (or did I miss any modules).

@pwittich I think that's right. All the modules have been updated to report an error on missing entries, and the MC was the only one to have this error (congratulations, @bryates).

bryates:
The cause of the MC issue turned out to indeed be the priority encoder. It had three layers, resulting in no candidate matches being examined until the third loop iteration. This only caused an error when a busy event needed all 108 loop iterations. I've written a new, simpler priority encoder, which now perfectly matches the test vectors.

aehart:
Thanks, @bryates. Looking at the job artifact, I see that the pipeline II is now 7 and the estimated clock period is 5.632 ns. Do you have ideas about how to fix these?

bryates:
@aehart I must have missed the II=7, I thought it was still 1. I'm looking into this, along with the timing now.

bryates:
Here is a bit of context for what is going on with the MC. The priority encoder is designed to always select the candidate match with the smallest projection index. The original priority encoder did this by pairs. This is done to mitigate the effects of the time it takes to perform these comparison (up to 0.713ns), as well as the 2 clock latency of the memories. The priority encoder required three layers, and each layer is manually pipelined. This means the first three iterations have no final match outputs. This only becomes a real issue when we have busy events which require all 108 iterations to process all the full matches. This is also why one of fixes worked by adding three extra iterations to the MC loop. Upon trying to write a priority encoder without multiple layers (and hence no iteration delays) the time required for the comparisons + the 2 clock read latency resulted in a final timing of ~15ns.

I have three ideas on how to move forward:
1. We simply let the MC run for 3 extra iterations (111 instead of 108 in the current configuration). I'm not sure what the implications would be on the full project. Would all modules down stream of the MC have three clocks of deadtime, or would this result in the modules becoming out of sync?
2. We reduce the MC emulation down to 105 iterations. This would ensure busy events in the emulation are truncated the same as the HLS implementation, and produce the correct tests vectors. However, this would still mean the first three clocks of the HLS MC would not produce any full match outputs.
3. I can continue to work on an improved priority encoder which removes this three clock delay. So far, this has progressed slowly, but I have gained a lot of knowledge of the various bottlenecks in the MC. Maybe 1 and/or 2 are acceptable short term solutions while we attempt to write a better priority encoder?

aehart:
@bryates Others should chime in if they have opinions, but I think 2 is probably the best option for now.

I think we just need to know what effect on tracking efficiency the removal of these three iterations has. If it is not too significant, which is what I expect, then we can just reduce the number of iterations in the emulation and call it good for now. And if you have extra brain cycles in the future, you can think of how to rework the priority encoder so that we use all 108 iterations.

djcranshaw:
I agree with @aehart, number 2 I think is the best option to allow this PR to proceed and cause the modules to generate errors if the outputs don't match the test files (while keeping an eye on eventually implementing 3). Number 1 won't work, because if MC isn't accepting new bx data every 108 clocks, we'd have to slow down the whole chain to 111 clocks to prevent ME from overwriting data in the CM memories. I ran a quick emulation with 100 ttbar PU200 events, one where all modules can process 108 memory entries, and one where the MC processes 105 in each memory, and the remaining modules process 108. The results are
Overall Tracking Efficiency - All modules 108: 95.3765
Overall Tracking Efficiency - MC 105, rest 108: 95.2884
So the effect on tracking is pretty minimal, as expected (although it isn't zero).

aehart:
> @aehart @pwittich what do you think?

I agree that for now, we should just let the MC fail its HLS-build test, and I made the corresponding change to .gitlab-ci.yml.

Then we should discuss when we want to do another synchronization with the emulation. We seem to have accumulated at least a couple of hacks related to our current set of test-vectors, so maybe that time is soon. But let's plan on discussing it in the next HLS chat, since it will affect multiple people.

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -868,9 +868,9 @@ void MatchCalculator(BXType bx,
     goodmatch      = goodmatch_next;
     projseed       = projseed_next;
 
-  }// end MC_LOOP 
+    if (istep==kMaxProc-kMaxProcOffset(module::MC)-1) bx_o = bx;
```
aehart, line 871: @bryates Looks like you undid changes here that @djcranshaw made in PR #130.

Could you please un-undo these changes? I think this file should match what's on the master branch.

aehart (approved):
Thanks @bryates. It all looks good to me now.

PR #132, opened 2021-02-11
Move to new repo

RobertGlein:
Test if it worked:
- updated gitlab-mirror-and-ci-action
- moved gitlab CI/CD repo to new shared group (cms-l1tk)
- new gitlab token
- updated github token

### DIFF
--- a/.github/workflows/GitLab_CI.yml
+++ b/.github/workflows/GitLab_CI.yml
@@ -1,6 +1,7 @@
 name: Mirror and run GitLab CI
 
 on: # Controls when the action will run.
+  workflow_dispatch:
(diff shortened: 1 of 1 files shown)


aperloff (commented):
Looks good. I can see that the pipeline on GitLab (https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/2309755) completed successfully. Now all tokens are stored within the repos, which all belong to groups. The only remaining item which belongs to an individual is the gitlab-mirror-and-ci-action, but this is unlikely to change in the near future.

aehart (approved):

PR #135, opened 2021-02-25
Use array for nentries port inside tf_mem*.vhd & replace clogb2

tomalin:
Several changes to make our VHDL shorter:

1) Individual "nent_o*" ports (one for each page) inside tf_mem.vhd & tf_mem_bin.vhd replaced with single "nent_o" port, using an array, where the array length is equal to the number of pages in the memory, configurable via GENERIC parameter. This substantially shortens the VHDL inside tf_mem*.vhd, and allows it to work for an arbitrary number of pages.

2) tf_top*.vhd & SectorProcessor.vhd updated to make use of modified tf_mem*.vhd from (1). They can now do the port assignment for "nent_o" in a single line of VHDL per memory, so are much shorter. To make this possible, the multidimensional "nentries" arrays needed changing, such that their first index now represents the ID of the memory (whereas this used to be in the last index). This means that all lines involving "*nentries*" have changed.

N.B. The new version of SectorProcessor.vhd is written by the updated python scripts in https://github.com/cms-L1TK/project_generation_scripts/tree/ian_MemNentArr , for which a separate PR will be made. 

3) A new implementation of clogb2(int N) added to tf_pkg.vhd. This now returns the number of bits needed to address an array with N elements. Comments in this function explain how this differs from the old version of this function (which I've temporarily renamed to clogb2_old()). The old & new functions give identical results if N is an exact power of two (unless N=2), which means that most of the code using this function didn't need changing.

4) Half a dozen new "type"'s added to tf_pkg.vhd to support the above changes. (For the moment, I've grouped these together after the other types, to make it easier to see them. We should move them to a more aesthetically pleasing position later).  The "nent_o" port inside tf_mem.vhd is based on type t_arr_7b, which hard-wires assumption that 7b are enough to encode "nent_o". We could alternatively switch to type t_arr_meb, which derives the number of bits from log2(max_nent).

5) Small change to tb_tf_top.vhd, modifying interface of its call to SectorProcessor.vhd, since the latters nentries port is now an array.

Validation: I've checked that the output of CompareMemPrintsFW.py is unchanged when run on the output of the Vivado simulation. And that the code still synthesises. Check done for both tf_top_full.vhd & SectorProcessor.vhd.

### DIFF
--- a/IntegrationTests/common/hdl/tf_mem.vhd
+++ b/IntegrationTests/common/hdl/tf_mem.vhd
@@ -35,7 +35,7 @@ entity tf_mem is
     INIT_FILE       : string := "";                --! Specify name/location of RAM initialization file if using one (leave blank if not)
     INIT_HEX        : boolean := true;             --! Read init file in hex (default) or bin
     RAM_PERFORMANCE : string := "HIGH_PERFORMANCE" --! Select "HIGH_PERFORMANCE" (2 clk latency) or "LOW_LATENCY" (1 clk latency)
-    ); 
+    );
   port (
     clka      : in  std_logic;                                      --! Write clock
     clkb      : in  std_logic;                                      --! Read clock
--- a/IntegrationTests/common/hdl/tf_mem.vhd
+++ b/IntegrationTests/common/hdl/tf_mem.vhd
@@ -30,11 +30,12 @@ use work.tf_pkg.all;
 entity tf_mem is
   generic (
     RAM_WIDTH       : natural := 18;               --! Specify RAM data width
-    RAM_DEPTH       : natural := 1024;             --! Specify RAM depth (number of entries)
+    NUM_PAGES       : natural := 2;                --! Specify no. Pages in RAM memory
+    RAM_DEPTH       : natural := NUM_PAGES*PAGE_OFFSET; --! Leave at default. RAM depth (no. of entries)
     INIT_FILE       : string := "";                --! Specify name/location of RAM initialization file if using one (leave blank if not)
     INIT_HEX        : boolean := true;             --! Read init file in hex (default) or bin
     RAM_PERFORMANCE : string := "HIGH_PERFORMANCE" --! Select "HIGH_PERFORMANCE" (2 clk latency) or "LOW_LATENCY" (1 clk latency)
-    );
+    ); 
   port (
     clka      : in  std_logic;                                      --! Write clock
     clkb      : in  std_logic;                                      --! Read clock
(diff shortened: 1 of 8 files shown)


IntegrationTests/common/hdl/tf_mem.vhd
```diff
@@ -30,11 +30,12 @@ use work.tf_pkg.all;
...
   generic (
     RAM_WIDTH       : natural := 18;               --! Specify RAM data width
-    RAM_DEPTH       : natural := 1024;             --! Specify RAM depth (number of entries)
+    NUM_PAGES       : natural := 2;                --! Specify no. Pages in RAM memory
+    RAM_DEPTH       : natural := NUM_PAGES*PAGE_OFFSET; --! Leave at default. RAM depth (no. of entries)
     INIT_FILE       : string := "";                --! Specify name/location of RAM initialization file if using one (leave blank if not)
     INIT_HEX        : boolean := true;             --! Read init file in hex (default) or bin
     RAM_PERFORMANCE : string := "HIGH_PERFORMANCE" --! Select "HIGH_PERFORMANCE" (2 clk latency) or "LOW_LATENCY" (1 clk latency)
-    );
+    );
```
RobertGlein, line 38: rm trailing whitespace

IntegrationTests/common/hdl/tf_pkg.vhd
```diff
@@ -63,22 +66,29 @@ package tf_pkg is
...
   type t_arr_1d_int is array(natural range <>) of integer;                  --! 1D array of int
   type t_arr_2d_int is array(natural range <>,natural range <>) of integer; --! 2D array of int
   type t_arr_2d_slv is array(natural range <>, natural range <>) of std_logic_vector(EMDATA_WIDTH-1 downto 0); --! 2D array of slv
 
-    -- ########################### Functions ################################################################
-  function clogb2 (bit_depth : integer) return integer;
+  -- clogb2(MAX_ENTRIES) = 6.
+  type t_arr_8_5b  is array(integer range<>) of t_arr8_5b;
+  type t_arr_7b  is array(integer range<>) of std_logic_vector(6 downto 0);
+  type t_arr_meb is array(integer range<>) of std_logic_vector(clogb2(MAX_ENTRIES)-1 downto 0);
```
RobertGlein, line 84: You do not really use t_arr_meb and t_arr8_2_meb, do you?
If you want to keep it a comment would be nice that it is for binned memories

RobertGlein (approved):
Great job with using the undefined array but defining the slv.
Just some minor comments

IntegrationTests/common/hdl/tf_mem.vhd
tomalin: fixed

IntegrationTests/common/hdl/tf_pkg.vhd
tomalin: t_arr_meb commented out and t_arr8_2_meb deleted. And added comment that t_arr_meb could optionally be used in place of t_arr_7b. (Doing so would have advantage that it would adjust #bits automatically if MAX_ENTRIES changed, but the disadvantage that the number of bits used would not be immediately visible, making it harder to check consistency with the HLS IP interface).

PR #141, opened 2021-03-15
Sync with emulation + TrackletProcessor

aehart:
This PR syncs the existing HLS modules with what will soon be the HEAD of the main development branch of the emulation. It is initially a draft PR until work on the emulation settles down and it can be tagged.

The first working version of the TrackletProcessor is also included.

PR #143, opened 2021-03-30
TC test-bench generalization

aehart:
This PR consolidates the multiple test-benches for the TC into one test-bench that can handle all currently supported TC instances. The new TBHelper class, defined in TestBenches/FileReadUtility.h, is used to organize the test-vector files.

I also expanded TrackletCalculator_parameters.h and TrackletCalculatorTop.{h,cc} so that all supported TC instances are now defined. These files were generated with a Python script, which I put in emData/ and called generate_TC.py. TrackletCalculator.h was also cleaned up a bit, mostly to eliminate redundant template parameters.

Finally, I updated project/script_TC.tcl so that the C-simulation is run for all TC instances that currently pass it. The C-synthesis, C/RTL cosimulation, and export are still only run for TC_L1L2E, since these steps take longer to run.

I would propose that we first merge these changes. Then we can decide when the other test-benches should be updated, so as to not interfere with PR #141.


TestBenches/Macros.h
```diff
@@ -0,0 +1,1083 @@
...
+#define __VMR_D3PHID__ 39
+#define __VMR_D4PHIA__ 40
+#define __VMR_D4PHIB__ 41
+#define __VMR_D4PHIC__ 42
+#define __VMR_D4PHID__ 43
+#define __VMR_D5PHIA__ 44
+#define __VMR_D5PHIB__ 45
+#define __VMR_D5PHIC__ 46
+#define __VMR_D5PHID__ 47
+#define __TE_L1PHIA1_L2PHIA1__ 48
```
pwittich, line 66: @aehart leading underscores in preprocessor macros are reserved for the compiler by the standard. can you fix this?

https://stackoverflow.com/questions/1449181/what-does-double-underscore-const-mean-in-c/1449301#1449301

project/script_TC.tcl
```diff
@@ -3,29 +3,55 @@
...
 # source files
 set CFLAGS {-std=c++11 -I../TrackletAlgorithm}
-set_top TrackletCalculator_L1L2G
 add_files ../TrackletAlgorithm/TrackletCalculatorTop.cc -cflags "$CFLAGS"
-add_files -tb ../TestBenches/TrackletCalculator_L1L2G_test.cpp -cflags "$CFLAGS"
-
-open_solution "solution1"
-
-# Define FPGA, clock frequency & common HLS settings.
-source settings_hls.tcl
```
pwittich, line 18: it seems like this test bench no longer sources the common `settings_hls.tcl` -- is this obsolete?

aehart: It does, down on line 44:
https://github.com/cms-L1TK/firmware-hls/blob/9cede70f59afb6a19e6c8f021fb42794e802f14d/project/script_TC.tcl#L44

It needs to do this inside the loop because the clock has to be created for each solution.

emData/generate_TC.py
```diff
@@ -0,0 +1,293 @@
...
+topFile.write("////////////////////////////////////////////////////////////////////////////////\n")
+topFile.write("// Top functions for various TrackletCalculators (TC). For each iteration of\n")
+topFile.write("// the main processing loop, a TC retrieves a pair of stub indices from one of\n")
+topFile.write("// the stub-pair memories, and in turn, these indices are used to retrieve one\n")
+topFile.write("// stub each from an inner and an outer all-stub memory. This pair of stubs is\n")
+topFile.write("// used to calculate a rough set of helix parameters, which are written to the\n")
+topFile.write("// tracklet-parameter memory if the tracklet passes cuts on rinv and z0. Rough\n")
+topFile.write("// projections to additional layers and disks are also calculated and are\n")
+topFile.write("// written to the appropriate tracklet-projection memories.\n")
+topFile.write("////////////////////////////////////////////////////////////////////////////////\n")
```
pwittich, line 144: you might be able to simplify this piece of code by using triple quotes for a single, multi-line string, similar to the docstring. convention

aehart: I have switched to multiline strings wherever possible. But I used parentheses instead of triple quotes so that I could indent the strings, which I think improves readability a little.

pwittich (approved):
Ready for merge

PR #144, opened 2021-04-01
PR-ME-MC VHDL now using arrays & loops

tomalin:
The top-level VHDL of the PR-ME-MC chain in this PR was created by the updated scripts in https://github.com/cms-L1TK/project_generation_scripts/pull/19 . 

In PRMEMC/hdl/, this consists of the following files created by the scripts
1) SectorProcessor.vhd (updated top-level VHDL)
2) SectorProcessorFull.vhd (new top-level VHDL with additional output ports from intermediate memories in chain)
3) memUtil_pkg.vhd (definitions of enums & arrays).

The test-bench PRMEMC/tb/tb_tf_top.vhd has been modified to be compatible with the new top-level VHDL:

a) Old signal types copied to new enum-indexed arrays before TB calls SectorProcessor*.vhd.
b) Added option to call SectorProcessorFull.vhd

Two bugs were also fixed in TB, which significantly improves FW vs SW agreement.


djcranshaw (approved):
All the changes look good to me. We should perhaps make an issue about the write_result process in the hand-written tb losing the last two entries per event so as not to lose track of it.

tomalin:
> All the changes look good to me. We should perhaps make an issue about the write_result process in the hand-written tb losing the last two entries per event so as not to lose track of it.

Agreed. I'll add an issue once this is merged.

PR #147, opened 2021-04-29
Simplification of PR-ME-MC VHDL test-bench & minor fixes

tomalin:
Simplification of PRMEMC VHDL test-bench + minor bug fixes:

1) Removed call from PRMEMC test-bench to R.Glein's hand-written top-level VHDL, as no longer needed, now top-level VHDL written by python scripts is so good.
2) Tidied up code doing debug printout to screen in PRMEMC test-bench. (Functionality unchanged)
3) Removed obsolete log2 function from common VHDL utility package. (Newer, improved one was added a few weeks ago, and is working fine).
4)  Simplified VHDL procedures in utility package for reading .txt files from emData/ and loading them into arrays. (Since arrays contain a bin for the event number, there was no need to bin the data also by page number, and this wasted memory). N.B. The signals passed to the top-level VHDL are derived from these arrays.
     Also fixed small bug in these procedures that was giving rise to "array out of bounds" warning.
5) Fixed bug in .tcl script that caused Vivado simulation to always be run twice.
6) Added missing timing constraints file. 
7) Improved synthesis .tcl script, so it prints timing report, and includes options to improve timing performance.
8) Added implementation .tcl script.

Validation: Checked that .txt files written by TB are unchanged. And checked that everything still synthesises.


tomalin:
@rzouCERN please delay review. I just spotted that the TB output is now wrong. I must have put in a last minute, unvalidated fix that messed it up ....

tomalin:
@rzouCERN You can now review. I've solved the problem mentioned above by commenting out the .xdc file in makeProject.tcl. For reasons I don't understand, it causes some of the signals (such as PR_DONE) to be delayed by 100ns, which then leads to incorrect .txt file output from Vivado simulation.

IntegrationTests/common/script/impl.tcl
```diff
@@ -0,0 +1,31 @@
+#####################
```
rzouCERN, line 1: Would we ever want to run implementation like this? Do we want to run post-implementation timing simulation here? I'm not sure if it really tells us anything accurate. In order to put this on chip we will need to compile this with the actual pin assignments and a top level that connects the actual input/output pins to this 'top level'. When I ran this I saw a lot of "Route 35-198" warnings in the log that complains that "Port XXX does not have an associated HD.PARTPIN_LOCS, which will prevent the partial routing of the signal XXX. Without this partial route, timing analysis to/from this port will not be accurate, and no routing information for this port can be exported." Even if we want to look at the post-implementation timing report for such a simple chain I'm not sure if this provides any accuracy.

IntegrationTests/common/script/synth.tcl
```diff
@@ -1,21 +1,34 @@
...
+
 # Synthesis
 update_compile_order -fileset sources_1
 launch_runs synth_1 -jobs 4
 wait_on_run synth_1
+
+# Make reports
 open_run synth_1
-report_utilization -hierarchical -file report_utilization_hierarchical.txt
+report_timing_summary -file timing_synth.rpt
```
rzouCERN, line 31: Since it takes no time to generate this report I'm ok with it. But we should keep in mind that post-synthesis timing report is not accurate at all and shouldn't be used to evaluate our timing performance.

rzouCERN (changes requested):
Thanks a lot Ian for such diligent work at simplifying and perfecting the test-bench! You clearly have put a lot of thoughts and efforts into this. I really appreciate it.

Besides what I already commented, when I ran simulation I noticed 3 warnings of the form: `"WARNING: [Wavedata 42-489] Can't add object "/tb_tf_top/VMSME_L3PHIC17to24n1_data_arr" to the wave window because it has 6963200 bits, which exceeds the display limit of 65536 bits.  To change the display limit, use the command "set_property display_limit <new limit> [current_wave_config]".` Not surprising as we switched to using large arrays. I think we should increase the display limit like the warning suggests so these waves can be displayed in the window.

tomalin:
> Thanks a lot Ian for such diligent work at simplifying and perfecting the test-bench! You clearly have put a lot of thoughts and efforts into this. I really appreciate it.
> 
> Besides what I already commented, when I ran simulation I noticed 3 warnings of the form: `"WARNING: [Wavedata 42-489] Can't add object "/tb_tf_top/VMSME_L3PHIC17to24n1_data_arr" to the wave window because it has 6963200 bits, which exceeds the display limit of 65536 bits. To change the display limit, use the command "set_property display_limit <new limit> [current_wave_config]".` Not surprising as we switched to using large arrays. I think we should increase the display limit like the warning suggests so these waves can be displayed in the window.

I agree that we could do that. These arrays are so big, because they are storing all 100 events, (which I'm not sure is really necessary). It's not too bad if we can't view them, as they are copied to the smaller signals VMSME_L3PHIC17to24n1_dataarray_data_V_din etc. before being passed into SectorProcessor.vhd.

IntegrationTests/common/script/impl.tcl
tomalin: Everything you say at true. Though presumably, we'll need to run implementation when we run the chain in the Apollo, so even if people don't use this script now, it's perhaps useful to have it there for the future? And it documents the useful Vivado implementation options.

rzouCERN:
> > Thanks a lot Ian for such diligent work at simplifying and perfecting the test-bench! You clearly have put a lot of thoughts and efforts into this. I really appreciate it.
> > Besides what I already commented, when I ran simulation I noticed 3 warnings of the form: `"WARNING: [Wavedata 42-489] Can't add object "/tb_tf_top/VMSME_L3PHIC17to24n1_data_arr" to the wave window because it has 6963200 bits, which exceeds the display limit of 65536 bits. To change the display limit, use the command "set_property display_limit <new limit> [current_wave_config]".` Not surprising as we switched to using large arrays. I think we should increase the display limit like the warning suggests so these waves can be displayed in the window.
> 
> I agree that we could do that. These arrays are so big, because they are storing all 100 events, (which I'm not sure is really necessary). It's not too bad if we can't view them, as they are copied to the smaller signals VMSME_L3PHIC17to24n1_dataarray_data_V_din etc. before being passed into SectorProcessor.vhd.

Oh NVM I assumed these are input data arrays for a single event. So we use huge arrays instead of large rams to store our input data in this tb (why don't we just use rams? not sure if there's any difference effectively)... I agree we don't need to display them in the window then.

IntegrationTests/common/script/impl.tcl
rzouCERN: Yes we need post-implementation timing report when we run the chain on Apollo. For the baseline FW we already do all the reports here: `https://github.com/apollo-lhc/CM_FPGA_FW/blob/develop/scripts/Build.tcl`.

IntegrationTests/common/script/synth.tcl
tomalin: It's true that synthesis timing reports are approximate. Although I find them useful, as they can give a hint as to whether things are going well, before one takes the time of running implementation.

tomalin:
> > > Thanks a lot Ian for such diligent work at simplifying and perfecting the test-bench! You clearly have put a lot of thoughts and efforts into this. I really appreciate it.
> > > Besides what I already commented, when I ran simulation I noticed 3 warnings of the form: `"WARNING: [Wavedata 42-489] Can't add object "/tb_tf_top/VMSME_L3PHIC17to24n1_data_arr" to the wave window because it has 6963200 bits, which exceeds the display limit of 65536 bits. To change the display limit, use the command "set_property display_limit <new limit> [current_wave_config]".` Not surprising as we switched to using large arrays. I think we should increase the display limit like the warning suggests so these waves can be displayed in the window.
> > 
> > 
> > I agree that we could do that. These arrays are so big, because they are storing all 100 events, (which I'm not sure is really necessary). It's not too bad if we can't view them, as they are copied to the smaller signals VMSME_L3PHIC17to24n1_dataarray_data_V_din etc. before being passed into SectorProcessor.vhd.
> 
> Oh NVM I assumed these are input data arrays for a single event. So we use huge arrays instead of large rams to store our input data in this tb (why don't we just use rams? not sure if there's any difference effectively)... I agree we don't need to display them in the window then.

These huge arrays that store all events are only used in the test-bench. So I'm not sure if Vivado bothers defining which kind of FPGA storage should be used for them? Their large size is only an issue because of the memory usage on the PC, and because of this wave display issue. And neither problems seems important at present.

README.md
```diff
@@ -27,7 +27,8 @@ This would create a project directory \<project> ("projrouter" in case of the ab
 4) vivado -mode batch -source runProject.tcl (runs Vivado simulation,
```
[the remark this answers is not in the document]
tomalin: Agreed. Now fixed.

rzouCERN:
> > > > Thanks a lot Ian for such diligent work at simplifying and perfecting the test-bench! You clearly have put a lot of thoughts and efforts into this. I really appreciate it.
> > > > Besides what I already commented, when I ran simulation I noticed 3 warnings of the form: `"WARNING: [Wavedata 42-489] Can't add object "/tb_tf_top/VMSME_L3PHIC17to24n1_data_arr" to the wave window because it has 6963200 bits, which exceeds the display limit of 65536 bits. To change the display limit, use the command "set_property display_limit <new limit> [current_wave_config]".` Not surprising as we switched to using large arrays. I think we should increase the display limit like the warning suggests so these waves can be displayed in the window.
> > > 
> > > 
> > > I agree that we could do that. These arrays are so big, because they are storing all 100 events, (which I'm not sure is really necessary). It's not too bad if we can't view them, as they are copied to the smaller signals VMSME_L3PHIC17to24n1_dataarray_data_V_din etc. before being passed into SectorProcessor.vhd.
> > 
> > 
> > Oh NVM I assumed these are input data arrays for a single event. So we use huge arrays instead of large rams to store our input data in this tb (why don't we just use rams? not sure if there's any difference effectively)... I agree we don't need to display them in the window then.
> 
> These huge arrays that store all events are only used in the test-bench. So I'm not sure if Vivado bothers defining which kind of FPGA storage should be used for them? Their large size is only an issue because of the memory usage on the PC, and because of this wave display issue. And neither problems seems important at present.

Yeah I'm not asking what types of resources they would end up on chip (I would guess dram). It's just that since we don't need to worry about resources in testbench we can also simply use very deep tf_mem's instead of all these loops to load 100 events from text files  and that naively seems simpler to me...

IntegrationTests/common/script/impl.tcl
rzouCERN: I think it's a complete waste of time to run implementation at this stage. We should aim to make our workflow as simple as possible and not introduce unnecessary steps. I fail to see any justification to put this here. Maybe others can chime in.

tomalin: I'd like to merge this PR soon, so everyone can work with this simplified TB. Is your only remaining concern impl.tcl? I'd like to keep this file, as I think it may be useful in the future. But would it satisfy you if we either: (1) Explain in README.md that running it is not required; or (2) Remove all mention of it from README.md?

tomalin: @rzouCERN -- I've just made mod to README.md agreed at meeting, making clear impl.tcl is optional, and not required for code validation.

rzouCERN: thanks! Looks good to me.

rzouCERN (approved):

tomalin:
I just update my branch with all recent changes to master.

PR #150, opened 2021-05-11
Combined VMR

meisonlikesicecream:
Here's a VMR for the combined modules. that passed the majority of tests (one phi region in L1-D1 each). 

Pipelined with II=1, and rewinds alright. For some layers/phi regions it's slightly above the timing limit (4.66 ns vs 4 ns), but it's alright post-implementation. Resource usage is in the same order of magnitude as for the normal VMR.

Sometimes fails simulation due to the testvectors:
1. Testvectors sometimes contain >108 stubs
2. In the disks, the HLS assumes IL stubs are being read in the order PS, negPS, 2S, neg2S, but in the testvectors they are in the order PS, 2S, negPS, neg2S.

It could potentially also fail simulations because I only allow 15 stubs per bin in the TEOuter memories instead of 16, due to nentries only using 4 bits per bin (see https://github.com/cms-L1TK/firmware-hls/blob/vmr_cm/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L135). Otherwise the bin would overflow and overwrite other stubs. I tried increasing the number of bits to 5, but that made TrackletProcessor sad and I couldn't figure out how to change the TP and TrackletEngineUnit.h to make it happy.

### DIFF
--- a/TestBenches/VMRouterCM_test.cpp
+++ b/TestBenches/VMRouterCM_test.cpp
@@ -1,0 +1,153 @@
+// Test bench for VMRouter Combined Modules
+#include "VMRouterTopCM.h"
+
+#include <algorithm>
+#include <iterator>
+
+#include "FileReadUtility.h"
+
+using namespace std;
+
+const int nEvents = 100;  //number of events to run
+
+// VMRouter Test that works for all regions
+// Sort stubs into smaller regions in phi, i.e. Virtual Modules (VMs).
+
+// NOTE: to run a different phi region, change the following
+//          - the base directory when instantiating TBHelper in VMRouterCM_test.cpp
+//          - the included top function in VMRouterCM_test.cpp (if file name is changed)
+//          - the top function and memory directory in script_VMR_CM.tcl (if file name is changed)
+//          - and the changes listed in VMRouterTopCM.cc/h
+
+int main() {
+
+  ////////////////////////////////////////////////////////////////
+  // Get lists of the input/output memory directory and file names
+  // I.e. the names of the test vector files
+
+  TBHelper tb("VMRCM/VMR_L2PHIA");
+
+  // String patterns of the memory file names
+  const string inputPattern = (kLAYER) ? "InputStubs*" : "InputStubs*PS*";
+  const string inputDisk2SPattern = "InputStubs*_D*2S*";
+  
+  const string allStubPattern = "AllStubs*PHI?_*";  // need "PHI?_*" because of bug in the MemPrintsCM?
+  const string allStubInnerPattern = "AllInnerStubs*";
+  const string mePattern = "VMStubs_VMSME*";
+  const string tePattern = "VMStubs_VMSTE*";
+  
+  // Number of files
+  const auto nInputStubs = tb.nFiles(inputPattern);
+  const auto nInputStubsDisk2S = tb.nFiles(inputDisk2SPattern);
+
+  const auto nASCopies = tb.nFiles(allStubPattern);
+  const auto nASInnerCopies = tb.nFiles(allStubInnerPattern);
+  const auto nVMSME = tb.nFiles(mePattern);
+  const auto nVMSTE = tb.nFiles(tePattern);
+  
+  // Make sure that the number of input and output memories are correct
+  assert((nInputStubs == numInputs) && (nInputStubsDisk2S == numInputsDisk2S) && (nASCopies == maxASCopies));
+  
+  // Open the files
+  cout << "Open files..." << endl;
+
+  auto &fin_inputstubs = tb.files(inputPattern);
+  auto &fin_inputstubs_disk2s = tb.files(inputDisk2SPattern);
+
+  auto &fout_allstubs = tb.files(allStubPattern);
+  auto &fout_allstubs_inner = tb.files(allStubInnerPattern);
+  auto &fout_vmstubme = tb.files(mePattern);
+  auto &fout_vmstubte = tb.files(tePattern);
+
+  ///////////////////////////
+  // Declare memories
+
+  // Input memories
+  static InputStubMemory<inputType> inputStubs[numInputs];
+  static InputStubMemory<DISK2S> inputStubsDisk2S[numInputsDisk2S]; //Only used for Disks
+
+  // output memories
+  static AllStubMemory<outputType> memoriesAS[maxASCopies];
+  static AllStubInnerMemory<outputType> memoriesASInner[maxASInnerCopies];
+  // ME memories
+  static VMStubMEMemoryCM<outputType, rzSize, phiRegSize> memoryME;
+  // TE Outer memories
+  static VMStubTEOuterMemoryCM<outputType,rzSize,phiRegSize,numTEOCopies> memoryTEO;
+
+  ///////////////////////////
+  // loop over events
+  
+  // error count
+  int err = 0;
+  
+  cout << "Start event loop ..." << endl;
+
+  for (unsigned int ievt = 0; ievt < nEvents; ++ievt) {
+    cout << "Event: " << dec << ievt << endl;
+
+    // clear output memories
+    for (int i=0; i<nASCopies; ++i) {
+      memoriesAS[i].clear();
+    }
+    for (int i=0; i<nASInnerCopies; ++i) {
+      memoriesASInner[i].clear();
+    }
+    memoryME.clear();
+    if (nVMSTE) {
+      memoryTEO.clear();
+    }
+
+    // read event and write to memories
+    for (unsigned int i = 0; i < numInputs; i++) {
+      writeMemFromFile<InputStubMemory<inputType>>(inputStubs[i], fin_inputstubs[i], ievt);
+    }
+    for (unsigned int i = 0; i < numInputsDisk2S; i++) {
+      writeMemFromFile<InputStubMemory<DISK2S>>(inputStubsDisk2S[i], fin_inputstubs_disk2s[i], ievt);
+    }
+
+    // bx - bunch crossing
+    BXType bx = ievt;
+    BXType bx_out;
+
+    // Unit Under Test
+    VMRouterTopCM(bx, bx_out, inputStubs
+#if kDISK > 0
+        , inputStubsDisk2S
+#endif
+        , memoriesAS, memoriesASInner//, maskASI
+        , memoryME
+#if kLAYER == 2 || kLAYER == 3 || kLAYER == 4 || kLAYER == 6 || kDISK == 1 || kDISK == 2 || kDISK == 4
+        , memoryTEO
+#endif
+      );
+
+    // compare the computed outputs with the expected ones
+    // add 1 to the error count per stub that is incorrect
+
+    bool truncation = false;
+
+    // AllStub memories
+    for (unsigned int i = 0; i < maxASCopies; i++) {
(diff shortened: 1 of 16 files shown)


TestBenches/VMRouterCM_test.cpp
```diff
@@ -0,0 +1,158 @@
...
+    //TE Outer memories
+    if (nVMSTE) {
+      err += compareBinnedMemCMWithFile<VMStubTEOuterMemoryCM<outputType, rzSizeTE, phiRegSize, numTEOCopies>>(memoryTEO, fout_vmstubte[0], ievt, "VMStubTEOuter", truncation);
+    }
+  } // End of event loop
+
+	cerr << "Exiting with return value " << err << endl;
+	// This is necessary because HLS seems to only return an 8-bit error count, so if err%256==0, the test bench can falsely pass
+	if (err > 255) err = 255;
+	return err;
```
aehart, line 155: Please move these lines to the left so they are inline with the bracket at the end of the event loop.

TrackletAlgorithm/Constants.h
```diff
@@ -81,22 +95,17 @@ constexpr double kphiprojderdisk = 1.64665e-05;
...
 // cut constants
 constexpr double ptcut = 1.91; // GeV
 constexpr double rinvcut = 0.01 * c * bfield / ptcut; // 0.01 to convert to 1/cm
 constexpr double z0cut = 15.0; // cm
 
+// cut constants for combined modules
+constexpr double VMROUTERCUTZL2 = 50.0;      //Min L2 z for inner allstub
+constexpr double VMROUTERCUTZL1L3L5 = 95.0;  //Max z for inner barrel layers
+constexpr double VMROUTERCUTZL1 = 70.0;      //Max z for L1 barrel seeding
+constexpr double VMROUTERCUTRD1D3 = 55.0;    //Max r for disk seeds
```
aehart, line 107: Would you mind adding units (I assume it's cm for all of these) in the comments, similar to those for the floating-point detector constants?

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -111,14 +121,18 @@ class MemoryTemplateBinnedCM{
   
   bool write_mem(BunchXingT ibx, ap_uint<NBIT_BIN> slot, DataType data) {
 #pragma HLS ARRAY_PARTITION variable=dataarray_ dim=1
+#pragma HLS ARRAY_PARTITION variable=nentries8_ complete dim=0
+#pragma HLS ARRAY_PARTITION variable=binmask8_ complete dim=0
```
aehart, line 125: For the standard modules, we have eliminated `nentries` ports associated with output memories, but these two data members result in a bunch of input and output ports associated with `memoryTEO`. I know you didn't introduce these data members, but would it be possible to remove these ports like the `nentries` ports were removed in PR #119?

TrackletAlgorithm/VMRouterCM.h
```diff
@@ -0,0 +1,434 @@
...
+			(Layer) ? nbitsrtablelayer : nbitsrtabledisk; // Number of MSBs of r used in LUT table
+	constexpr auto vmbits =
+			(Layer) ? nbitsvmlayer[Layer - 1] : nbitsvmdisk[Disk - 1]; // Number of bits for standard VMs
+	constexpr unsigned int nbitsall = (Layer) ? nbitsallstubs[Layer-1] : nbitsallstubs[N_LAYER+Disk-1]; // Number of bits for the number of Alltub memories in a layer/disk
+
+	// Set values to VMStub
+	stub.setBend(bend);
+	stub.setIndex(index);
+	stub.setFinePhi(
+				iphivmFineBins<InType>(phicorr, nbitsall + vmbits, nbitsfinephi));
```
aehart, line 126: Could you please go through this file and make sure the indentation is consistent and reasonable? For example, I'm not sure why this is on a separate line and indented so much.

TrackletAlgorithm/VMRouterCM.h
```diff
@@ -0,0 +1,434 @@
...
+	// Coarse z. The bin the stub is going to be put in, in the memory
+	int bin = lutValue >> nbitsfinerz; // 3 bits, i.e. max 8 bins within each VM
+
+	if (negDisk) {
+		if (isMEStub) bin += 1 << MEBinsBits; // bin 8-16 are for negative disk
+		else bin += (1 << TEBinsBits)/2; // += 4
+	}
+	
+	auto ivm = phicorr.range(phicorr.length() - nbitsall - 1, phicorr.length() - (nbitsall + vmbits)); //get the phi bits that corresponds to the old vms
+	slot = (Disk && isMEStub) ? ivm * 16 + bin : ivm * 8 + bin; //1 << 3 is the number of bins NBINS?
```
aehart, line 163: Is there a way to write this without the magic 16 and 8? Like instead of `ivm * 16`, you could have something like `(ivm << NBITS)`, where `NBITS` is some named constant equal to 4, and similarly for `ivm * 8`.

TrackletAlgorithm/VMRouterCM.h
```diff
@@ -0,0 +1,434 @@
...
+								passPhiCut = !(iphipos>=phicutmin);
+							} else if (n == 8 || n==11) { // R
+								passPhiCut = !(iphipos<phicutmax);
+							} else if (n == 0 || n==3 || n==5) { // A, D, F
+								passPhiCut = !(iphipos<4);
+							} else if (n == 1 || n==2 || n==4) { // B, C, E
+								passPhiCut = !(iphipos>=4);
+							} else { // M
+								passPhiCut = true;
+							}
```
aehart, line 364: There are several magic numbers here related to the index into `maskASI`. And based on your comments, I guess this index corresponds to the letters in the names of the AS memories? If that's the case, then would it make sense to define an `enum` somewhere so that you can use letters instead of magic numbers in these comparisons?

TrackletAlgorithm/VMRouterCMTop.h
```diff
@@ -0,0 +1,83 @@
...
+
+////////////////////////////////////////////
+// Variables for that are specified with regards to the VMR region
+// Changed manually
+
+#define kLAYER 2 // Which barrel layer number the data is coming from
+#define kDISK 0 // Which disk number the data is coming from, 0 if not disk
+
+constexpr char phiRegion = 'A'; // Which AllStub/PhiRegion
+constexpr int sector = 4; //  Specifies the sector
```
aehart, line 23: I think this should be moved to the test bench, since it doesn't have anything to do with defining the top function.

TrackletAlgorithm/VMRouterCMTop.h
```diff
@@ -0,0 +1,83 @@
...
+
+// Maximum number of memory "copies" for this Phi region
+constexpr int numASCopies(4); // Allstub memory
+constexpr int numASInnerCopies(4); // Allstub memory Note: can't use 0 if we don't have any memories of a certain type. Use 1.
+constexpr int numTEOCopies(3); // TE Outer memories, can be 0 when no TEOuter memories
+
+// Number of inputs
+constexpr int numInputs(2); // Number of input memories, EXCLUDING DISK2S
+constexpr int numInputsDisk2S(0); // Number of DISK2S input memories
```
aehart, line 32: I guess these are not actually free parameters, but instead are functions of the layer/disk and phi region. So would it make sense to define them as `constexpr` functions that take in the layer/disk and phi region as arguments or template parameters? You could define the functions for just VMRCM_L2PHIA for now, and then expand it to other instances of the combined VMR in the future, but this way, only a couple parameters would need to change in order to switch between VMR instances.

TrackletAlgorithm/VMRouterCMTop.cc
```diff
@@ -0,0 +1,169 @@
...
+	#pragma HLS resource variable=inputStubs[0].get_mem() latency=2
+	#pragma HLS resource variable=inputStubs[1].get_mem() latency=2
+	// #pragma HLS resource variable=inputStubs[2].get_mem() latency=2
+	// #pragma HLS resource variable=inputStubs[3].get_mem() latency=2
+
+	// Masks of which memories that are being used. The first memory is represented by the LSB
+	// and a "1" implies that the specified memory is used for this phi region
+	// First three bits (LSB) are the six A-F for Barrel, then the three after that are L,M,R for Barrel and disk, last three are L,M,R for Overlap
+	// NOTE: read from right to left (OR, OM, OL, BR/DR, BM/DM, BL/DL, BF, BE, BD, BC, BB, BA)
+	static const ap_uint<maskASIsize> maskASI = 0b110110000000; // Change me when switching phi region
```
aehart: As with the parameters in VMRouterCMTop.h, would it make sense to also define this as a `constexpr` function, since you indicate that it depends on the phi region?

project/script_VMRCM.tcl
```diff
@@ -0,0 +1,34 @@
...
+open_solution "solution1"
+
+# Define FPGA, clock frequency & common HLS settings.
+source settings_hls.tcl
+
+# data files
+add_files -tb ../emData/VMRCM/tables/
+add_files -tb ../emData/VMRCM/
+
+csim_design -compiler gcc -mflags "-j8"
```
aehart, line 27: Please remove the `-compiler` option. Apparently it isn't supported in future versions of Vitis HLS, and we want to future-proof our code when it's easy.

TrackletAlgorithm/VMRouterCMTop.cc
aehart: I see you moved this to the header file, but I think the question is still relevant.

TrackletAlgorithm/VMRouterCMTop.h
meisonlikesicecream: Not quite sure I understand. Do you want a function that computes the number of copies somehow, or more like a LUT? Because I have no idea how one determines how many memories one needs, I just look at the wiring (and/or what was being downloaded)...

aehart: I mean something like what's done for the TP in TrackletProcessor_parameters.h or for the TC in TrackletCalculator_parameters.h. These functions aren't doing any real computation, they just return values that are hard-coded based on the wiring. But the point is that once these functions are eventually set up for all instances, only the layer/disk and phi region would have to be changed, and not all these additional constants.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
meisonlikesicecream: Eliminating the nentries ports reduced the timing after synthesis, so all of them meets timing now. Hurrah!

meisonlikesicecream:
@aehart I think I've addressed all your comments now. I've tried the updated version for L1-L6 phi A with success, and all layers meet timing after synthesis now. The disks still fail simulation due to the testvectors (and potentially some other things that I might've missed because it's so hard to check when there are loads of inconsistencies).

TrackletAlgorithm/MemoryTemplateBinnedCM.h
aehart: That's great!

aehart (approved):
@meisonlikesicecream I don't think you need to add additional instances to download.sh yet. Eventually, we'll probably want to test multiple instances, e.g., when we're ready to integrate this module in a chain to test on hardware. But for now, it's fine to test just one by default.

This looks really good now. Feel free to click the merge button when you're ready.

meisonlikesicecream:
@aehart Okay, thanks!

PR #154, opened 2021-05-24
VMR Testbench Update

meisonlikesicecream:
Updated the VMR testbench using the TBHelper class. Added a VMR parameters file so one only needs to change kLAYER/kDISK and phiRegion in VMRouterTop.h to run a different VMR (given that the necessary numbers have been defined in the parameters file). Removed the D1PHIA top files as they are now obsolete and no longer needed thanks to the parameters file. 

Changed the default VMR to VMR_L2PHIA as layer 2 includes all types of memories (and it happens to be a part of the simplest IR-VMR chain).


project/script_VMR.tcl
```diff
@@ -22,8 +22,7 @@ source settings_hls.tcl
 
 # data files
 add_files -tb ../emData/VMR/tables/
```
aehart, line 24: I believe this line can also be removed.

TrackletAlgorithm/VMRouter_parameters.h
```diff
@@ -0,0 +1,1565 @@
...
+}
+template<> inline const int* getRzBitsOuterTable<false, TF::D3>(){
+  return nullptr;
+}
+// Uncomment when needed
+// template<> inline const int* getRzBitsOuterTable<false, TF::D4>(){
+//   static int lut[] =
+// #include "../emData/VMR/tables/VMTableOuterD4.tab"
+//   return lut;
+// }
```
aehart, line 207: Is there a reason for leaving this particular specialization commented out?

TrackletAlgorithm/VMRouter_parameters.h
```diff
@@ -0,0 +1,1565 @@
...
+template<bool isLayer, int layer, phiRegions phi, int size> const ap_uint<size>* getBendCutOuterTable();
+
+template<bool isLayer, int layer, phiRegions phi> constexpr int getNumInputs();
+template<bool isLayer, int layer, phiRegions phi> constexpr int getNumInputsDisk2S();
+template<bool isLayer, int layer, phiRegions phi> constexpr int getNumASCopies();
+template<bool isLayer, int layer, phiRegions phi> constexpr int getNumTEICopies();
+template<bool isLayer, int layer, phiRegions phi> constexpr int getNumOLCopies();
+template<bool isLayer, int layer, phiRegions phi> constexpr int getNumTEOCopies();
+
+template<bool isLayer, int layer, phiRegions phi> constexpr int getBendCutTableSize();
```
aehart, line 27: The `isLayer` parameter seems to be redundant, since it's completely determined by `layer`. Can it be removed?

TrackletAlgorithm/VMRouterTop.h
```diff
@@ -2,46 +2,44 @@
...
-constexpr int maxTEOCopies(1); // TE Outer memories
 
-// Number of inputs
-constexpr int numInputs(7); // Number of input memories, EXCLUDING DISK2S
-constexpr int numInputsDisk2S(0); // Number of DISK2S input memories
+///////////////////////////////////////////////
+// Variables that don't need manual changing
 
-constexpr int bendCutTableSize(8); // Number of entries in each bendcut table. Can't use 0.
+constexpr bool isLayer = (kLAYER) ? true : false;
```
aehart, line 29: If the parameter in VMRouter_parameters.h is indeed redundant, then I think this can be removed.

TestBenches/VMRouter_test.cpp
```diff
@@ -15,245 +14,112 @@ const int nEvents = 100;  //number of events to run
...
-  if (not valid) {
-    cout << "Could not find wiring file." << endl;
-    return false;
+// Count the number of copies of each memory in a file name vector
+vector<int> countCopies(vector<string> fileNames) {
+  vector<int> numCopies; // Each element in vector corresponds to one memory, i.e. L1PHIE17
+  string firstCopyIndex = fileNames.begin()->substr(fileNames.begin()->find("n")+1, 1); // The number after n
+  for (auto f = fileNames.begin(); f != fileNames.end(); ++f) {
+    if (f->find("n"+firstCopyIndex) != string::npos) numCopies.push_back(1);
+    else numCopies.back() += 1;
```
aehart, line 29: Could this function be useful in any other test benches? If so, you could convert it to a method of the `TBHelper` class.

TrackletAlgorithm/VMRouter_parameters.h
meisonlikesicecream: Ah, I didn't add any VMR_D4 in the download script, so it complained about the file no existing. I could uncomment this and add VMR_D4PHIA (or any other phi region) to the download script if that's better?

project/script_VMR.tcl
meisonlikesicecream: removed!

TrackletAlgorithm/VMRouter_parameters.h
aehart: No, I don't think you need to add any D4 VMRs to download.sh right now. I was just curious why this particular one was commented out, but your explanation makes sense.

TrackletAlgorithm/VMRouterTop.h
meisonlikesicecream: see reply above

TestBenches/VMRouter_test.cpp
meisonlikesicecream: I don't think so. I need it because I use two-dimensional arrays of memories, but no other module uses that. And what those two dimensions correspond to seems quite specific to the VMR.

TrackletAlgorithm/VMRouter_parameters.h
aehart: Ah I see the issue, and in fact, right now `TF::L1` and `TF::D1` are both equal to zero.

I would propose creating a new enum in the `TF` namespace in Constants.h that includes both layers and disks. Something like:
```c++
  enum layerDisk {L1 = 0, L2 = 1, L3 = 2, L4 = 3, L5 = 4, L6 = 5, D1 = 6, D2 = 7, D3 = 8, D4 = 9, D5 = 10};
```
We can keep the `layer` and `disk` enums for now, but eventually they would be phased out in favor of `layerDisk` (or whatever you choose to name it). This way, we eliminate the ambiguity between the disks and some of the layers, and you can also use `TF::layerDisk` as the type for the template parameter in your functions, instead of the generic `int`.

TestBenches/VMRouter_test.cpp
aehart: OK, makes sense. Then could you please change `vector<string> fileNames` to `const vector<string> &fileNames`? That will prevent an unnecessary copy of the input vector.

TrackletAlgorithm/VMRouter_parameters.h
meisonlikesicecream: I removed the layer and disk enums as we can't have multiple L1s etc. Only the MC was using them, so I fixed that. Hope that was okay :-)

aehart: Perfect.

I realized this morning though that this change _should_ also break the TC and TP because `TF::D1`, `TF::D2`, etc. are used as indices into `zmean`. Unfortunately, this isn't reflected in the results of the CI tests because both modules have bugs in their respective test benches. I'll open a separate PR to fix these bugs, but I just added a commit to your branch that extends `zmean` and `rmean` so that the TC and TP work again.

meisonlikesicecream: Ah okay thanks, I missed that

aehart (approved):
Looks good to me now. Approved!

PR #157, opened 2021-05-26
TC/TP test bench bug fixes

aehart:
The TC test bench currently does not check the TPROJ memories for disks, and the TP test bench always returns zero.

This PR fixes these bugs.


tomalin (approved):
Looks fine. And I checked it works for TC.

We'll ultimately want to be able to compile any given HLS IP in chain. As pragmas specified in .tcl are ignored during synthesis, it's unclear if the current solution is viable long-term. Possible alternative: Define short TC_pragmas.h file, where pragmas are defined, to be included into other TC code, and overwrite this file via .tcl commands to change pragmas defining which specific TC is to be instantiated?

PR #158, opened 2021-05-28
Add tf_lut.vhd and latency_monitor.vhd to IntegrationTests/common/hdl/

rzouCERN:
Adding two small pieces of vhd code to IntegrationTests/common/hdl/. tf_lut.vhd is a simple look-up table that uses distributed ram as resources. latency_monitor.vhd counts clock cycles between any triggering signal such as ap_done and compares it with a set value. It might be useful in debugging clock frequency related errors.  

tf_lut.vhd synthesizes in a te-tc top level created by project_generation_script. latency_monitor.vhd functionality tested in current testbench.


IntegrationTests/common/hdl/latency_monitor.vhd
```diff
@@ -0,0 +1,66 @@
+-------------------------------------------------------------------------------
+-- Counter to count clock cycles in between ap_done
```
tomalin, line 2: Please clarify comment "Counter to count clock cycles between two successive ap_done pulses". And add comment explaining typical use for this application.

IntegrationTests/common/hdl/latency_monitor.vhd
```diff
@@ -0,0 +1,66 @@
...
+        if (error = '1') then
+          error_sticky <= '1';
+        end if;
+        if (ap_done = '1') then
+          count_start <= '1';
+          latency_count <= 1;
+          if (count_start = '1') then
+            latency <= latency_count;
+            if not (latency = latency_expected) then
+                error <= '1';
```
tomalin, line 52: Consider throwing a fatal error with a VHDL "assert" statement, instead of just returning error=1.

rzouCERN: Done. I've added a default "assert" statement and an option to mute it.

IntegrationTests/common/hdl/latency_monitor.vhd
rzouCERN: Done.

IntegrationTests/common/hdl/latency_monitor.vhd
tomalin: Is there a reason you don't set the severity level to "failure"? This would mean no output ports were needed.

rzouCERN: I would like for this to be run on board as well not just in simulation. The "assert" statement is just for simulation.

tomalin (approved):
Looks useful.

PR #159, opened 2021-05-28
TPAR BX bits

aehart:
This PR increases the number of bits used for the BX in TPAR memories to three. This is needed since these memories are filled by the TC, but are not read until the TrackBuilder/PurgeDuplicates.


rzouCERN (approved):
thanks a lot for the quick fix!

Issue #146, opened 2021-04-15
Small bug in PR-ME-MC VHDL test-bench


tomalin:
In IntegrationTests/PRMEMC/tb/tb_tf_top.vhd , the code following the comment "FIX: This is buggy. As mem read takes 2 clks, it will lose last 2 samples" is buggy.

It means that only the first 106 out of total 108 FM memory data entries will be read by the test-bench, which may lead to inconsistencies with the emData/ predictions for events with more than 106 valid FM entries. This seems to happen only rarely.

tomalin:
Fixed by https://github.com/cms-L1TK/firmware-hls/pull/160

PR #160, opened 2021-06-01
Add FileWriter*.vhd utilities to simplify L1Trk chain test-benches

tomalin:
Simplified .txt file writing of results from VHDL test-bench by:
1) Adding new utility common/hdl/FileWriter.vhd to write to .txt file results from an intermediate BRAM memory in the L1Trk chain.
2) Adding new utility common/hdl/FileWriterFromRAM.vhd to write to .txt file results from an unbinned BRAM memory at the end of the L1Trk chain.
3) Small changes to CompareMemPrintsFW.py to handle simplified column headings in .txt files.
4) Resolves issue https://github.com/cms-L1TK/firmware-hls/issues/146

Validation: cmp.txt files produced by CompareMemPrintsFW.py are unchanged. And Vivado simulation running much faster.

### DIFF
--- a/IntegrationTests/common/script/CompareMemPrintsFW.py
+++ b/IntegrationTests/common/script/CompareMemPrintsFW.py
@@ -112,7 +112,8 @@ def compare(comparison_filename="", fail_on_error=False, file_location='./', pre
             raise IndexError("Unable to determine the reference file type from the file name")
         finally:
             reference_type = ReferenceType[reference_type_string]
-        print("Comparing " + reference_type.FullName() + " values ... ")
+
+        print("Comparing TB results to ref. file "+str(reference_filename)+" ... ")
 
         # Parse the reference data
         reference_data = parse_reference_file(file_location+"/"+reference_filename)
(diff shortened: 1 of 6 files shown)


IntegrationTests/common/script/CompareMemPrintsFW.py
```diff
@@ -123,48 +114,26 @@ def compare(comparison_filename="", fail_on_error=False, file_location='./', pre
             reference_type = ReferenceType[reference_type_string]
         print("Comparing " + reference_type.FullName() + " values ... ")
 
-        # Find the layer names for the reference file
-        if reference_type == ReferenceType.FM:
-            layers = re.search("(L[0-9]L[0-9])",reference_filename).group(0)
-            print("Comparing the values for layers "+str(layers)+" to the reference file "+str(reference_filename)+" ... ")
```
aehart, line 129: I found it useful to have the reference filename printed out. Could we keep that, and actually print it out for all memory types, not just FullMatches?

tomalin: Fixed. It now prints out a message such as "Comparing TB results to ref. file ../../../emData/MemPrints/TrackletProjections/AllProj_AP_L3PHIC_04.dat ..." before each comparison with a new file.

aehart (approved):
Looks very good to me.

PR #161, opened 2021-06-04
Set nentries when running in CMSSW

EmyrClement:
We need the nentries variable to be set for the memories when running inside CMSSW.  As we are also replicating the testbenches in CMSSW (using inputs from text files rather than from the CMSSW event), we have to avoid incrementing nentries twice when reading from a text file.  Following a short email exchange with @aehart, using the ```CMSSW_GIT_HASH``` was suggested as the way to implement this.

I implemented the changes for ```MemoryTemplate.h``` and ```MemoryTemplateBinned.h``` so far.  It wasn't obvious to me how to implement the same changes for ```MemoryTemplateBinnedCM.h``` (for both nentries 8 and binmask8), so I would appreciate some guidance there.

I also added a few preprocessor guards around the ```cout``` statements in the VMR code so they don't appear when running in CMSSW.

### DIFF
--- a/TrackletAlgorithm/MemoryTemplate.h
+++ b/TrackletAlgorithm/MemoryTemplate.h
@@ -60,6 +60,11 @@ public:
 #pragma HLS inline
     if (addr_index < (1<<NBIT_ADDR)) {
       dataarray_[ibx][addr_index] = data;
+      
+      #ifdef CMSSW_GIT_HASH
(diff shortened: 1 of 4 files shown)


aehart:
@pwittich Could you please add @EmyrClement to HLS-experts so that he can push his branch to the main repo?

PR #162, opened 2021-06-04
CI updates

aehart:
Turns on the REBASE_MASTER option and moves the tarball download to a separate stage, as discussed in the HLS chat here:
https://indico.cern.ch/event/1039221/#2-ci-updates


pwittich:
Nice optimization.

pwittich (approved):

PR #163, opened 2021-06-07
Set nentries when running in CMSSW

EmyrClement:
We need the nentries variable to be set for the memories when running inside CMSSW. As we are also replicating the testbenches in CMSSW (using inputs from text files rather than from the CMSSW event), we have to avoid incrementing nentries twice when reading from a text file. Following a short email exchange with @aehart, using the CMSSW_GIT_HASH was suggested as the way to implement this.

I implemented the changes for MemoryTemplate.h and MemoryTemplateBinned.h so far. It wasn't obvious to me how to implement the same changes for MemoryTemplateBinnedCM.h (for both nentries 8 and binmask8), so I would appreciate some guidance there.

I also added a few preprocessor guards around the cout statements in the VMR code so they don't appear when running in CMSSW.

### DIFF
--- a/TrackletAlgorithm/MemoryTemplate.h
+++ b/TrackletAlgorithm/MemoryTemplate.h
@@ -98,10 +98,10 @@ public:
 	int nent = nentries_[ibx]; 
 	bool success = write_mem(ibx, data, nent);
 
-  #ifndef CMSSW_GIT_HASH
-  if (success) nentries_[ibx] ++;
-  #endif
-  return success;
+	#ifndef CMSSW_GIT_HASH
+	if (success) nentries_[ibx] ++;
+	#endif
+	return success;
   }
 
   bool write_mem(BunchXingT ibx, const std::string datastr, int base=16)
@@ -110,10 +110,10 @@ public:
 	int nent = nentries_[ibx];
 	bool success = write_mem(ibx, data, nent);
 
-  #ifndef CMSSW_GIT_HASH
-  if (success) nentries_[ibx] ++;
-  #endif
-  return success;
+	#ifndef CMSSW_GIT_HASH
+	if (success) nentries_[ibx] ++;
+	#endif
+	return success;
   }
 
   // print memory contents
--- a/TrackletAlgorithm/MemoryTemplate.h
+++ b/TrackletAlgorithm/MemoryTemplate.h
@@ -60,6 +60,11 @@ public:
 #pragma HLS inline
     if (addr_index < (1<<NBIT_ADDR)) {
       dataarray_[ibx][addr_index] = data;
+      
+      #ifdef CMSSW_GIT_HASH
+        nentries_[ibx] = addr_index + 1;
+      #endif
+      
       return true;
     } else {
       return false;
@@ -92,7 +97,10 @@ public:
 	DataType data(datastr, base);
 	int nent = nentries_[ibx]; 
 	bool success = write_mem(ibx, data, nent);
+
+  #ifndef CMSSW_GIT_HASH
 	if (success) nentries_[ibx] ++;
+  #endif
         return success;
   }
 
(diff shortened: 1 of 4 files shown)


EmyrClement:
Replacing #161

tomalin:
At https://github.com/EmyrClement/firmware-hls/blob/nentriesForCMSSW/TrackletAlgorithm/MemoryTemplate.h#L102 nentries is now only incremented if running in CMSSW. Whereas in the current "master" it is always incremented. Same comment at 114. Doesn't this change the non-CMSSW behaviour?

EmyrClement:
> At https://github.com/EmyrClement/firmware-hls/blob/nentriesForCMSSW/TrackletAlgorithm/MemoryTemplate.h#L102 nentries is now only incremented if running in CMSSW. Whereas in the current "master" it is always incremented. Same comment at 114. Doesn't this change the non-CMSSW behaviour?

nentries is only incremented here when *not* running in CMSSW (it's an ```ifndef``` not an ```ifdef```).  I included these ```ifndef``` guards to avoid incrementing nentries by two when we read inputs from text files in CMSSW.

The non-CMSSW behaviour looks unchanged to me from checking the VMRCM csim/csynth/cosim.  It looks as if the CI has also passed.

tomalin:
> > At https://github.com/EmyrClement/firmware-hls/blob/nentriesForCMSSW/TrackletAlgorithm/MemoryTemplate.h#L102 nentries is now only incremented if running in CMSSW. Whereas in the current "master" it is always incremented. Same comment at 114. Doesn't this change the non-CMSSW behaviour?
> 
> nentries is only incremented here when _not_ running in CMSSW (it's an `ifndef` not an `ifdef`). I included these `ifndef` guards to avoid incrementing nentries by two when we read inputs from text files in CMSSW.
> 
> The non-CMSSW behaviour looks unchanged to me from checking the VMRCM csim/csynth/cosim. It looks as if the CI has also passed.

OK

aehart:
@EmyrClement Looks good to me. I would only ask that you make the indentation consistent. The convention in CMSSW seems to be to have the preprocessor directives all in the first column and the code within the `#ifdef`/#ifndef`…`#endif` blocks aligned as it would be if the directives were not there.

TrackletAlgorithm/MemoryTemplate.h
```diff
@@ -92,7 +97,10 @@ class MemoryTemplate
 	DataType data(datastr, base);
 	int nent = nentries_[ibx]; 
 	bool success = write_mem(ibx, data, nent);
+
+  #ifndef CMSSW_GIT_HASH
 	if (success) nentries_[ibx] ++;
+  #endif
         return success;
```
aehart, line 104: Could you also move `return success;` to the left two spaces?

TrackletAlgorithm/MemoryTemplate.h
```diff
@@ -101,7 +109,10 @@ class MemoryTemplate
 	DataType data(datastr.c_str(), base);
 	int nent = nentries_[ibx];
 	bool success = write_mem(ibx, data, nent);
+
+  #ifndef CMSSW_GIT_HASH
 	if (success) nentries_[ibx] ++;
+  #endif
         return success;
```
aehart, line 116: Could you also move `return success;` to the left two spaces?

EmyrClement:
The indentation/alignment should now be fixed.  I still need to work out what the correct treatment in MemoryTemplateBinnedCM.h should be.

aehart:
>  I still need to work out what the correct treatment in MemoryTemplateBinnedCM.h should be.

Ah sorry. I skimmed over that when reading the description. I think you just need to add the following two lines to `MemoryTemplateBinnedCM::write_mem`:
https://github.com/cms-L1TK/firmware-hls/blob/b4fe97a7e2842d48a35fbd80a61269ad92e5739b/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L129-L131
And of course, these should be within the `#ifdef CMSSW_GIT_HASH` guard that you added to the other classes.

EmyrClement:
Thanks - think I implemented this correctly, where I derive ```ireg``` and ```ibin``` from ```slot``` in the same way as in the other ```write_mem``` functions.  I've not yet tested this in CMSSW, but should do soon.

aehart (approved):
It looks correct to me, so I will approve the PR. Feel free to merge when the topPRMEMC-synth CI test finishes running.

PR #167, opened 2021-06-17
Future CMSSW changes for compilation with CMSSW_11_3

carriganm95:
This pull request has the changes needed by the future CMSSW emulation to compile and run the match processor. The changes made are

getName() and getSector() functions are added to MemoryTemplate.h, MemoryTemplateBinned.h, and MemoryTemplateBinnedCM.h
The functions getBitStr() and decodeToBits() have been added to AllStubsMemory.h, FullMatchMemory.h, TrackletProjectionsMemory.h, and VMStubsMEMemoryCM.h in order to allow the emulation to write out the memories to file.
Ifndef statements have been added around #include hls_math.h in order to exclude them when in CMSSW
Ifndef statements are added to Constants.h to prevent redefining of N_LAYERS and N_DISKS
Changed ints to unsigned int in MatchProcessor.h, MemoryTemplate.h, MemoryTemplateBinned.h, and MemoryTemplateBinnedCM.h

### DIFF
--- a/TrackletAlgorithm/AllStubMemory.h
+++ b/TrackletAlgorithm/AllStubMemory.h
@@ -186,15 +186,17 @@ public:
     data_.range(kASBendMSB,kASBendLSB) = bend;
   }
 
+#ifdef CMSSW_GIT_HASH
   std::string getBitStr() const {
-    std::string str = decodeToBits(getR(),AllStubBase<ASType>::kASRSize);
-    str += "|"+decodeToBits(getZ(),AllStubBase<ASType>::kASZSize);
-    str += "|"+decodeToBits(getPhi(),AllStubBase<ASType>::kASPhiSize);
+    std::string str = MemoryTemplate<AllStub<ASType>, 3, kNBits_MemAddr>::decodeToBits(getR(),AllStubBase<ASType>::kASRSize);
+    str += "|"+MemoryTemplate<AllStub<ASType>, 3, kNBits_MemAddr>::decodeToBits(getZ(),AllStubBase<ASType>::kASZSize);
+    str += "|"+MemoryTemplate<AllStub<ASType>, 3, kNBits_MemAddr>::decodeToBits(getPhi(),AllStubBase<ASType>::kASPhiSize);
     // Different get method for DISKPS?  Where is it?     
     //     str += "|"+decodeToBits(getAlpha(),AllStubBase<ASType>::kASAlphaSize);
-    str += "|"+decodeToBits(getBend(),AllStubBase<ASType>::kASBendSize);
+    str += "|"+MemoryTemplate<AllStub<ASType>, 3, kNBits_MemAddr>::decodeToBits(getBend(),AllStubBase<ASType>::kASBendSize);
     return str;
   }
+#endif
 
 private:
 
--- a/TrackletAlgorithm/AllStubMemory.h
+++ b/TrackletAlgorithm/AllStubMemory.h
@@ -3,6 +3,7 @@
 
 #include "Constants.h"
 #include "MemoryTemplate.h"
+#include "globalFunctions.h"
 
 // AllStubBase is where we define the bit widths, which depend on the class template parameter
 template<int ASType> class AllStubBase {};
@@ -188,12 +189,10 @@ public:
 
 #ifdef CMSSW_GIT_HASH
   std::string getBitStr() const {
-    std::string str = decodeToBits(getR(),AllStubBase<ASType>::kASRSize);
-    str += "|"+decodeToBits(getZ(),AllStubBase<ASType>::kASZSize);
-    str += "|"+decodeToBits(getPhi(),AllStubBase<ASType>::kASPhiSize);
-    // Different get method for DISKPS?  Where is it?     
-    //     str += "|"+decodeToBits(getAlpha(),AllStubBase<ASType>::kASAlphaSize);
-    str += "|"+decodeToBits(getBend(),AllStubBase<ASType>::kASBendSize);
+    std::string str = decodeToBits(getR());
+    str += "|"+decodeToBits(getZ());
+    str += "|"+decodeToBits(getPhi());
+    str += "|"+decodeToBits(getBend());
     return str;
   }
 #endif
--- a/TrackletAlgorithm/AllStubMemory.h
+++ b/TrackletAlgorithm/AllStubMemory.h
@@ -188,12 +188,12 @@ public:
 
 #ifdef CMSSW_GIT_HASH
   std::string getBitStr() const {
-    std::string str = MemoryTemplate<AllStub<ASType>, 3, kNBits_MemAddr>::decodeToBits(getR(),AllStubBase<ASType>::kASRSize);
-    str += "|"+MemoryTemplate<AllStub<ASType>, 3, kNBits_MemAddr>::decodeToBits(getZ(),AllStubBase<ASType>::kASZSize);
-    str += "|"+MemoryTemplate<AllStub<ASType>, 3, kNBits_MemAddr>::decodeToBits(getPhi(),AllStubBase<ASType>::kASPhiSize);
+    std::string str = decodeToBits(getR(),AllStubBase<ASType>::kASRSize);
+    str += "|"+decodeToBits(getZ(),AllStubBase<ASType>::kASZSize);
+    str += "|"+decodeToBits(getPhi(),AllStubBase<ASType>::kASPhiSize);
     // Different get method for DISKPS?  Where is it?     
     //     str += "|"+decodeToBits(getAlpha(),AllStubBase<ASType>::kASAlphaSize);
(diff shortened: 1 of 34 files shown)


TrackletAlgorithm/MemoryTemplate.h
```diff
@@ -130,24 +130,32 @@ class MemoryTemplate
...
   static constexpr int getWidth() {return DataType::getWidth();}
-  
+ 
+  std::string name_;
+  void setName(std::string name) { name_ = name;}
+  std::string const& getName() const { return name_;}
+
+  unsigned int iSector_;
+  void setSector(unsigned int iS) { iSector_ = iS;}
+  unsigned int getSector() const { return iSector_;}
```
aehart, line 157: @carriganm95 These lines and the similar ones in the other template classes cause the C/RTL cosimulation to fail for most of the processing modules. After playing around a bit, I think the problem is having data members (`name_` and `iSector_` in this case) that do not exist during synthesis, but do exist when the C++ test bench is run. Somehow this difference in the class members screws up the cosimulation.

I would suggest protecting these lines with `#ifdef CMSSW_GIT_HASH` instead of `#ifndef __SYNTHESIS__`. That should allow the cosimulation to run successfully again.

TrackletAlgorithm/AllStubMemory.h
```diff
@@ -186,6 +186,27 @@ class AllStub : public AllStubBase<ASType>
...
+  // TO DO: This belongs in some sort of helper class.
+  std::string decodeToBits(unsigned int field, unsigned int size) const {
+    unsigned int valtmp = field;
+    std::string str = "";
+    for(unsigned int i=0; i< size; i++) {
+      str = ((valtmp & 1) ? "1" : "0") + str;
+      valtmp >>= 1;
+    }
+    return str;
+  }
```
aehart, line 208: @carriganm95 Would it make sense to make this a static method of `MemoryTemplate`? That would avoid having this code duplicated multiple times.

TrackletAlgorithm/Constants.h
```diff
@@ -3,6 +3,11 @@
 
 #include "ap_int.h"
 
+#ifdef CMSSW_GIT_HAS
```
aehart, line 6: @carriganm95 Should this be `CMSSW_GIT_HASH`?

TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -10,7 +10,11 @@
 #include "AllStubMemory.h"
 #include "AllProjectionMemory.h"
 #include "FullMatchMemory.h"
-#include "hls_math.h"
```
aehart, line 13: @carriganm95 I just checked, and we don't actually need to include hls_math.h in any of the modules, so you can just remove this include statement everywhere you find it.

The only exception is this file, where you need to replace it with `#include <cassert>`.

TrackletAlgorithm/AllStubMemory.h
pwittich: does this need to be in a class at all? as far as I can tell it can just be a function.

aehart: No it doesn't need to be in a class, but I thought `MemoryTemplate` would be a good place for it, since it's only used within the memory classes.

TrackletAlgorithm/MemoryTemplate.h
carriganm95: Thanks. I noticed the fail yesterday but hadn’t figured out what went wrong. I will make this change.

TrackletAlgorithm/AllStubMemory.h
pwittich: I'm not commenting on what file it should sit in, just suggesting that it can be a simple function rather than a member function of some class.

aehart: At least in my mind, it seems intimately tied to the `MemoryTemplate` class, which is why I proposed making it a static method of that class.

TrackletAlgorithm/AllStubMemory.h
```diff
@@ -186,6 +186,27 @@ class AllStub : public AllStubBase<ASType>
     data_.range(kASBendMSB,kASBendLSB) = bend;
   }
 
+  std::string getBitStr() {
```
aehart, line 189: @carriganm95 Each of these `getBitStr` methods should be made `const`, i.e.,
`std::string getBitStr()` → `std::string getBitStr() const`

pwittich:
@carriganm95 we still now have two identical `decodeToBits` functions — can we just have it in one place? And if it’s not needed in the HLS can you also protect all those functions with #ifdef to make it clear that they are not needed in the HLS workflow.

TrackletAlgorithm/AllStubMemory.h
carriganm95: I did make some changes to put this function in MemoryTemplate. Brian also pointed out that because this function is used when we write memories to file it could go in FileReadUtility.h as a simple function

aehart:
> @carriganm95 we still now have two identical `decodeToBits` functions — can we just have it in one place? And if it’s not needed in the HLS can you also protect all those functions with #ifdef to make it clear that they are not needed in the HLS workflow.

For now, I would propose keeping just `MemoryTemplate::decodeToBits`. Later, I think we should introduce a common base class for all of the `MemoryTemplate*` classes, which can include this and other member functions that are currently duplicated.

I would also propose simplifying `decodeToBits` by removing the `size` argument. Since you're always passing in an `ap_int`/`ap_uint`, you should be able to infer the size with something like the following:
```c++
template<class T>
static std::string decodeToBits(const T field) {
    constexpr unsigned int size = T::width;
    …
```
And the compiler _should_ be able to deduce the template from the function argument, so you shouldn't have to specify the template when you call the function.

aehart:
> For now, I would propose keeping just `MemoryTemplate::decodeToBits`. Later, I think we should introduce a common base class for all of the `MemoryTemplate*` classes, which can include this and other member functions that are currently duplicated.

@carriganm95 Sorry, I just saw your reply in the inline comment thread. I would also be OK with putting `decodeToBits` in FileReadUtility.h as a standalone function.

TrackletAlgorithm/AllStubMemory.h
carriganm95: One thing I hadn't considered when I said that I would put this function (decodeToBits) in FileReadUtility.h is that the getBitStr() functions in AllStubMemory.h, etc. rely on this function and those files cannot directly include FileReadUtility.h. I could just use 

#include ../TestBenches/FileReadUtility.h

but I am thinking this is not good practice. Another suggestion from Brent was to just create a new file in TrackletAlgorithm called "GlobalFunctions.h" and include it there. Let me know if you have a preference on how this is handled and I can change it.

TrackletAlgorithm/MemoryTemplate.h
```diff
@@ -130,25 +130,35 @@ class MemoryTemplate
...
+
+  static std::string decodeToBits(unsigned int field, unsigned int size) {
+    unsigned int valtmp = field;
+    std::string str = "";
+    for(unsigned int i=0; i< size; i++) {
+      str = ((valtmp & 1) ? "1" : "0") + str;
+      valtmp >>= 1;
+    }
+    return str;
+  }
```
aehart, line 159: @carriganm95 Can you remove the versions of this function in the `MemoryTemplate`* classes?

TrackletAlgorithm/globalFunctions.h
```diff
@@ -0,0 +1,16 @@
+#ifndef TrackletAlgorithm_globalFunctions_h
+#define TrackletAlgorithm_globalFunctions_h
+
+#ifdef CMSSW_GIT_HASH
+
+ static std::string decodeToBits(unsigned int field, unsigned int size) {
```
aehart, line 6: @carriganm95 Is there any point in this being `static`, now that it is outside of a class?

Also, have you tried making this a template function, as I suggested?
```c++
template<class T>
static std::string decodeToBits(const T field) {
    constexpr unsigned int size = T::width;
    …
```
I think this would make the function more foolproof, since you don't need to give the size separately. And since the template parameter is the type of an argument, you shouldn't need to give it explicitly when calling the function, i.e., you can just call it as
```c++
decodeToBits(getR());
```
for example.

TrackletAlgorithm/MemoryTemplate.h
carriganm95: removed

TrackletAlgorithm/globalFunctions.h
carriganm95: I have implemented this templated function.

carriganm95:
I added the getName and getSector functions back into the MemoryTemplate files. I had removed them when there was an unknown error during compilation on our side. All of the changes previously suggested have been applied to those functions.

aehart (approved):
Looks good to me. Approved!

PR #168, opened 2021-06-21
InputRouter : fix to logic to assign ptr for LUT

sseifeln:
@aehart  : @meisonlikesicecream noticed that there were mismatches in the output of the IR for some events. I had a look a the code and : 

- the test bench was for some reason not up-to date with the version I had locally; this was also the case on the IR_emulatorSync branch (one commit was missing ..) - that I have fixed (wasn't the issue though) 
- the issue was here https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/InputRouter.h#L279-L291 - since I can't type I'd assigned the wrong phi correction to some layers (actually.. more like all layers except for the first PS layer) 

I've fixed the issue and made sure that the test bench is up-to date with my local copy. Seems to work now without mismatches... so.. a quick PR to fix this in the master branch as well!

### DIFF
--- a/TestBenches/InputRouter_test.cpp
+++ b/TestBenches/InputRouter_test.cpp
@@ -54,6 +54,10 @@ static const int kMaxNEvents = 100;  // max number of events to run
 // };
 
 
+// const int kIRNmemsLUT[] =
+// #include "../emData/LUTs/IR_Nmems.tab"
+// ;
+
 
 // map of detector regions read out [ per DTC ]
 using CablingMap = std::map<int, std::vector<uint8_t>> ;
@@ -67,6 +71,8 @@ using InputStubs = std::map<int, std::vector<std::string>> ;
 using Stubs = std::vector<std::bitset<kNBits_DTC>>; 
 // DTC layer map 
 typedef std::map<int, std::vector<int>> TkMap;
+
+const std::vector<uint8_t> kBnLbls = { 'A' , 'B' , 'C' , 'D' , 'E' , 'F' , 'G' , 'H'};
 
 
 using namespace std;
@@ -147,7 +153,7 @@ DtcMap getCablingMap( std::string pInputWiresMap )
       uint8_t cLayerId = (cLyrName.substr(0,1).find("D") != std::string::npos)  ? (10 + cLyr): cLyr; 
       auto cDtcIter = std::find(cDtcNames.begin(), cDtcNames.end(), cDtcName ) ;
       // detector region printout   
-      if( cDtcIter == cDtcNames.end() || cDtcNames.empty())
+      if( cDtcIter == cDtcNames.end() || cDtcNames.size() == 0)
       {
         // push back dtc name
         cDtcNames.push_back( cDtcName );
@@ -250,14 +256,22 @@ DtcMemWrd getNMemories( int pLinkId
   DtcMemWrd cDtcMemWrd;
   std::string cDtcName = getDTCName( pLinkId, pInputFile_Wires ); 
   std::vector<uint8_t> cLyrs = getLyrs( pLinkId , pInputFile_Wires );
-  //std::cout << "Found " << +cLyrs.size() << " layer(s) readout out by [DTC]" << cDtcName << "\n";
+  std::cout << "Found " << +cLyrs.size() << " layer(s) readout out by [DTC]" << cDtcName << "\n";
   std::vector<int> cNmems(cLyrs.size(),0);
   ap_uint<kBINMAPwidth> cLinkNPhiBns=0;
   int cOffst=0; 
+
+  ofstream cOstream_LUT;
+  if( pLinkId == 0 )
+    cOstream_LUT.open ("emData/IR_Nmems.tab",ios::out);
+  else
+    cOstream_LUT.open ("emData/IR_Nmems.tab",ios::app);
+  
   for( auto cLyr : cLyrs )
   {
     //std::cout << "Lyr:" << +cLyr << "\n";
-    int cNmemories=0;
+    cOstream_LUT << +pLinkId << "\t" << +cLyr << "\t"; 
+    std::vector<int> cPhiBns(0);
     for( size_t cPhiBn=0; cPhiBn<kMaxPhiBnsPrLyr; cPhiBn++)
     {
       // re-construct file name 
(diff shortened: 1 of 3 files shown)


sseifeln:
Sorry .. stop .. changed the link number and I still have a problem. I will keep checking ..

TestBenches/InputRouter_test.cpp
```diff
@@ -147,7 +109,7 @@ DtcMap getCablingMap( std::string pInputWiresMap )
       uint8_t cLayerId = (cLyrName.substr(0,1).find("D") != std::string::npos)  ? (10 + cLyr): cLyr; 
       auto cDtcIter = std::find(cDtcNames.begin(), cDtcNames.end(), cDtcName ) ;
       // detector region printout   
-      if( cDtcIter == cDtcNames.end() || cDtcNames.empty())
```
meisonlikesicecream, line 150: The CI seems to complain about using size instead of empty

TestBenches/InputRouter_test.cpp
```diff
@@ -612,5 +671,11 @@ int main(int argc, char * argv[])
...
 
+  // remove temp LUT 
+  if( remove( "emData/LUTs/IR_Nmems.tab" ) == 0 )
+  {
+    std::cout << "Temperorary LUT created by IR test bench deleted.\n";
+  }
+
   return cTotalErrCnt;
-}
+}
```
meisonlikesicecream, line 681: Add a newline! :-)

project/script_IR.tcl
```diff
@@ -27,10 +27,10 @@ create_clock -period 240MHz -name slow_clock
 create_clock -period 360MHz -name fast_clock
 
 set nProc [exec nproc]
-csim_design -mflags "-j$nProc" -argv "--link,8 --dtcSplit,0 --tkNonant,4"
+csim_design -compiler gcc -mflags "-j$nProc" -argv "--link,8 --dtcSplit,0 --tkNonant,4 --lastBx,100"
```
meisonlikesicecream, line 30: Apparently we're not supposed to use the compiler flag anymore due to Vitis compatibility (quote Andrew)

sseifeln: thanks - didn't catch that I'd changed it back. will fix.

TestBenches/InputRouter_test.cpp
sseifeln: I'll check the CI failure and fix..

meisonlikesicecream (approved):

meisonlikesicecream:
@aehart would you mind having a quick look at this as well? I don't dare to press merge :-)

project/script_IR.tcl
```diff
@@ -27,10 +27,10 @@ create_clock -period 240MHz -name slow_clock
 create_clock -period 360MHz -name fast_clock
 
 set nProc [exec nproc]
-csim_design -mflags "-j$nProc" -argv "--link,8 --dtcSplit,0 --tkNonant,4"
+csim_design -mflags "-j$nProc" -argv "--link,8 --dtcSplit,0 --tkNonant,4 --lastBx,100"
```
aehart, line 30: @sseifeln Does the addition of this argument do anything, since the test bench is processing 100 events by default?

TestBenches/InputRouter_test.cpp
```diff
@@ -429,19 +472,21 @@ int main(int argc, char * argv[])
         cNonant  = std::stol( cArg.substr( cArg.find(",")+1 , cArg.length()-1) , nullptr, 10 );
         std::cout << "Tk nonant passed from cmd line is " << cDTCsplit << "\n";
       }
+        if( cArg.find("lastBx") != std::string::npos ) 
+      {
+        cLastBx  = std::stol( cArg.substr( cArg.find(",")+1 , cArg.length()-1) , nullptr, 10 );
+        std::cout << "Last Bx to process assed from cmd line is " << cLastBx << "\n";
```
aehart, line 478: @sseifeln I'm guessing "assed" is a typo. Can you fix it before we merge :)

TestBenches/InputRouter_test.cpp
```diff
@@ -398,14 +417,38 @@ void prepareInputStreams( ifstream * pInputStreams
...
+  do
+  {
+    auto cDtcName = cMapIter->first;
+    auto cLyrs = cMapIter->second; 
+    auto cNmems = getNMemories( cLinkId, pDTCsplit, pNonant, pInputFile_Wires);
+    auto cBnWrd = cNmems.second; 
+    cLinkId++;  
+    cMapIter++;
+  }while( cMapIter != cDtcMap.end() );
+}
```
aehart, line 440: @sseifeln The only thing this function seems to do is create IR_Nmems.tab via the call to `getNMemories`. And then I don't see IR_Nmems.tab being used before it is removed at the end. I also don't see how `cDtcName`, `cLyrs`, `cNmems`, and `cBnWrd` are used here. Can this be cleaned up?

TestBenches/InputRouter_test.cpp
```diff
@@ -612,5 +671,11 @@ int main(int argc, char * argv[])
   }
   cLinkDataStream.close();
 
+  // remove temp LUT 
+  if( remove( "emData/LUTs/IR_Nmems.tab" ) == 0 )
+  {
+    std::cout << "Temperorary LUT created by IR test bench deleted.\n";
+  }
+
```
aehart, line 679: @sseifeln I guess you should remove "emData/IR_Nmems.tab", although again, I'm not sure what the purpose of this file is since the test bench both creates and removes it.

TestBenches/InputRouter_test.cpp
sseifeln: 🤦 yes.. typo! Removed the option of modifying this from the cmd line (was using it during testing the logic but it shouldn't be needed now)

TestBenches/InputRouter_test.cpp
sseifeln: Sorry.. legacy! Now removed.

TestBenches/InputRouter_test.cpp
sseifeln: Also removed (again). Sorry!

aehart:
OK, looks good to me. Merging now.

PR #166, opened 2021-06-14
Future CMSSW changes for compilation with CMSSW_11_3

carriganm95:
This pull request has the changes needed by the future CMSSW emulation to compile and run the match processor. The changes made are

1. getName() and getSector() functions are added to MemoryTemplate.h, MemoryTemplateBinned.h, and MemoryTemplateBinnedCM.h
2. The functions getBitStr() and decodeToBits() have been added to AllStubsMemory.h, FullMatchMemory.h, TrackletProjectionsMemory.h, and VMStubsMEMemoryCM.h in order to allow the emulation to write out the memories to file.
3. Ifndef statements have been added around #include hls_math.h in order to exclude them when in CMSSW
4. Ifndef statements are added to Constants.h to prevent redefining of N_LAYERS and N_DISKS
5. Changed ints to unsigned int in MatchProcessor.h, MemoryTemplate.h, MemoryTemplateBinned.h, and MemoryTemplateBinnedCM.h

PR #170, opened 2021-06-23
Changes to FileReadUtility to Work with Future CMSSW Emulation

carriganm95:
In order to run the writeMemFromFile function in the CMSSW emulation it was necessary to make some changes to the function. Because the event loop in CMSSW opens and closes the files each event the writeMemFromFile function needs to start at the beginning of the txt files and look for the correct memory. There are also some functions that needed to be inline to avoid multiple definitions at compile time.

### DIFF
--- a/TrackletAlgorithm/AllStubMemory.h
+++ b/TrackletAlgorithm/AllStubMemory.h
@@ -186,15 +186,17 @@ public:
     data_.range(kASBendMSB,kASBendLSB) = bend;
   }
 
+#ifdef CMSSW_GIT_HASH
   std::string getBitStr() const {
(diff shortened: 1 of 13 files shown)


tomalin:
This PR claims you've changed 13 files. Is this really the case? Perhaps you just need to rebase to the latest master?

Back up your code, in case git messes up
git pull
git rebase origin/master
Check that your local directory contains a merge of your changes and other peoples.
git diff --stat --cached origin/PR_6_23 (check what you will push)
git push --force-with-lease origin PR_6_23

pwittich:
@carriganm95 it looks like a lot of these changes are the same as in PR #166; is that right? I think that might be part of the confusion?

PR #172, opened 2021-06-28
IR-VMR chain VHDL testbench

meisonlikesicecream:
Added a VHDL testbench for the IR-VMR chain along with other files and modifications to make it run. 

Some updates:
- CompareMemPrintsFW.py can now handle binned memories and I have added predefined comparisons for the IR-VMR chain, and for a single VMR (used when I was debugging, might delete this? Or add the VMR VHDL testbench files?)
- Added a FileReaderFIFO.vhd that reads the .txt data files and sends them to the IR
- Added a FileWriterFromRAMBinned.vhd that can write the dataOut/*.txt output files from binned memories

Some temporary bodges I had to make:
1. FileReader.vhd assumes all unbinned memory addresses starts with "0x" in the testvectors, but this is not the case for the IL memories. This should probably be properly fixed in the C++ emulation.
2. The empty_neg signal in FileReaderFIFO.vhd is set to '1' all the time, i.e. saying that it contains data even though it might not be true. This is because the IR stopped processing stubs if empty_neg was '0' for some weird reason. If we have read all the data in the testvectors for a specific event, it just sends out zeros as the data.
3. The AllStub memories don't have nentries ports (because the module that reads those memories doesn't use them), which causes FileWriterFromRAM.vhd to not write out any stubs in the dataOut/*.txt files. To be able to make somewhat of a comparison of the output, I let FileWriterFromRAM write out the stubs in all 108 data addresses although this causes some stubs from previous events to be written out for multiple events (if the current event contains less stubs thus doesn't overwrite them). Therefore, I commented out the AllStub memory comparisons in CompareMemPrintsFW.py. There is another memory that also will have this issue.
4. The IR originally looped over +200 stubs instead of the normal 108 like all the other modules do. So to make the IR-VMR testbench run properly I had to decrease the number of loop iterations to 108, but this causes some missing stubs in the output for both the IR and the VMR as the testvectors for the DTC Links contain more than 108 stubs. To avoid the IR HLS simulations failing I changed the number of events tested to 2... what do we do about this?

When running the IR-VMR chain, most events are good, although some stubs are missing due to bodge number 4 (might be some other hidden issues as well as I haven't checked every single bad event). I made this PR a draft as I'm not sure what to do with some of the bodges, especially 3 and 4. If anyone has any thoughts about them, that would be nice :-)


TrackletAlgorithm/InputRouter.h
```diff
@@ -238,7 +237,7 @@ void InputRouter( const BXType bx
 	ClearCounters<nOMems>(nMems,  hNStubs);
 
 	LOOP_ProcessIR:
-	for (int cStubCounter = 0; cStubCounter < kMaxStubsFromLink; cStubCounter++) 
+	for (int cStubCounter = 0; cStubCounter < kMaxProc; cStubCounter++)
```
tomalin, line 240: How does  the code fail without this change? Naively the VHDL produced by the HLS compiler shouldn't depend on the assumed length of the FIFO (or equivalently on the number of loops per TM period).

tomalin:
To answer you question 4: Since we'll run entire FW chain this summer at 240MHz, all txt files should be remade with the constant maxStep_ of the IR in the C++ file Settings.h set accordingly.

TrackletAlgorithm/InputRouter.h
meisonlikesicecream: It fails because the IR sends out the done signal every 256 (kMaxStubsFromLink) clocks even if the input bx changes every 108 clocks.

tomalin:
Your bodge 1 is fine for a temporary fix. But we should ultimately fix this in the C++ emulation. i.e. Before the line https://github.com/cms-L1TK/cmssw/blob/L1TK-PR-june2021/L1Trigger/TrackFindingTracklet/src/InputLinkMemory.cc#L26 , there is a missing line 'out_ << "0x";' (as is clear from comparison with TrackletProjectionsMemory.cc).  @aehart  should consider porting this to the fw_sync C++ branch.

tomalin:
To avoid bodge 3, either:
1) Modify python scripts, so whenever a memory is at the end of the chain, it is always given an nentries output port, even if populate_numEntries_out() https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L225 says it doesn't need one.
2) Thomas Schuh suggestion: debate with HLS group in we can redefine all chains so as to exclude the final memory.

tomalin:
https://github.com/cms-L1TK/firmware-hls/blob/irvmr_chain_tb/IntegrationTests/IRVMR/script/makeProject.tcl#L16 says that both IR in this chain share the same IP. If it is known which IR in a nonant can share the same IP, then then each shared IP should be given a unique IP name in https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L248 . This avoids the need to compile two or more copies of it in makeProject.tcl.

IntegrationTests/IRVMR/tb/tb_tf_top.vhd
```diff
@@ -0,0 +1,373 @@
...
+  -- Specify version of chain to run from TB:
+  --    0 = SectorProcessor.vhd from python script.
+  --    1 = SectorProcessorFull.vhd from python script (gives intermediate MemPrints).
+  --    N.B. Change this also in makeProject.tcl !
+  constant INST_TOP_TF   : integer := 1; 
+  --=========================================================================
+
+  constant CLK_PERIOD        : time    := 4 ns;       --! 250 MHz
+  constant DEBUG             : boolean := false;      --! Debug off/on
+  constant DL_DELAY       : integer := 0;          --! Number of BX delays (can be written early 8 pages)
```
tomalin, line 58: The comment for DL_DELAY doesn't make sense. The DL is a FIFO so doesn't have pages. (Please also fix https://github.com/cms-L1TK/firmware-hls/blob/master/IntegrationTests/PRMEMC/tb/tb_tf_top.vhd#L58 , which also doesn't make sense, since TPROJ is only 2 pages). If there are 8 pages, then read & write to a given page are widely seperated in time, so getting things synchronized can be more relaxed.

IntegrationTests/IRVMR/tb/tb_tf_top.vhd
```diff
@@ -0,0 +1,373 @@
...
+    if START_FIRST_LINK = '1' then
+      if rising_edge(CLK) then
+
+        if (CLK_COUNT < MAX_ENTRIES) then
+          CLK_COUNT := CLK_COUNT + 1;
+        else
+          CLK_COUNT := 1;
+          EVENT_COUNT := EVENT_COUNT + 1;
+
+          IR_START <= '1';
```
tomalin, line 176: This logic appears to set IR_START to 1 in the same clk cycle that it releases the RESET. If possible, release RESET a few clks (or more) earlier than IR_START, so as to be sure the reset has finished.

IntegrationTests/IRVMR/tb/tb_tf_top.vhd
```diff
@@ -0,0 +1,373 @@
...
+    writeAS_36 : entity work.FileWriterFromRAM 
+    generic map (
+      FILE_NAME  => FILE_OUT_AS_36&memory_enum_to_string(var)&outputFileNameEnding,
+      RAM_WIDTH  => 36,
+      NUM_PAGES  => 8
+    )
+    port map (
+      CLK => CLK,
+      DONE => VMR_DONE,
+      NENT_ARR => AS_36_mem_AAV_dout_nent(var), -- always 0, will that cause issues in FileWriterFromRAM?
```
tomalin, line 312: Regarding question in comment, yes in nent=0, FileWriterFromRAM will not attempt to read any data from the RAM.

tomalin:
Regarding bodge 2: I suspect this is necessary as https://github.com/cms-L1TK/firmware-hls/blob/irvmr_chain_tb/TrackletAlgorithm/InputRouter.h#L245 always reads 108 entries per event from its input data. So FileReaderFifo.vhd must provide them. If InputRouter.h had used hls::stream for its stream input, (which it can't because hls::stream is unavailable in CMSSW), then then InputRouter.h could have called stream::empty() to check if no more data was available, and then the VHDL produced by the compiler from it would have checked the empty_neg signal. If we want to achieve that without using hls::stream, we could try adding an ap_uint<1> empty argument to InputRouter.h, but I'm not sure it's worth it.

IntegrationTests/common/script/CompareMemPrintsFW.py
```diff
@@ -50,6 +50,10 @@ class ReferenceType(Enum):
     CM     = 'CandidateMatches'
     FM     = 'FullMatches'
     VMPROJ = 'VMProjections'
+    VMSME  = 'VMStubs'
+    VMSTE  = 'VMStubs'
+    AS     = 'AllStubs'
+    IL     = 'InputStubs'
```
tomalin, line 56: If easy, please add lines for all the memories in the whole L1Trk chain here, so script will work for TE-TC etc.

meisonlikesicecream:
> https://github.com/cms-L1TK/firmware-hls/blob/irvmr_chain_tb/IntegrationTests/IRVMR/script/makeProject.tcl#L16 says that both IR in this chain share the same IP. If it is known which IR in a nonant can share the same IP, then then each shared IP should be given a unique IP name in https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L248 . This avoids the need to compile two or more copies of it in makeProject.tcl.

I don't know which IRs share the same IP core, or if there is an easy way to check this. It just happened to be true in this specific case. It depends on the number of inputs/outputs, and also which layer it is because of the LUTs guess?

IntegrationTests/IRVMR/tb/tb_tf_top.vhd
meisonlikesicecream: Ah, that's an old comment, I'll remove it.

IntegrationTests/IRVMR/tb/tb_tf_top.vhd
meisonlikesicecream: The RESET is set to zero one clock cycle before IR_START is set to one (looking at the waveforms confirms this). Is one clock not enough?

IntegrationTests/IRVMR/tb/tb_tf_top.vhd
meisonlikesicecream: OK, I just removed the parentheses now.

IntegrationTests/common/script/CompareMemPrintsFW.py
meisonlikesicecream: OK!

IntegrationTests/common/script/CompareMemPrintsFW.py
```diff
@@ -233,7 +341,7 @@ def comparePredefined(args):
     parser.add_argument("-c","--comparison_filename",default="output.txt",help="The filename of the testbench output file (default = %(default)s)")
     parser.add_argument("-f","--fail_on_error",default=False,action="store_true",help="Raise an exception on the first error as opposed to simply printing a message (default = %(default)s)")
     parser.add_argument("-l","--file_location",default="./",help="Location of the input files (default = %(default)s)")
-    parser.add_argument("-p","--predefined",default=False,action="store_true",help="Run predefined comparisons (default = %(default)s)")
+    parser.add_argument("-p","--predefined",default="",help="Run predefined comparisons. Options: PRMEMC, IRVMR, VMR (default = %(default)s)")
```
tomalin, line 349: Could script get default value of option "-p" from:
import os
l1TrkChain = os.getcwd().split('/')[-2]

IntegrationTests/common/script/CompareMemPrintsFW.py
```diff
@@ -182,31 +221,105 @@ def compare(comparison_filename="", fail_on_error=False, file_location='./', pre
...
+                           reference_filenames=["../../../emData/MemPrints/Matches/FullMatches_FM_L5L6_L3PHIC_04.dat"], save=args.save, verbose=args.verbose)
+
+    elif (args.predefined == "IRVMR"):
+        # IL
+        for i in ['L', 'D']:
+            for j in ['A', 'B', 'C', 'D']:
+                if os.path.exists(('./dataOut/IL_%c2PHI%c_PS10G_3_A.txt' % (i,j))):
+                    ret_sum += compare(comparison_filename=("./dataOut/IL_%c2PHI%c_PS10G_3_A.txt" % (i,j)), fail_on_error=False, file_location=args.file_location, predefined=args.predefined,
+                                       reference_filenames=[("../../../emData/MemPrints/InputStubs/InputStubs_IL_%c2PHI%c_PS10G_3_A_04.dat" % (i,j))], save=args.save, verbose=args.verbose)
+        # AllStubs - temporarily removed due to issues with not having nentries port
```
tomalin, line 259: We need a way to recover the comparison for the AllStub memory.

IntegrationTests/common/script/CompareMemPrintsFW.py
```diff
@@ -182,31 +221,105 @@ def compare(comparison_filename="", fail_on_error=False, file_location='./', pre
...
+            if os.path.exists(('./dataOut/CM_L3PHIC%i.txt' % (i))):
+                ret_sum += compare(comparison_filename=("./dataOut/CM_L3PHIC%i.txt" % (i)), fail_on_error=False, file_location=args.file_location, predefined=args.predefined,
+                                   reference_filenames=[("../../../emData/MemPrints/Matches/CandidateMatches_CM_L3PHIC%i_04.dat" % (i))], save=args.save, verbose=args.verbose)
+        
+        ret_sum += compare(comparison_filename="./dataOut/FM_L1L2_L3PHIC.txt", fail_on_error=False, file_location=args.file_location, predefined=args.predefined,
+                           reference_filenames=["../../../emData/MemPrints/Matches/FullMatches_FM_L1L2_L3PHIC_04.dat"], save=args.save, verbose=args.verbose)
+        ret_sum += compare(comparison_filename="./dataOut/FM_L5L6_L3PHIC.txt", fail_on_error=False, file_location=args.file_location, predefined=args.predefined,
+                           reference_filenames=["../../../emData/MemPrints/Matches/FullMatches_FM_L5L6_L3PHIC_04.dat"], save=args.save, verbose=args.verbose)
+
+    elif (args.predefined == "IRVMR"):
```
tomalin, line 252: I assume all this hand-written code will need rewriting whenever we include different modules in the IRVMR chain or PRMEMC chain etc. Not required now, but we should think how to automate this.

meisonlikesicecream: I automated this.

IntegrationTests/common/script/CompareMemPrintsFW.py
meisonlikesicecream: Not if you want to be able to run it without any predefined option? But this is irrelevant now when I have automated the predefined comparison.

IntegrationTests/common/script/CompareMemPrintsFW.py
meisonlikesicecream: The allstub comparison is back now, although this increased the number of errors from 74 to 614.

IntegrationTests/common/script/CompareMemPrintsFW.py
tomalin: clever

meisonlikesicecream:
> To answer you question 4: Since we'll run entire FW chain this summer at 240MHz, all txt files should be remade with the constant maxStep_ of the IR in the C++ file Settings.h set accordingly.

I tried creating my own testvectors with at most 108 for the DL links, and ignoring the allstub memories, there were no errors in the IR-VMR chain. So the remaining errors in this chain are only due to issue 3 and 4.

tomalin:
> To avoid bodge 3, either:
> 
> 1. Modify python scripts, so whenever a memory is at the end of the chain, it is always given an nentries output port, even if populate_numEntries_out() https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L225 says it doesn't need one.
> 2. Thomas Schuh suggestion: debate with HLS group in we can redefine all chains so as to exclude the final memory.

Now fixed with solution 1

tomalin (approved):
Looks really great. Approved.

PR #173, opened 2021-06-29
Decreased number of stubs in binned memory

meisonlikesicecream:
Temporary decrease the number of stubs per bin in the binned memories by -1 to match the testvectors/C++ emulation. The VMR now passes the simulations for all phi regions and layers tested.


aehart (approved):
Looks good to me.

PR #174, opened 2021-07-05
TE-TC chain

tomalin:
TE-TC chain, including VHDL test bench. The top-level VHDL was produced by the python scripts.  The CompareMemPrintsFW.py scripts says that 98% of events agree with C++ emulation.

This include a few bug fixes to common/hdl/tf_lut.vhd, including changing the read latency to 2, to correspond to that specified in the HLS https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/TrackletEngineTop.cc#L11 . (Though I doubt that an output reg is required simply to read ROM data).

A new function getDirEMDATA has been added to tf_pkg.vhd. It returns the relative directory path to emData/  . This function solves the issue then when tf_lut.vhd reads the .tab file, it must use a different relative dir path during the SIM vs SYNTH steps.

P.S. TrackletEngine_test.cpp edited to set default TE to one used inside this chain, to faciltate debugging.

### DIFF
--- a/IntegrationTests/TETC/hdl/SectorProcessor.vhd
+++ b/IntegrationTests/TETC/hdl/SectorProcessor.vhd
@@ -1,0 +1,1528 @@
+--! Standard libraries
+library IEEE;
+use IEEE.STD_LOGIC_1164.ALL;
+--! User packages
+use work.tf_pkg.all;
+use work.memUtil_pkg.all;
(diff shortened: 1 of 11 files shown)


aehart (approved):
Looks good, @tomalin. Approved!

PR #178, opened 2021-07-06
Top level generation script for PR, MC, TC, TB

### LINKED ISSUE #177: Scripts to generate top functions
There is currently a Python script that can generate the supported top functions of the TC from the wiring configuration:
https://github.com/cms-L1TK/firmware-hls/blob/master/emData/generate_TC.py
This script also generates all the bit masks and other constants that are unique to each instance of the TC.

All of the modules should have similar scripts written for them that generate top functions from a given wiring configuration, at least for the instances that will appear in the summer chain (L1L2 seeds with matches only in the barrel).

bryates:
Resolves #177 for MC, other modules can push to this PR.


emData/generate_MC.py
```diff
@@ -0,0 +1,170 @@
...
+        "}\n"
+    )
+
+    # Print out prototype for top function for this MC.
+    topHeaderFile.write(
+        "\n"
+        "constexpr int maxMatchCopies(" + str(len(TF_index)) + ");\n"
+        "constexpr int maxFullMatchCopies(" + str(len(TF_index)) + ");\n"
+        "\n"
+        "void MatchCalculator_" + seed + iMC + "(\n"
```
aehart, line 106: @bryates This is nitpicky, but could you use the same suffix style as is used in the wiring, e.g. something like `MatchCalculator_L3PHIC` instead of `MatchCalculator_L3C`?

aehart:
@bryates Thanks for responding so quickly.

Could you please also replace the files in the repo with those generated by generate_MC.py, run on the current wiring (i.e., that in emData/LUTs/wires.dat after running emData/download.sh)? I think all the modules should do this in order to verify that the scripts produce correct output.

Also, are you planning to expand the script to generate other MC top functions? For the reduced config, we'll need MCs for L3, L4, L5, and L6. And the current reduced config uses MC_L*PHIB, so we'll need support for that phi sector at least, though generating top functions for all phi sectors would be ideal.

TrackletAlgorithm/MatchCalculator.h
```diff
@@ -227,6 +227,22 @@ void readTable_Cuts(ap_uint<width> table[depth]){
...
+    return BARRELPS;
+  }
+
+  template<> constexpr regionType APRegion<TF::L3>() {
+    return BARRELPS;
+  }
+
+  template<> constexpr regionType FMRegion<TF::L3>() {
+    return BARREL;
+  }
```
bryates, line 245: I added these lines to make the templating easier, but I'm not sure if this is the right file to put them in. Should we put things like this in Constants.h, or should the generation scripts be able to calculate the template parameters?

emData/generate_MC.py
bryates: Good catch!

bryates:
> @bryates Thanks for responding so quickly.
> 
> Could you please also replace the files in the repo with those generated by generate_MC.py, run on the current wiring (i.e., that in emData/LUTs/wires.dat after running emData/download.sh)? I think all the modules should do this in order to verify that the scripts produce correct output.
> 
> Also, are you planning to expand the script to generate other MC top functions? For the reduced config, we'll need MCs for L3, L4, L5, and L6. And the current reduced config uses MC_L*PHIB, so we'll need support for that phi sector at least, though generating top functions for all phi sectors would be ideal.

I've replaced the default files with the generated ones. Should we modify the script to directly place them in the correct directory, instead of in `emData`?
I'll start playing with the other layers and phi sector B.

aehart:
> I've replaced the default files with the generated ones. Should we modify the script to directly place them in the correct directory, instead of in `emData`?

Yeah, I think it makes sense to do that. Good suggestion.

TrackletAlgorithm/MatchCalculator.h
aehart: As we move to automatically generating the top functions, I think the simplest thing is to have the scripts calculate these parameters directly. So, e.g., instead of the scripts generating a call to `ASRegion<TF::L3>()`, they would just write `BARRELPS` where appropriate. I'll need to update generate_TC.py to be consistent with this approach, but I think it's the way we should go.

bryates: Done.

bryates:
> > @bryates Thanks for responding so quickly.
> > Could you please also replace the files in the repo with those generated by generate_MC.py, run on the current wiring (i.e., that in emData/LUTs/wires.dat after running emData/download.sh)? I think all the modules should do this in order to verify that the scripts produce correct output.
> > Also, are you planning to expand the script to generate other MC top functions? For the reduced config, we'll need MCs for L3, L4, L5, and L6. And the current reduced config uses MC_L*PHIB, so we'll need support for that phi sector at least, though generating top functions for all phi sectors would be ideal.
> 
> I've replaced the default files with the generated ones. Should we modify the script to directly place them in the correct directory, instead of in `emData`?
> I'll start playing with the other layers and phi sector B.

I've tested the updates with `L3PHIB` and `L5PHIB`, and both work correctly.

pwittich:
This code is not safe for python 3. The fixes are trivial.

1. add `from __future__ import absolute_import` to each file
2. add `from __future__ import print_function` and add parens around each print call, and resolve any remaining complaints.

```
(py2) ➜  emData git:(generate_top) pylint --py3k generate_*.py
No config file found, using default configuration
************* Module generate_MC
W:  7, 0: import missing `from __future__ import absolute_import` (no-absolute-import)
E: 27, 4: print statement used (print-statement)
************* Module generate_TC
W:  8, 0: import missing `from __future__ import absolute_import` (no-absolute-import)
E: 64, 4: print statement used (print-statement)

------------------------------------------------------------------
Your code has been rated at 9.45/10 (previous run: 9.60/10, -0.15)
```

emData/generate_MC.py
```diff
@@ -0,0 +1,186 @@
...
+# dictionaries with the MC names as keys.
+wiresFile = open(wiresFileName)
+CMMems = {}
+FMMems = {}
+for line in wiresFile:
+    # Only barrel-only seeds are supported right now.
+    # Works for MC L3 phi sector C
+    if not any(["MC_" + l + "PHIB" in line for l in ["L3", "L4", "L5", "L6"]]) and \
+       not any(["MC_" + l + "PHIC" in line for l in ["L3"]]): #MC_L3PHIC was used for testing
+        continue
```
aehart, line 43: @bryates Is there any reason we can't generate top functions for all MCs, or at least all barrel MCs? We don't need to test them all right now, but it would be nice if they were all defined.

emData/generate_MC.py
```diff
@@ -0,0 +1,186 @@
...
+    line = line.rstrip()
+    tcName = re.sub(r".*MC_(......).*", r"MC_\1", line)
```
aehart, line 45: @bryates Please change `tcName` → `mcName`.

emData/generate_MC.py
```diff
@@ -0,0 +1,186 @@
...
+            FMMems[tcName] = []
+        FMMems[tcName].append(FM)
+wiresFile.close()
+
+# Open and print out preambles for the parameters and top files.
+dirname = os.path.dirname(os.path.realpath('__file__'))
+parametersFile = open(os.path.join(dirname, "../TrackletAlgorithm/MatchCalculator_parameters.h"), "w")
+parametersFile.write(
+    "#ifndef TrackAlgorithm_MatchCalculator_parameters_h\n"
+    "#define TrackAlgorithm_MatchCalculator_parameters_h\n"
```
aehart, line 63: @bryates Please change `TrackAlgorithm` → `TrackletAlgorithm`

emData/generate_MC.py
```diff
@@ -0,0 +1,186 @@
...
+    "// FMMask. The bits of this mask, from least significant to most\n"
+    "// significant, represent the memories in the order they are passed to\n"
+    "// MatchCalculator; e.g., the LSB corresponds to\n"
+    "// TF::L1L2. If a bit is set, the corresponding memory is\n"
+    "// valid, if it is not, the corresponding memory is not valid.\n"
+)
+topHeaderFile = open(os.path.join(dirname, "../TrackletAlgorithm/MatchCalculatorTop.h"), "w")
+topHeaderFile.write(
+    "#ifndef TrackAlgorithm_MatchCalculatorTop_h\n"
+    "#define TrackAlgorithm_MatchCalculatorTop_h\n"
```
aehart, line 78: @bryates Please change `TrackAlgorithm` → `TrackletAlgorithm`

emData/generate_MC.py
```diff
@@ -0,0 +1,186 @@
...
+    "\n"
+    "////////////////////////////////////////////////////////////////////////////////\n"
+    "// Top functions for various MatchCalculators (MC). For each iteration of\n"
+    "// the main processing loop, a MC retrieves an array of Candidate Matches,\n"
+    "// and computes Full Matches from each.\n"
+    "////////////////////////////////////////////////////////////////////////////////\n"
+)
+
+# Calculate parameters and print out parameters and top function for each MC.
+for tcName in sorted(CMMems.keys()):
```
aehart, line 97: @bryates Please change `tcName` → `mcName`.

emData/generate_PR.py
```diff
@@ -0,0 +1,103 @@
+# Automatically generates ProjectionRouterTop.h, and ProjectionRouterTop.cc
```
aehart, line 1: @djcranshaw Please add the following imports to ensure Python3 compatibility:
```python
from __future__ import absolute_import, print_function
```
They have to be included before anything else.

emData/generate_PR.py
```diff
@@ -0,0 +1,103 @@
...
+import sys, os
+
+modules_to_build = ["PR_L1PHIB","PR_L2PHIB","PR_L3PHIB","PR_L4PHIB","PR_L5PHIB","PR_L6PHIB",
+                    "PR_L1PHIC","PR_L2PHIC","PR_L3PHIC","PR_L4PHIC","PR_L5PHIC","PR_L6PHIC"]
```
aehart, line 5: @djcranshaw Is there any reason we can't expand this list to include all PRs, or at least all barrel PRs? We don't need to test them all right now, but it would be nice if they were all defined.

emData/generate_PR.py
```diff
@@ -0,0 +1,103 @@
...
+
+def getNTProjAndNVMProj(module, wiresFileName):
+    wiresFile = open(wiresFileName)
+    wiresInfo = wiresFile.read()
```
aehart, line 9: @djcranshaw Close `wiresFile` after reading it.

emData/generate_PR.py
```diff
@@ -0,0 +1,103 @@
...
+    for i in range(TProjCount):
+        strSource += "#pragma HLS resource variable=projin[" + str(i) + "]->get_mem() latency=2\n"
+    strSource += "  constexpr unsigned int nInMem = " + str(TProjCount) + ";\n"
+    strSource += "  constexpr unsigned int nOutMem = " + str(VMProjCount) + ";\n"
+    strSource += "  constexpr int layer = " + str(layer) + ";\n"
+    strSource += "  constexpr int disk = " + str(disk) + ";\n"
+    strSource += module + ": ProjectionRouter<" + TProjRegion + ", " + VMProjRegion + ", nInMem, nOutMem, layer, disk>\n"
+    strSource += "    (bx, projin, bx_o, allprojout, vmprojout);\n"
+    strSource += "}\n\n"
+    return strSource
```
aehart, line 75: @djcranshaw I guess this function will return nonsense if the module is not in the given wiring? I think either it should return an empty string when `TProjCount` and `VMProjCount` are zero, or you should generate `modules_to_test` from the wiring itself.

emData/generate_MC.py
bryates: Done. I'll push the change soon.

emData/generate_MC.py
bryates: Done. I'll push the change soon.

emData/generate_MC.py
bryates: Done. I'll push the change soon.

emData/generate_MC.py
bryates: Done. I just have disks disabled now.
https://github.com/cms-L1TK/firmware-hls/blob/a4edeb22db02d914e2987074766de0f12aedbd19/emData/generate_MC.py#L40

emData/generate_MC.py
bryates: Done.
https://github.com/cms-L1TK/firmware-hls/blob/a4edeb22db02d914e2987074766de0f12aedbd19/emData/generate_MC.py#L59-L60

emData/generate_MC.py
```diff
@@ -0,0 +1,185 @@
...
+)
+
+# Calculate parameters and print out parameters and top function for each MC.
+for mcName in sorted(CMMems.keys()):
+    seed = re.sub(r"MC_(..)....", r"\1", mcName)
+    iMC = re.sub(r"MC_.....(.)", r"\1", mcName)
+
+    # numbers of memories
+    nCMMem = len(CMMems[mcName])
+    nFMMem = 8#FIXME after TBHelper is added to the test bench len(FMMems[tpName])
```
bryates, line 98: @aehart is this ok? I basically made this PR hang on #179. If this one is merged first, I can just push this change to the #179, otherwise I can push it here.

aehart:
FYI, I just rebased and removed the commits related to the MatchProcessor, so as to expedite the review for modules in the summer chain.

@bryates I moved the MP commits to the mp_generate_top branch. Please wait until this PR is merged to open a PR for that branch.

meisonlikesicecream:
BTW, do we want to add something about the top level scripts in the README?

emData/generate_PR.py
djcranshaw: @aehart Yes, you're right that this function would return nonsense if the module isn't in the given wiring. This is related to the above comment on modules_to_build as well - I had originally envisioned that variable to just hold the names of modules that we know will build top-levels that work, but maybe the right approach is just to build all modules out of the box, and only add the modules that we want to test into the tcl script. And if we do that, there's no need to list the modules individually in their own variable, they can just be built from the wiring file. This approach would mean that we build some top-levels that don't work, but if we're not testing/using them yet anyways maybe that's fine. Also all the modules needed for the summer chain do work.

emData/generate_PR.py
djcranshaw: Thanks, done.

emData/generate_PR.py
djcranshaw: Thanks, done.

aehart (approved):
Looks good with the most recent changes. Approved!

PR #179, opened 2021-07-06
TBHelper conversion

### LINKED ISSUE #176: Use TBHelper in all C++ test benches
Update all C++ test benches for modules included in the summer chain to use the TBHelper class to access test-vector files.

See slides from 18 March for details on the TBHelper class:
https://indico.cern.ch/event/1010698/#3-generalizing-hls-test-benche
Also, see VMRouter_test.cpp, VMRouterCM_test.cpp, and TrackletCalculator_test.cpp for real-life examples of how to use the class.

aehart:
Resolves #176 for the TrackBuilder. Other modules can push to this PR.

### DIFF
--- a/TrackletAlgorithm/MatchCalculator.h
+++ b/TrackletAlgorithm/MatchCalculator.h
@@ -816,49 +816,49 @@ void MatchCalculator(BXType bx,
       switch (projseed) {
         case 0:
         if(FMMask<LAYER, PHISEC, TF::L1L2>()) {
-          fullmatch[0].write_mem(bx,bestmatch,nmcout1); // L1L2 seed
+          fullmatch[FMCount<LAYER, PHISEC, TF::L1L2>()].write_mem(bx,bestmatch,nmcout1); // L1L2 seed
           nmcout1++;
         }
         break;
         case 1:
         if(FMMask<LAYER, PHISEC, TF::L2L3>()) {
-          fullmatch[1].write_mem(bx,bestmatch,nmcout2); // L2L3 seed
+          fullmatch[FMCount<LAYER, PHISEC, TF::L2L3>()].write_mem(bx,bestmatch,nmcout2); // L2L3 seed
           nmcout2++;
         }
         break;
         case 2:
         if(FMMask<LAYER, PHISEC, TF::L3L4>()) {
-          fullmatch[2].write_mem(bx,bestmatch,nmcout3); // L3L4 seed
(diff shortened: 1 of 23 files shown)


TrackletAlgorithm/MatchCalculator.h
```diff
@@ -816,49 +816,49 @@ void MatchCalculator(BXType bx,
...
         if(FMMask<LAYER, PHISEC, TF::L3L4>()) {
-          fullmatch[2].write_mem(bx,bestmatch,nmcout3); // L3L4 seed
+          fullmatch[FMCount<LAYER, PHISEC, TF::L3L4>()].write_mem(bx,bestmatch,nmcout3); // L3L4 seed
           nmcout3++;
         }
         break;
         case 3:
         if(FMMask<LAYER, PHISEC, TF::L5L6>()) {
-          fullmatch[3].write_mem(bx,bestmatch,nmcout4); // L5L6 seed
+          fullmatch[FMCount<LAYER, PHISEC, TF::L5L6>()].write_mem(bx,bestmatch,nmcout4); // L5L6 seed
```
bryates, line 837: This is a some what complicated solution I came up with to count how many slots are not masked. That way, seeding in L5L6 (which would be slot 3) is placed in the correct slot (slot 2 if the TBHelper is only reading L1L2 and L5L6). This relies on the mask to agree with the number of files TBHelper loads, which seems reasonable.

aehart: @bryates I'm not sure I understand what's happening here. I thought the `FMMask` essentially told you which FM memories are "on" for a given MC. That way, the size of the `fullmatch` array can be the same for every MC, but HLS will only generate ports for the memories that are enabled by the mask.

So why would the index of, e.g., the L5L6 FM memory ever change?

bryates: `FMMask` disables ports so Vivado will trim them, but this does not change the index, which was hard coded to 3 in the past. `FMCount` allows it to determine how many ports are actually active, and allocate the correct spot. I needed this because the `TBHelper` uses vectors that are only as large as the active ports.

aehart:
FYI, I just rebased and removed the commits related to the MatchProcessor, so as to expedite the review for modules in the summer chain.

@bryates I moved the MP commits to the mp_tb_helper branch. Please wait until this PR is merged to open a PR for that branch.

pwittich (approved):

PR #182, opened 2021-07-21
Modify Macros.h to use X-Macros

aperloff:
Try to reduce the amount of un-compiled code in TrackletCalculator_test.cpp and also simplify the code. This PR makes the following changes:
1. Chang the preprocessor macro module names into enums for use at runtime using X-Macros. This style of programming is described [here](https://www.geeksforgeeks.org/x-macros-in-c/) and [here](https://stackoverflow.com/questions/11714325/how-to-get-enum-item-name-from-its-value). The upside is that the enums, names, and values are now all available at runtime. The downside is that the preprocessor macros themselves are no longer available. That is why there is a huge reduction in the number of preprocessor macros used in TrackletCalculator_test.cpp.
2. In order to remove the dependence on the name of the top function, this PR now makes the tcl script pass in the name of the top function to the test bench. It is difficult to impossible to form the needed name using either preprocessor macros or function pointers. This solution seemed both elegant and easy.


pwittich:
I learned something new. The x-macros are interesting. Overall it doesn't look like it reduces the amount of code, but it does organize it in a useful way?

aehart:
I actually really like this. Once all the the modules have a generate_*.py script and we have test-vectors for the reduced config, I think it will be useful to have the ability to test many module instances with the same test bench, as is done now for the TC. And these changes make that a lot easier.

I just have two suggestions/questions:
- Since the modules are now in an `enum`, which automatically assigns indices, we should probably remove the "module index" field of the X-list.
- Can you think of any way we could return the top function dynamically (either at compile-time or run-time)?
<br/>I think it would have to be done at compile-time because the top functions have different signatures, which means the pointers to the top functions have different types and precludes putting them all in a common array without some ill-advised casting. So I guess we would need something like a preprocessor macro that undergoes, e.g., the following replacement during preprocessing: `TOP_FUNC_(TC_L1L2E_)` → `TrackletCalculator_L1L2E`.
<br/>Now, we could add the top function names to our X-list, but I'm not sure how to extract them with a macro like this. Maybe it's impossible with this kind of structure, or maybe I just lack imagination. Any thoughts?

aperloff:
@pwittich @aehart 
1. I have updated the PR to remove the redundant indices.
2. I tried for a long time to dynamically generate the top function before realizing it was hopeless using X macros. The X macro are stored using a `define` preprocessor macro, but the resulting enum and strings are accessed at runtime using the accessor functions. The enums and strings are not available at compile time. Unfortunately the top function name needs to be available at compile time. So the only way to dynamically generate the names is to use preprocessor macros and build up the full top function name or pass it in from the tcl scripts. We could store the list of top functions for each module in the X macro definition, but that is less dynamic and not my favorite solution. I don't mind passing in the top function because we have to us CCFLAGS to pass in things like seeds and modules anyway. One function name doesn't hurt, especially because this method greatly simplifies the code.

TestBenches/ProjectionRouter_test.cpp
```diff
@@ -28,47 +31,12 @@ int main()
...
-  const auto projMemType = BARREL2S;
-#else
-  #error "Undefined Module"
-#endif
-
-#if MODULE_ == PR_L1PHIB_ || MODULE_ == PR_L2PHIB_ || MODULE_ == PR_L3PHIB_ || MODULE_ == PR_L4PHIB_ || MODULE_ == PR_L5PHIB_ || MODULE_ == PR_L6PHIB_ || MODULE_ == PR_L1PHIC_ || MODULE_ == PR_L2PHIC_ || MODULE_ == PR_L3PHIC_ || MODULE_ == PR_L4PHIC_ || MODULE_ == PR_L5PHIC_ || MODULE_ == PR_L6PHIC_
+  assert(MODULE_ >= 413 && MODULE_ <= 440); // Select for PR modules
+  assert(std::string(module_name[MODULE_]).find("PHIB") != string::npos ||
+         std::string(module_name[MODULE_]).find("PHIC") != string::npos); // Select for PHIB and PHIC modules
+  const auto projMemType = (MODULE_ >= 413 && MODULE_ <= 428) ? BARRELPS : BARREL2S;
```
aehart, line 37: @aperloff Is there any reason we can't replace the magic numbers here with the members of the `Module` enum? For example:
```c++
  assert(MODULE_ >= PR_L1PHIA_ && MODULE_ <= PR_L6PHID_); // Select for PR modules
```

aperloff: Fixed

aehart (approved):
Looks good. Approved!

PR #183, opened 2021-07-22
Reconfigure the MatchEngine code for easier top function instantiation

aperloff:
This PR does the following:
1. Splits out the top function for the match engine into its own header and cc files. It also moves the main MatchEngine function into the MatchEngine.h header file. Now there are three files, but the files which instantiate the top function are very simple.This will make it easier to automatically generate the files needed to instantiate multiple ME modules.
2. Splits out the preprocessor macro into its own file for common use.

I ran csim, csynth, and co-sim before and after making these modifications. There were no changes to the timing, latency, or resource utilization. Only some changes to the names of the resources, which is unimportant.


TrackletAlgorithm/MatchEngine.h
```diff
@@ -104,18 +67,225 @@ constexpr unsigned int kZAdjustment = 8;
...
+	#elif LAYER == 3
+#include "../emData/ME/tables/METable_L3.tab"
+	#elif LAYER == 4
+#include "../emData/ME/tables/METable_L4.tab"
+	#elif LAYER == 5
+#include "../emData/ME/tables/METable_L5.tab"
+	#elif LAYER == 6
+#include "../emData/ME/tables/METable_L6.tab"
+	#else
+{};
```
aehart, line 85: @aperloff It would make more sense to me to make `LAYER` a template parameter of this function, and to not define the values of `LAYER` and `PROJECTIONTYPE` in this file. For all the processing modules, we want the code outside the \*Top.\* files to be as generic as possible.

TrackletAlgorithm/MatchEngine.h
```diff
@@ -104,18 +67,225 @@ constexpr unsigned int kZAdjustment = 8;
...
+	#endif
+
+	for (int i=0;i<LSIZE;i++){
+		table[i]=tmp[i];
+	}
+}
 
 //template<int L, regionType VMSMEType>
-template<int L, int VMSMEType>
+template<int L, int VMSMEType, int VMPMEType>
```
aehart, line 94: @aperloff I guess we need `VMPMEType` to differentiate between the barrel and the disks, but do we actually need `VMSMEType`? It looks like it can be determined from the value of `L`.

TrackletAlgorithm/Macros.h
```diff
@@ -0,0 +1,43 @@
...
+				  + IS_REPRESENTIBLE_IN_D_BITS(27, N)    \
+				  + IS_REPRESENTIBLE_IN_D_BITS(28, N)    \
+				  + IS_REPRESENTIBLE_IN_D_BITS(29, N)    \
+				  + IS_REPRESENTIBLE_IN_D_BITS(30, N)    \
+				  + IS_REPRESENTIBLE_IN_D_BITS(31, N)    \
+				  + IS_REPRESENTIBLE_IN_D_BITS(32, N)    \
+				  )                                      \
+	  )
+
+#endif
```
aehart, line 43: @aperloff Please add a newline at the end of the file.

TrackletAlgorithm/MatchEngineTop.h
```diff
@@ -0,0 +1,11 @@
...
+#define TrackletAlgorithm_MatchEngineTop_h
+
+#include "MatchEngine.h"
+
+void MatchEngineTop(const BXType bx, BXType& bx_o,
+					const VMStubMEMemory<MODULETYPE, NBITBIN>& inputStubData,
+					const VMProjectionMemory<PROJECTIONTYPE>& inputProjectionData,
+					CandidateMatchMemory& outputCandidateMatch);
+
+#endif
```
aehart, line 11: @aperloff Please add a newline at the end of the file.

TrackletAlgorithm/MatchEngine.h
aperloff: It's needed for the template of one of the parameters in the `MachEngine` function (https://github.com/cms-L1TK/firmware-hls/pull/183/files#diff-8c6bdc6c6e457d4f237d95aeb3341adffd3be2757f53f1744c72ea6852ce4af8R96). Unless this is pre-computed in Constants.h it needs to be a template parameter. Would it be okay with you if we defined the following at https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/Constants.h#L132?
```c++
// List of layer and disk types
constexp unsigned int layerDiskType[N_LAYER + N_DISK] = {BARRELPS, BARRELPS, BARRELPS, BARREL2S, BARREL2S, BARREL2S, DISK, DISK, DISK, DISK, DISK};
```

TrackletAlgorithm/Macros.h
aperloff: Done

TrackletAlgorithm/MatchEngineTop.h
aperloff: Done

TrackletAlgorithm/MatchEngine.h
aehart: Yeah I think that's OK. But I also think keeping `VMPMEType` as a parameter is OK. I just didn't notice that it's used in the function arguments.

TrackletAlgorithm/MatchEngine.h
aperloff: Okay, I have changed this (and a lot more). Now the only user configurable parameter is the layer/disk number. Everything else is computed from this. This function is now templated and specialized as needed.

TrackletAlgorithm/MatchEngine.h
aperloff: Done

aperloff:
@aehart I have made some major changes to this PR. I think it would be good for you to review it again. I have tested that all of the layers work. The disks will not work until #187 is resolved and a few changes are made to `MatchEngine.h`. These changes will all be needed to make the `generate_ME.py` script work.

aperloff:
@aehart I have rebased this PR in light of #189 being merged. Also, the failure had to do with some transient issue in GitLab. The [pipeline#2929251](https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/2929251) actually passed when I restarted the Toptf-synth stage.

aperloff:
I'm just re-running the jobs because it was bugging me that it was marked as failing when everything was actually okay.

aehart (approved):
This looks good to me, especially in light of the additional changes in PR #193. Approved!

PR #185, opened 2021-07-26
IR  top level generation script in emData

sseifeln:
@aehart @tomalin - I've created the python script based on what was done for the TC .. but now I see that the VMR also has the option to generator for one/all... should I add something similar for the IR or is this enough?


sseifeln:
Sorry - I created the pull request to share the code (still need to modify the test bench .. but since I also have to update that I wanted to check about the generation script first)

TrackletAlgorithm/InputRouterTop.h
```diff
@@ -1,25 +0,0 @@
...
-
-
-// at the moment I'm changing this for each link 
-// is that really the only way to do this? 
-
-// only for PS10G_3_A for now 
-// other cases to be added 
-// when I have final link map
-//constexpr unsigned int cTest = (const)(std::atoi(getenv ("TEST")));
-constexpr unsigned int cNMemories = 7;
```
aehart, line 14: @sseifeln The C/RTL cosimulation currently crashes with a seg-fault, and the reason seems to be that this constant has changed. Looking in the InputRouter_parameters.h that is generated by generate_IR.py, I see that now
```c++
constexpr unsigned int cNMemories_IR_DTC_PS10G_3_A = 14;
```
Manually setting this back to 7 allows the cosimulation to run and pass.

Any idea why this constant changed?

sseifeln: Yeah I see where the mistake is now (not looking for an exact match when calculating the constant in the script) - just trying to fix it and will commit

aehart (approved):
Looks good to me. Approved!

PR #186, opened 2021-08-05
Update README.md for emulation instructions

tomalin:
Updated instructions for running fw_sync emulation. 
1) Instruction to change maxstepoffset_ to zero no longer needed, as this is the default.
2) New instruction to change IR truncation maxstep_ to 108, since IR in current FW only runs at 240MHz (vs 360MHz ultimately).
3) New instruction to change TE truncation maxstep_ to 107, though this will soon be the default.

### DIFF
--- a/README.md
+++ b/README.md
@@ -87,7 +87,7 @@ git cms-checkout-topic -u cms-L1TK:fw_synch_210611
 git clone https://github.com/cms-data/L1Trigger-TrackFindingTracklet.git L1Trigger/TrackFindingTracklet/data
 ```
 
-A few cfg changes were made in order to adjust truncation, output test vectors and lookup tables, and to disable multiple matches in the MatchCalculator. This required editing parameter values in L1Trigger/TrackFindingTracklet/interface/Settings.h to match the following excerpts:
+A few cfg changes were made in order to output test vectors & lookup tables, adjust truncation, and to disable multiple matches in the MatchCalculator. This required editing parameter values in L1Trigger/TrackFindingTracklet/interface/Settings.h to match the following excerpts:
 
 ```c++
 …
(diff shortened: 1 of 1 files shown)


aehart:
@tomalin I'd like the instructions to correspond with the actual files downloaded by emData/download.sh. So could you move the instruction regarding changing `maxstep_` to a separate subsection for "future test-vectors"? Whenever the actual files are updated, then we can merge that instruction into the main recipe.

tomalin:
> @tomalin I'd like the instructions to correspond with the actual files downloaded by emData/download.sh. So could you move the instruction regarding changing `maxstep_` to a separate subsection for "future test-vectors"? Whenever the actual files are updated, then we can merge that instruction into the main recipe.

I'm concerned that if I move the maxstep instruction to a completely separate section, people generating new files might overlook it. Instead I've put it in a separate paragraph, with a sentence beforehand clearly explaining that this was not used for the existing emData/ files should is recommended for making new ones. Good compromise?

tomalin:
I've also deleted the instructions to create by hand with mkdir data/MemPrints/ data/LUTs/ etc. , since this version of CMSSW already made these automatically.

aehart:
> > @tomalin I'd like the instructions to correspond with the actual files downloaded by emData/download.sh. So could you move the instruction regarding changing `maxstep_` to a separate subsection for "future test-vectors"? Whenever the actual files are updated, then we can merge that instruction into the main recipe.
> 
> I'm concerned that if I move the maxstep instruction to a completely separate section, people generating new files might overlook it. Instead I've put it in a separate paragraph, with a sentence beforehand clearly explaining that this was not used for the existing emData/ files should is recommended for making new ones. Good compromise?

That's a good compromise. I think you just need to remove the `maxstep_` excerpt from the main code block, so that it only appears after your "N.B. In future …" text.

tomalin:
> > > @tomalin I'd like the instructions to correspond with the actual files downloaded by emData/download.sh. So could you move the instruction regarding changing `maxstep_` to a separate subsection for "future test-vectors"? Whenever the actual files are updated, then we can merge that instruction into the main recipe.
> > 
> > 
> > I'm concerned that if I move the maxstep instruction to a completely separate section, people generating new files might overlook it. Instead I've put it in a separate paragraph, with a sentence beforehand clearly explaining that this was not used for the existing emData/ files should is recommended for making new ones. Good compromise?
> 
> That's a good compromise. I think you just need to remove the `maxstep_` excerpt from the main code block, so that it only appears after your "N.B. In future …" text.

DONE

aehart (approved):
Looks good. Thanks, @tomalin.

PR #189, opened 2021-08-10
DummyMessageLogger

ryanm124:
Replaced cout statements with message logger. Created DummyMessageLogger that couts statements using the same syntax as the logger. DummyMessageLogger.h or MessageLogger.h is conditionally compiled depending on whether running in CMSSW or not but print syntax remains the same.

### DIFF
--- a/TrackletAlgorithm/Constants.h
+++ b/TrackletAlgorithm/Constants.h
@@ -5,7 +5,6 @@
 
 #ifdef CMSSW_GIT_HASH
 #include "L1Trigger/TrackFindingTracklet/interface/Settings.h"
-using namespace trklet;
 #endif
 
 // Inline function to convert floating point values to integers, given a
@@ -66,12 +65,14 @@ constexpr double c = 0.299792458; // m/ns
 
 // detector constants
 #ifndef CMSSW_GIT_HASH
+namespace trklet{
   constexpr int N_LAYER = 6; // # of barrel layers assumed
   constexpr int N_DISK = 5; // # of endcap disks assumed
+}
 #endif
 constexpr double bfield = 3.8112; // T
-constexpr int rmean[N_LAYER + N_DISK] = { 851, 1269, 1784, 2347, 2936, 3697,   -1,   -1,   -1,   -1,   -1 }; // valid for layers
-constexpr int zmean[N_LAYER + N_DISK] = {  -1,   -1,   -1,   -1,   -1,   -1, 2239, 2645, 3163, 3782, 4523 }; // valid for disks
+constexpr int rmean[trklet::N_LAYER + trklet::N_DISK] = { 851, 1269, 1784, 2347, 2936, 3697,   -1,   -1,   -1,   -1,   -1 }; // valid for layers
+constexpr int zmean[trklet::N_LAYER + trklet::N_DISK] = {  -1,   -1,   -1,   -1,   -1,   -1, 2239, 2645, 3163, 3782, 4523 }; // valid for disks
 constexpr double zlength = 120.0; // cm
 constexpr double rmindiskvm = 22.5; // cm
 constexpr double rmindisk = 20.0; // cm
@@ -83,7 +84,7 @@ constexpr double kphi = 7.71867e-06;
 constexpr double kz = 0.0585938;
 
 // stub digitization constants for combined modules
-constexpr double kz_cm[N_LAYER] = {2*zlength/(1<<12), 2*zlength/(1<<12), 2*zlength/(1<<12), 2*zlength/(1<<8), 2*zlength/(1<<8), 2*zlength/(1<<8)};
+constexpr double kz_cm[trklet::N_LAYER] = {2*zlength/(1<<12), 2*zlength/(1<<12), 2*zlength/(1<<12), 2*zlength/(1<<8), 2*zlength/(1<<8), 2*zlength/(1<<8)};
 
 // tracklet digitization constants
 constexpr double krinv = 1.04549e-06;
@@ -110,17 +111,19 @@ constexpr double z0cut = 15.0; // cm
 
 // cut constants for combined modules
 #ifndef CMSSW_GIT_HASH
+namespace trklet{
 constexpr double VMROUTERCUTZL2 = 50.0;      //Min L2 z for inner allstub in cm
 constexpr double VMROUTERCUTZL1L3L5 = 95.0;  //Max z for inner barrel layers in cm
 constexpr double VMROUTERCUTZL1 = 70.0;      //Max z for L1 barrel seeding in cm
 constexpr double VMROUTERCUTRD1D3 = 55.0;    //Max r for disk seeds in cm
+}
 #endif
 
 // various bit widths
-constexpr unsigned nbitsallstubs[N_LAYER + N_DISK] = {3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2};
+constexpr unsigned nbitsallstubs[trklet::N_LAYER + trklet::N_DISK] = {3, 2, 2, 2, 2, 2, 2, 2, 2, 2, 2};
 constexpr unsigned int nbits_maxvm = 5; // number of bits needed for max number of VMs per layer/disk (max number is 32)
 // number of bits used to distinguish VMs in one allstub block for each layer
-constexpr unsigned int nbits_vmmeall[N_LAYER + N_DISK]={2,3,3,3,3,3,3,2,2,2,2};
+constexpr unsigned int nbits_vmmeall[trklet::N_LAYER + trklet::N_DISK]={2,3,3,3,3,3,3,2,2,2,2};
 
 // List of regions for memory template parameters
 enum regionType {BARRELPS, BARREL2S, BARRELOL, BARREL, DISKPS, DISK2S, DISK, BARREL_FOR_MC, DISK_FOR_MC};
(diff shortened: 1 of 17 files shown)


pwittich:
@ryanm124 can you take a look at the failing CI (you can see it here [on the CERN Gitlab](https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/2910604)). The ones I see look to be of the flavor

```
/home/gitlab-runner/builds/biTwR5NQ/0/cms-l1tk/firmware-hls/project/../TrackletAlgorithm/DummyMessageLogger.h:8:40: error: converting integer literal to bool, use bool literal instead [modernize-use-bool-literals,-warnings-as-errors]
    LogVerbatim(std::string type) { if(0) {std::cout<<type<<std::endl;} }
                                       ^~
                                       false
```
Unclear to me why you want to have an `if (0)` in your code anyhow?

TrackletAlgorithm/DummyMessageLogger.h
```diff
@@ -0,0 +1,40 @@
+#ifndef MessageLogger_MessageLogger_h
+#ifndef L1Trigger_TrackFindingTracklet_interface_Logger_h
+#define L1Trigger_TrackFindingTracklet_interface_Logger_h
+#include<iostream>
+namespace edm {
+  class LogVerbatim{
+  public:
+    LogVerbatim(std::string type) { if(0) {std::cout<<type<<std::endl;} }
```
pwittich, line 8: why do you have these odd constructors here?

pwittich:
I see you chose "FWHLS" for the message logger tag. I would think something like "L1TK" might be more descriptive for a CMSSW user. How did you come to choose FWHLS?

ryanm124:
The if(0) must've been used for bug checking in the past and shouldn't be included here, so I'll make a new commit and remove them. Other than that, the constructors cout the tag given to the dummy logger to mimic what the real message logger does. The tag, L1track, is already used by the TrackFindingTMTT messages and I used "HLS" for a related cout issue for TrackFindingTrackletHLS. I figured this slightly different tag would allow users to filter out the firmware messages but I'm happy to change it if you have any alternatives.

pwittich:
@tomalin re message logger output, what do you think we should do here? the current tag ("FWHLS") is probably incomprehensible for non-experts.

pwittich:
@ryanm124 the CI is still failing; can you take a look and figure out what the failure is. A quick glance suggests that in some places the message logger/cout code is getting into the synthesis part of the build and hence failing.

tomalin:
> @tomalin re message logger output, what do you think we should do here? the current tag ("FWHLS") is probably incomprehensible for non-experts.

I'd recommend using category name "L1trackHLS". (N.B. The category name should be short, and used to specify a class of messages we want to switch on or off together. It will be accompanied by a specific longer-printout for each individual MessageLogger print statement).

TrackletAlgorithm/Constants.h
```diff
@@ -4,8 +4,8 @@
 #include "ap_int.h"
 
 #ifdef CMSSW_GIT_HASH
-  #include "Settings.h"
-  using namespace trklet;
+#include "L1Trigger/TrackFindingTracklet/interface/Settings.h"
+using namespace trklet;
```
tomalin, line 8: During CMSSW code reviews, we're told never to put "using namespace" in a header file. Such statements must only appear in .cc files, since otherwise they can affect multiple files.

tomalin:
All header files should have a comment at the top, explaining what they are for. So please add one to DummyMessageLogger.h

TrackletAlgorithm/DummyMessageLogger.h
```diff
@@ -0,0 +1,40 @@
+#ifndef MessageLogger_MessageLogger_h
```
tomalin, line 1: What is the point of the line "#ifndef MessageLogger_MessageLogger_h", given that this pragma is never defined anywhere?

TrackletAlgorithm/DummyMessageLogger.h
```diff
@@ -0,0 +1,40 @@
...
+#ifndef L1Trigger_TrackFindingTracklet_interface_Logger_h
```
tomalin, line 2: Why does this line "#ifndef L1Trigger_TrackFindingTracklet_interface_Logger_h" need the CMSSW directory path, given that DummyMessageLogger is not part of CMSSW?

tomalin:
Please confirm that you've check that compilation with VIvado HLS still works after these changes. Also, I don't believe we need to #include either MessageLogger.h or DummyMessageLogger.h if __SYNTHESIS__ is defined. Also, we can we surround all the code in DummyMessageLogger.h by #ifndef __SYNTHESIS ... #endif, to make clear that none of this code is used during HLS synthesis?

TrackletAlgorithm/DummyMessageLogger.h
ryanm124: It was an attempt to use both the MessageLogger, which defines this macro, and my dummy but it's not needed now so I'll remove it.

TrackletAlgorithm/DummyMessageLogger.h
ryanm124: Good catch. I'll remove the path and update the name.

TrackletAlgorithm/Constants.h
ryanm124: The namespace line was pushed by someone else, but I can remove it if that's appropriate to do for this issue.

tomalin: It should be removed, but you may then have then add it to .cc files to make the code compile in CMSSW.

ryanm124:
I've committed your suggestions and all the Vivado checks done automatically by github have passed. For further checks with Vivado, someone would have to pull my code as my institution does not currently have a Vivado license.

pwittich:
> I've committed your suggestions and all the Vivado checks done automatically by github have passed. For further checks with Vivado, someone would have to pull my code as my institution does not currently have a Vivado license.

the CI at CERN ([this one here](https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/2921243)) does the CI with vivado so it should be fine. 
I presume you checked that this also works fine in the emulation?

ryanm124:
I checked that I could compile and run over an event with TrackFindingTrackletHLS/config,py which handles the DTC, IR, and VMR which worked fine. Is there anything else that needs to be checked?

TrackletAlgorithm/Constants.h
```diff
@@ -4,8 +4,7 @@
 #include "ap_int.h"
 
 #ifdef CMSSW_GIT_HASH
-  #include "Settings.h"
```
pwittich, line 7: According to the [CMSSW style guide](http://cms-sw.github.io/cms_coding_rules.html#3--style-rules-1) please move all preprocessor directives to the left margin.

TrackletAlgorithm/InputRouter.h
```diff
@@ -250,7 +255,7 @@ void InputRouter( const BXType bx
 	  if( hVldBt == 0 ){ 
 	  	#ifndef __SYNTHESIS__
```
pwittich, line 256: same comment re indents. I realize this is not in code that you changed (?) but let's take care of this now too.

TrackletAlgorithm/TrackletCalculator.h
```diff
@@ -279,7 +279,7 @@ TC::barrelSeeding(const AllStub<InnerRegion<Seed>()> &innerStub, const AllStub<O
   );
 
 // Determine which layer projections are valid.
-  valid_proj: for (ap_uint<3> i = 0; i < N_LAYER - 2; i++) {
+ valid_proj: for (ap_uint<3> i = 0; i < trklet::N_LAYER - 2; i++) {
```
pwittich, line 282: I think indent got screwed up here? I realize this is a label....

TrackletAlgorithm/InputRouter.h
ryanm124: Sure thing, should I do this for the whole file? There are several #pragma lines that aren't in the left margin.

tomalin:
1) https://github.com/cms-L1TK/firmware-hls/blob/featureLog/TrackletAlgorithm/ProjectionRouter.h#L195 calls LogVerbatim() if "#ifndef __SYNTHESIS"
2) https://github.com/cms-L1TK/firmware-hls/blob/featureLog/TrackletAlgorithm/MatchEngine.cc#L73 calls it if "#ifdef DEBUG", where DEBUG can optionally be set by the compiler during Vivado HLS CSIM.
3) https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/InputRouter.h#L251 calls if if "#ifdef IR_DEBUG and #ifdef __SYNTHESIS__" .

Messages of type (1) will always be written during CSIM, and optionally during CMSSW. 
Messages of type (2) can optionally be written during CSIM, but could only be written during CMSSW if a much more unusual pragma were defined than "DEBUG", since adding -DDEBUG to the CMSSW compiler options would switch on many things ...
Approach (3) gives same functionality as approach (2). The code is longer, but the added ifdef __SYNTHESIS might improve clarity.

Not for this PR, but I suggest you ask around, and converge on common style.

ryanm124:
I've just pushed the changes to the indentation. Regarding the message types, maybe we can discuss that during a HLS meeting and create a new issue?

tomalin (approved):
Looks good to me now.

pwittich:
> I've just pushed the changes to the indentation. Regarding the message types, maybe we can discuss that during a HLS meeting and create a new issue?

please open a new issue

PR #190, opened 2021-08-13
Update VHDL files from script

meisonlikesicecream:
Updated the firmware repo with the latest VHDL files created by the project generation scripts, including the new testbenches.


tomalin:
Please state how the new code was validated.

meisonlikesicecream:
> Please state how the new code was validated.

The errors are the same as before for the PRMEMC and IRVMR chain. Don't know much about the errors of the TETC, but the number of errors are the same.

tomalin (approved):
Largely cosmetic updates with latest scripts.

PR #191, opened 2021-08-13
FileWriterFromRAM bug fixes

meisonlikesicecream:
Fixed two bugs in FileWriterFromRAM.vhd that I noticed when I was debugging the VMR in Vivado:

1. NENT was at most 107, thus the 108th stub was never written to the output text files.
2. The output text file could be closed even if the latch wasn't empty, thus the last couple of stubs in the 100th event were never written to the files.

Unclear if FileWriterFromRAMBinned.vhd needs the same changes. Haven't looked into that yet.

### DIFF
--- a/IntegrationTests/common/hdl/FileWriterFromRAM.vhd
+++ b/IntegrationTests/common/hdl/FileWriterFromRAM.vhd
@@ -52,6 +52,7 @@ begin
 procFile : process(CLK)
   variable INIT        : boolean := false; --! File not yet open
   variable FILE_STATUS : file_open_status;
+  variable CLOSE_FILE  : boolean := false; --! File is ready to be closed
   file     FILE_OUT    : text;   
   variable LINE_OUT    : line;                              
   variable BX_CNT      : integer := -1;  --! Event counter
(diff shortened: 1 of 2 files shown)


IntegrationTests/common/hdl/FileWriterFromRAM.vhd
```diff
@@ -96,14 +97,22 @@ begin
...
 
       if (BX_CNT = MAX_EVENTS) then
-        -- All events processed, so close file.
-        file_close(FILE_OUT);
+        -- All events processed, so close file once all stubs has been written to file.
+        CLOSE_FILE := true;
       end if;
+    else 
+      -- update NENT as NENT_ARR might not have finished counting when DONE signal goes high (NENT would never be 108 otherwise)
+      NENT := to_integer(unsigned(NENT_ARR(PAGE)));
```
tomalin, line 110: NENT is now defined both at L102 & L110, so in both halves of the "if ... else ... end if" statement. If this is needed, why not simply define NENT once, outside this "if ... else ... end if" statement?

tomalin:
It seems likely that these fixes are also needed for FileWriterFromRAMBinned.vhd . Please try running the IR-VMR chain with and without these corrections to FileWriterFromRAMBinned.vhd , and if the agreement with emulation doesn't get worse, then include them in this PR.

IntegrationTests/common/hdl/FileWriterFromRAM.vhd
meisonlikesicecream: Ah yes :-)

meisonlikesicecream:
> It seems likely that these fixes are also needed for FileWriterFromRAMBinned.vhd . Please try running the IR-VMR chain with and without these corrections to FileWriterFromRAMBinned.vhd , and if the agreement with emulation doesn't get worse, then include them in this PR.

Adding it to the binned version as well did not seem to change the results for the IRVMR chain.

tomalin (approved):
Thanks. Useful fix.

Issue #187, opened 2021-08-05
New bit needed in VMStubsMEMemory


aperloff:
There is no bit assigned to the VMStubsME memories to distinguish whether a stub is from a 2S or a PS module. This information is needed for the disks. The `isPSmodule` function [[1]](https://github.com/cms-L1TK/cmssw/blob/d76cbdf3c36e90775ee5b062acd42a04a93af3c7/L1Trigger/TrackFindingTracklet/interface/Stub.h#L71) is used in the ME emulation [[2]](https://github.com/cms-L1TK/cmssw/blob/d76cbdf3c36e90775ee5b062acd42a04a93af3c7/L1Trigger/TrackFindingTracklet/src/MatchEngine.cc#L255), but not in the firmware. From [[1]](https://github.com/cms-L1TK/cmssw/blob/d76cbdf3c36e90775ee5b062acd42a04a93af3c7/L1Trigger/TrackFindingTracklet/interface/Stub.h#L71) you can see that the emulation uses the stub `r_` value, which we don't have available in firmware. That is why a new bit needs to be added ahead of time to distinguish the type of stub.

This change will require another emulator change, new tarball, and a modification to the HLS VMRouter to write the extra bit.

@djcranshaw @aryd

aperloff:
As noted in #193, this will need to be settled before the disk MEs will work.

aperloff:
This issue will be resolved/nullified by #202. It turns out we already had enough stub information in order to determine if the stub came from a PS module.

PR #193, opened 2021-08-19
Top level generation script for ME

### LINKED ISSUE #177: Scripts to generate top functions

aperloff:
This resolved #177 for the ME and is similar in spirit to #178. It relies upon #183 and will need to be rebased after that is merged. This PR would also benefit from #179, though that is not strictly required.

There is one open question remaining. Right now the script will create a new top level function for each module in the wiring file (i.e. ME_L3PHIC20). However, the ME code doesn't really change based on the phi sector (PHIC20 in this example). It only depends on the layer/disk. The code could be changed to only generate one module per layer/disk, which would save on time when synthesizing all of the modules. However, this change is not strictly necessary.


aperloff:
@aehart This PR is now ready for review. It has been rebased on top of the master branch. I also implemented a feature in the `generate_ME.py` script to create a single top function for each layer/disk. This is turned on by default in the `download.sh` script.

There are a few features we may want in a future PR:
1. In order for the disk test to be successful #187 will need to be settled.
2. Right now there are specific phi sectors which can be tested due to the hard coding of the test vectors in the test bench. Technically all phi sectors will work, but it's not convenient to test all 288 modules. Implementing #182 might make it easier to allow for the automated testing of more phi sectors.
Neither of these are necessary or should hold up this PR. I just want to put them here for the record.

pwittich:
> @aehart This PR is now ready for review. It has been rebased on top of the master branch. I also implemented a feature in the `generate_ME.py` script to create a single top function for each layer/disk. This is turned on by default in the `download.sh` script.
> 
> There are a few features we may want in a future PR:
> 
> 1. In order for the disk test to be successful [New bit needed in VMStubsMEMemory #187](https://github.com/cms-L1TK/firmware-hls/issues/187) will need to be settled.
> 2. Right now there are specific phi sectors which can be tested due to the hard coding of the test vectors in the test bench. Technically all phi sectors will work, but it's not convenient to test all 288 modules. Implementing [Modify Macros.h to use X-Macros #182](https://github.com/cms-L1TK/firmware-hls/pull/182) might make it easier to allow for the automated testing of more phi sectors.
>    Neither of these are necessary or should hold up this PR. I just want to put them here for the record.

Can you put these comments into new issue(s)? they will disappear when this Pr is closed.

aperloff:
@pwittich I have added a comment to #187 and created issue #197. So these comments will no longer be forgotten.

aehart:
@aperloff This looks very good to me. Can you remove MatchEngineTop.*, since it is now automatically generated?

aperloff:
> @aperloff This looks very good to me. Can you remove MatchEngineTop.*, since it is now automatically generated?

Done, tests running.

aehart (approved):
Also looks good. Approved!

PR #194, opened 2021-08-20
Ir add top level generation scripts

sseifeln:
Modified InputRouter so that it works both in the mater and reduced_config branches (still outstanding - need to modify the test bench to use the TBhelper)


sseifeln:
no longer needed - will wait for merge of reduced_config_pr branch into master

PR #199, opened 2021-09-07
Changes For Future SW Emulation

carriganm95:
Changes are made to FileReadUtility.h in order to work with the new CMSSW future SW emulation memory templates. Previous PR added templates to MemoryTemplate.h, MemoryTemplateBinned.h, etc to store all bunch crossings as the 0th. These fixes to FileReadUtility.h ensure that we always read/write using the 0th bunch crossing when using CMSSW.

### DIFF
--- a/TestBenches/FileReadUtility.h
+++ b/TestBenches/FileReadUtility.h
@@ -115,18 +115,10 @@ void writeMemFromFile(MemType& memory, std::ifstream& fin, int ievt, int base=16
       return;
     } else {
       if (split(line,' ').size()==4) {
-      #ifdef CMSSW_GIT_HASH
-       memory.write_mem(0, line, base);
-       #else
        memory.write_mem(ievt, line, base);
-       #endif
       } else {
 	const std::string datastr = split(line, ' ').back();
-	#ifdef CMSSW_GIT_HASH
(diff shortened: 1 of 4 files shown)


aehart:
@carriganm95 These changes look fine to me, but I wonder if it might be better to move them further upstream to the memory template classes.

For example, these lines
https://github.com/cms-L1TK/firmware-hls/blob/4c37847c3da6d814a5b7faae6335449793348247/TrackletAlgorithm/MemoryTemplate.h#L53-L57
could become
```c++
  DataType read_mem(BunchXingT ibx, ap_uint<NBIT_ADDR> index) const
  {
        if (!NBIT_BX) ibx = 0;
	return dataarray_[ibx][index];
  }
```
Similar changes would be needed in the other memory template classes and in the `write_mem` methods.

This way is a little cleaner I think. Thoughts from you or @pwittich?

pwittich:
there is a lot of copy-paste here, moving it to some place where you reduce the copy/paste and hard-to-read ifdefs would be good.

FWIW the CMS standard is to put all preprocessor directives in column 0, i.e., not indent them. Please fix this in any case

aehart (approved):
Looks good to me. Approved!

Issue #184, opened 2021-07-22
Use 'needs:' keyword to establish job dependencies in the CI/CD pipelines


aperloff:
If we start using the ['needs:' keyword](https://gitlab.cern.ch/help/ci/yaml/README.md#needs) for the CI/CD jobs we may be able to run some jobs more quickly by allowing some jobs to run before their stage is reached. Jobs would then run as soon as their needs relationships are met, which might speed up the pipelines. We would also be able to view the needs relationships between jobs as a [Directed Acyclic Graph (DAG)](https://gitlab.cern.ch/help/ci/directed_acyclic_graph/index.md).

PR #200, opened 2021-09-14
Use python3 for all generate scripts

aperloff:
Switch to using python3 for all of the generate scripts. Python2 is deprecated and should no longer be used. This PR also adds some deterministic sorting to the PR and IR scripts so that they return the same module order every time. The other modules are not affected by this issue as far as I can tell.

This PR was tested by comparing the scripts generated with python2 and python3. All of them matched except for the ones without a deterministic order. For those I did my best to check by hand that specific modules matched.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -21,6 +21,15 @@ stages: # ---------------------------------------------------------------------
   before_script:
     - source $VIVADO_PATH_SH
 
+.job_template: &template_static-check-python
+  tags:
+    - xilinx-tools
+  stage: lint
+  before_script:
+    - which python3
+    - python3 --version
+    - python3 -m pylint --version
+
 .job_template: &template_quality-check
   needs: ["download"]
   tags:
@@ -96,18 +105,16 @@ stages: # ---------------------------------------------------------------------
     
 # Jobs ------------------------------------------------------------------------
 # Lint ------------
-lint:
-  tags:
-    - xilinx-tools
-  stage: lint
-  before_script:
-    - which python3
-    - python3 --version
-    - python3 -m pylint --version
+py3check:
+  <<: *template_static-check-python
   script:
     # If changes need to be made you might refer to: https://portingguide.readthedocs.io/en/latest/tools.html#automated-fixer-python-modernize
     - python3 -m pylint --py3k emData/*.py # Check for python3 compatibility
-    - python3 -m pylint --exit-zero emData/*.py # Static code checker
+lint:
+  <<: *template_static-check-python
+  allow_failure: true
+  script:
+    - python3 -m pylint emData/*.py # Static code checker
 # Download ------------
 download:
   tags:
@@ -239,7 +246,7 @@ ME-vivado-hls-build:
 ME-vitis-hls-build:
   <<: *template_hls-build
   needs: ["download", "ME-quality-check"]
-  allow_failure: true # FIXME: remove after avitis updated to 2021.1 and confirmed that pre-synthesis failure is gone.
+  allow_failure: true # FIXME: remove after vitis updated to 2021.1 and confirmed that pre-synthesis failure is gone.
   variables:
     EXECUTABLE: 'vivado_hls'
     VIVADO_VERSION: "2020.2"
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,6 +6,7 @@ variables:
   CLANG_TIDY_COMP_OPTIONS: '-std=c++11 -I../TrackletAlgorithm -I../TestBenches -I/nfs/data41/software/Xilinx/Vivado/${VIVADO_VERSION}/include'
 
 stages: # ---------------------------------------------------------------------
+  - lint
   - download
   - quality-check
   - hls-build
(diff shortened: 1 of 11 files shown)


aperloff:
@aehart @pwittich I've implemented a python linter (pylint) for the package. I tried to fix a lot of the problems with the generate scripts, but I didn't want to go any further without talking to you. I turned off a lot of the warnings already (https://github.com/cms-L1TK/firmware-hls/pull/200/files#diff-ba293004ae8f18975051e825bd8d5c70137a8b5d42d16e0f7c9354aa634c446cR156-R162) and modified a few of the others (look for `# Default:` in the .pylintrc file. Do you agree with these modifications to the defaults? We could ask all of the authors to simply conform to the PEP standards. What do you think?

aperloff:
One other note, the introduction of the `needs` keyword in the `.gitlab-ci.yml` file addresses #184.

aperloff:
We should decide on a style guide to follow and determine which of `[camelCase, snake_case, PascalCase, UPPER_CASE]` to use for argument, variable, function, class, constant, and global constant names. That is where a lot of the linting errors are coming from.

https://pylint.pycqa.org/en/latest/user_guide/options.html

pwittich:
I think you should consider changing it to run two separate jobs, one for the pylint and one for the HLS job on gitlab, to keep the two cleaner/separate. 

also for now I'd suggest only running the py3 tests on pylint, and not try to impose another style. Personally I don't think we need to get that prescriptive.

pwittich:
One thing I think at some point I would like to encourage is using py typing
https://docs.python.org/3/library/typing.html

aperloff:
@pwittich
1. Why do you think we should use type hints? As far as I can tell none of our types or return statements are that complex, not will it add much of a benefit to the generate scripts where the ultimate goal isn't to return a variable, but to write a file. Have you used this feature before and found it helpful?
2. We already have some static checking taking place on GitLab with the clang-tidy (possibly clang-format) checks. I figured the linting and static checks for python could also take place there. You're right that they don't need to take place on GitLab, but why separate the work? The linting takes seconds to complete and probably won't produce too many errors after the initial cleanup. Why do you say running the linting on GitHub would be cleaner?
3. I understand the sentiment of not wanting to impose more work when it won't functionally change the code. That said, I do support having a basic coding style that we can follow. Besides making the code look uniform even when written by many authors, it also makes the various pieces of code easier to read. Then our brains don't have to adjust when reading code from two different authors. It also increases the maintainability of the code. One last point is that the static code checker also makes recommendations on how to modify the code to avoid introducing bugs. All of this put together makes the initial annoyance of setting a style seem fairly minor, at least to me.

pwittich:
@aperloff 

RE typing, I think it is good practice and avoids confusion. Also it will allow some more modern IDEs to help with writing code and checking code as you write it

Re splitting it into parts, I guess I think it is just cleaner to have different functionality in different CI jobs rather than have one gigantic mother of all jobs. I don't feel strongly. 

Re the last point, I understand the sentiment; for me that's never been that important as long as the code is written well enough. We are limited by our personnel's time; this doesn't seem like a good way to spend it. (I also think CMS wastes a colossal amount of time trying to make our papers all 'sound the same' -- as if someone some day were going to sit down and read them all in one sitting and might be put off by a chance in tone.)

aehart:
@aperloff @pwittich

Maybe a decent compromise is to add the `--exit-zero` option when we run the static checker:
```
    --exit-zero         Always return a 0 (non-error) status code, even if
                        lint errors are found. This is primarily useful in
                        continuous integration scripts.
```

Then as people occasionally make updates to these scripts, we can use the results of the static checker to encourage them to improve their code quality during the subsequent reviews without it being mandatory for the CI tests to pass. This way, we don't have the up-front manpower cost, but the code quality still improves over time, especially if these scripts are maintained for an extended period.

pwittich:
@aehart that sounds good, we could run once with the py3k checks and enforce that (to avoid regression to py2) and the other one in a more advisory way.

pwittich:
> 2. We already have some static checking taking place on GitLab with the clang-tidy (possibly clang-format) checks. I figured the linting and static checks for python could also take place there. You're right that they don't need to take place on GitLab, but why separate the work? The linting takes seconds to complete and probably won't produce too many errors after the initial cleanup. Why do you say running the linting on GitHub would be cleaner?

I think the reason the clang-tidy ended up where it did was because it needed the headers from vivado, but I could be wrong. also I wasn't as versed in CI as I am now (not that I'm hugely versed) and github actions have evolved.

aperloff:
I've added the `--exit-zero`, so now the linting is a non-blocking check. Alternatively I could split the linting and the py3 compatibility into two jobs and set it so that the linting is allowed to fail. This way there is a graphical way to see if all of the linting passes or if there are messages we should look at. Right now it will show a green check no matter what.

Do you have a preference?

pwittich:
> I've added the `--exit-zero`, so now the linting is a non-blocking check. Alternatively I could split the linting and the py3 compatibility into two jobs and set it so that the linting is allowed to fail. This way there is a graphical way to see if all of the linting passes or if there are messages we should look at. Right now it will show a green check no matter what.
> 
> 
> 
> Do you have a preference?

Mild preference to 2nd option but only if it's zero work

aehart:
@aperloff This looks really good. The only further suggestion I have is to make the "download" job depend on both of the lint jobs. That way, the Python scripts, which are run via emData/download.sh, are not run until the lint jobs pass.

aehart (approved):
Approved!

PR #202, opened 2021-09-17
Update MatchEngine so the disk modules will work

aperloff:
These changes allow the disk modules to be successfully run. Needed to add an isPSmodule function for the disks and make sure there are enough read addresses for the disks. Also added some debugging information that I found useful.

This PR resolves the need for #187. It turns out that we already had enough information to test if a stub came from a PS module.

I have tested that the ME now works for at least one phi sector in every layer and disk.


aehart (approved):
The changes are pretty straightforward and well written. Approved!

PR #203, opened 2021-09-23
ME test bench improvements

aperloff:
Make sure the ME test bench can run any module without a lot of if statements. Previously, each layer/disk had a specific module which could be run; selected via an if statement. Now the test bench can run any module based on what is entered in the tcl script. I think this removes the last vestiges of hard coded module names in the ME code. The typical caveat remains that only modules with the correct links in the emData folder will work.

### DIFF
--- a/TestBenches/MatchEngine_test.cpp
+++ b/TestBenches/MatchEngine_test.cpp
@@ -36,7 +36,7 @@ constexpr auto kLayerDisk = static_cast<TF::layerDisk>(KLAYERDISK);
 #endif
 template<TF::layerDisk LayerDisk> constexpr int getLayerDiskNumber() {
 	// Convert from the enum index (0-based) to the module name indexing (1-based)
-  return LayerDisk <= TF::L6 ? LayerDisk + 1 : LayerDisk - 5;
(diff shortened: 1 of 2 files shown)


TestBenches/MatchEngine_test.cpp
```diff
@@ -27,9 +28,16 @@ using namespace std;
...
 constexpr auto kLayerDisk = static_cast<TF::layerDisk>(KLAYERDISK);
+#if !defined KMODULE
+	#define KMODULE ME_L3PHIC20_
+#endif
 #if !defined TOPFUNCTION
   #define TOPFUNCTION MatchEngineTop_L3
 #endif
+template<TF::layerDisk LayerDisk> constexpr int getLayerDiskNumber() {
+	// Convert from the enum index (0-based) to the module name indexing (1-based)
+  return LayerDisk <= TF::L6 ? LayerDisk + 1 : LayerDisk - 5;
```
aehart, line 39: @aperloff This is super nitpicky, but could you replace `5` with `trklet::N_LAYER - 1`? The constant is defined, so we might as well use it.

aperloff: Done

aehart (approved):
Looks good. Approved!

PR #206, opened 2021-09-29
VMRCM kNTEUnits fix

meisonlikesicecream:
A recent change in `kNTEUnits` made the VMRCM crash for layer 1. This is a quick fix for that. Instead of using the new `kNTEUnits` array that is indexed by seed type, I added a similar array `kNTEUnitsLayerDisk` that is indexed by the layer/disk number instead. Now works for all barrel layers again. As before, it runs for the disks, but with inconsistent stubs as the stubs in the testvectors are processed in a different order than what is done in the HLS, i.e. not due to `kNTEUnits`. (I have reminded Anders about this fix in the emulation)

Added `generate_VMRCM.py` to the download script, thus could remove the VMRCM top functions from the repo.

Also changed `edm::LogWarning` to `edm::LogVerbatim` for the memory templates as this was requested by people working on the "new" emulation.


pwittich:
Hi @meisonlikesicecream thanks for making this fix; when you say 'quick fix' does that mean you think it should be fixed differently in the long term? I don't really understand the issue well enough to know if it's that or if you're just saying that the fix was easy and quick to make.

jasonfan393:
> Hi @meisonlikesicecream thanks for making this fix; when you say 'quick fix' does that mean you think it should be fixed differently in the long term? I don't really understand the issue well enough to know if it's that or if you're just saying that the fix was easy and quick to make.

I think she's just saying it was a pretty easy fix to make, the issue at hand is mainly me making a pretty silly calculation mistake when changing this kNTEUnits array when updating TP.

meisonlikesicecream:
> Hi @meisonlikesicecream thanks for making this fix; when you say 'quick fix' does that mean you think it should be fixed differently in the long term? I don't really understand the issue well enough to know if it's that or if you're just saying that the fix was easy and quick to make.

I guess it's a bit of both... I think it's a bit ugly to have two `kNTEUnits` arrays with the same numbers but in different orders due the original one being sorted by seed type (which is incompatible with the VMRCM) and the one I added is in the order of layer/disk number (incompatible with the TP, I think). However, I couldn't immediately come up with an easy way of converting "seed type" to layer/disk number. So if the constants in kNTEUnits change, we need to change the numbers in two places instead of one. It's not the end of the world, but it would be a tiny bit nicer to have a single `kNTEUnits`, but haven't had time to think about how one would be able to do that without making either the TP or the VMRCM sad. Suggestions are welcome :-)

PR #207, opened 2021-10-13
MC X-Macros

bryates:
This PR updates the MC test bench to work with X-Macros. This fully works for L3 - L6 (simulation, synthesis, and implantation). L1 and L2 currently fail simulation, most likely due to the test vectors including seeding in the disk layers. Once I finish implementing the disks in the MC, all modules should work.

The simulation does give a warning about the timing bine `3.5055ns`, which is above the `3.5ns` constraint. However, the amount over is small, and the simulation is usually an overestimate.

### DIFF
--- a/TrackletAlgorithm/MatchCalculator.h
+++ b/TrackletAlgorithm/MatchCalculator.h
@@ -291,6 +291,7 @@ void MatchCalculator(BXType bx,
 #pragma HLS inline
 #pragma HLS array_partition variable=match complete dim=1
 #pragma HLS array_partition variable=fullmatch complete dim=1
+  constexpr int totalMatchCopies(8);
 
   // Initialization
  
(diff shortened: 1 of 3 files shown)


TrackletAlgorithm/MatchCalculator.h
```diff
@@ -334,26 +335,21 @@ void MatchCalculator(BXType bx,
...
-  ap_uint<kNBits_MemAddr> addr8 = 0; 
+  ap_uint<kNBits_MemAddr> addr[totalMatchCopies];
+#pragma HLS array_partition variable=addr complete dim=0
+  addr_loop: for(int i = 0; i < totalMatchCopies; ++i) { // priority encoder ALWAYS expects 8
+#pragma HLS unroll
+    addr[i] = 0;
+  }
 
   // Read signals for the input candidate matches
   bool read1 = false;
```
aehart, line 355: @bryates It looks like `read1`, `read2`, …, and `read1_next`, `read2_next`, … are no longer needed. Can they be removed?

bryates: Good point. I've removed them.

aehart (approved):
Looks good. Approved!

PR #209, opened 2021-10-18
VMRCM Disk Debug

meisonlikesicecream:
TL;DR: fixed some bugs for the VMRCM in disk as that I hadn't noticed before.

Up until now, I hadn't been able to properly check the output of the VMRCM in the disks as the testvectors contained the stubs in a different order from what the HLS VMRCM was processing them in. So even though the VMRCM disk was running fine, it was not possible to compare them to the testvectors. Managed to get some new testvectors from Anders where they were in the correct order, and could therefore finally check the output from VMRCM in the disks. Noticed a few bugs related to some bit-widths that I have fixed in this PR.

I have not updated the testvectors in this PR, I guess that'll happen sometime in the future, so if someone wants to run it themselves (without failing the simulation as it currently does, and will continue to do until we have updated the testvectors) they'll have to use the following in `download.sh`:
```
# Combined modules
memprints_url_cm="https://aryd.web.cern.ch/aryd/MemPrints_Combined_211012.tgz"
luts_url_cm="https://aryd.web.cern.ch/aryd/LUTs_Combined_211012.tgz"
```

### DIFF
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -38,7 +38,8 @@ class MemoryTemplateBinnedCM{
   enum BitWidths {
     kNBxBins = 1<<NBIT_BX,
     kNSlots = 1<<NBIT_BIN,
-    kNMemDepth = 1<<NBIT_ADDR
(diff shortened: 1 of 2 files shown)


aehart (approved):
Looks good. Approved!

PR #212, opened 2021-10-25
TE/TC updates

aehart:
This PR includes a few updates to the TE and TC so that a complete barrel-only configuration is possible:
 - TE top functions for PS+2S (L3L4) and 2S+2S (L5L6) have been added.
 - The `ASInnerMask` and `ASOuterMask` of the TC have been expanded so that each SP memory is represented by two bits. This is necessary because some TCs take in more than two AS memories, so at least two bits are needed to indicate which should be used with a given SP memory.
 - Support for L2L3 has been added to the TC following the example of the TP.
 - The "phicrit" cut in the TC has been updated to match the emulation, as well as the HLS for the TP.

At least one TE for each barrel-only seed has been successfully tested.

All of the barrel-only TCs have also been successfully tested, with the exception of three, for which there are some inconsistencies in the disk projections. These should be straightforward to debug in the near future, but should not be necessary for a full barrel-only configuration.


tomalin (approved):
Looks very good. Thanks.

PR #213, opened 2021-10-26
MP X-Macros

bryates:
This is the MP equivalent of PR #207. It fully compiles and runs for the MP in `L3PHIC` using the following test vectors
```bash
memprints_url_cm="https://aryd.web.cern.ch/aryd/MemPrints_Combined_211012.tgz"
luts_url_cm="https://aryd.web.cern.ch/aryd/LUTs_Combined_211012.tgz"
```
without these, there are 5 events where the HLS has an extra `fullmatch` output.

It also adds generate_MP.py, which is needed by the CMSSW Future Emulation code.


TrackletAlgorithm/MatchEngineUnit_parameters.h
```diff
@@ -0,0 +1,118 @@
+#ifndef TrackAlgorithm_MatchEngineUnit_parameters_h
```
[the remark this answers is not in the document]
aehart: @bryates Unless I've missed something, this file doesn't depend on the wiring at all. If that's true, then you can remove it from generate_MP.py. The Python scripts only need to generate files that have some dependence on the wiring.

TrackletAlgorithm/MatchProcessor_parameters.h
```diff
@@ -1,115 +1,170 @@
 #ifndef TrackletAlgorithm_MatchProcessor_parameters_h
```
aehart, line 1: @carriganm95 I'd like to avoid having duplicate code if possible. Is there any way to adapt the workflow for the emulation so that you use the version of this file produced by generate_MP.py?

TrackletAlgorithm/MatchEngineUnit_parameters.h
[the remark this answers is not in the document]
bryates: That's a good point, I've removed it.

TrackletAlgorithm/MatchProcessor_parameters.h
carriganm95: yes we can do that

TestBenches/Macros.h
```diff
@@ -1176,7 +1176,55 @@ X(TP_L3L4D_, "TP_L3L4D") \
 X(TP_L5L6A_, "TP_L5L6A") \
 X(TP_L5L6B_, "TP_L5L6B") \
 X(TP_L5L6C_, "TP_L5L6C") \
-X(TP_L5L6D_, "TP_L5L6D") 
+X(TP_L5L6D_, "TP_L5L6D") \
```
pwittich, line 1179: FWIW I think the xmacros could be simplified to just be X(TP_L5L6D) and then generate the underscore and the text substitution directly. This is more universally true for all X-macros. @bryates @aperloff

tomalin:
The TC HLS module fails the CSIM CI check here, but passes it in other PRs. Does this PR need rebasing?

bryates:
> The TC HLS module fails the CSIM CI check here, but passes it in other PRs. Does this PR need rebasing?

I just tired to rebase with the master, but that did not fix the CI issues. @aryd could this be related to the new test vectors we're using?

tomalin:
> > The TC HLS module fails the CSIM CI check here, but passes it in other PRs. Does this PR need rebasing?
> 
> I just tired to rebase with the master, but that did not fix the CI issues. @aryd could this be related to the new test vectors we're using?

Given that PR https://github.com/cms-L1TK/firmware-hls/pull/219 passed the CI tests yesterday, and is apparently using the same download.sh as your PR, I'm dubious.

bryates:
> > > The TC HLS module fails the CSIM CI check here, but passes it in other PRs. Does this PR need rebasing?
> > 
> > 
> > I just tired to rebase with the master, but that did not fix the CI issues. @aryd could this be related to the new test vectors we're using?
> 
> Given that PR #219 passed the CI tests yesterday, and is apparently using the same download.sh as your PR, I'm dubious.

This PR is using 
https://github.com/cms-L1TK/firmware-hls/blob/bd72f8787be43ac579e62b54b58d38a207cae6cf/emData/download.sh#L12-L18
while #219 is using
https://github.com/cms-L1TK/firmware-hls/blob/066f4fc4e1119a4ba1fea3ed0633fffaef056f15/emData/download.sh#L12-L18

aehart:
> > > > The TC HLS module fails the CSIM CI check here, but passes it in other PRs. Does this PR need rebasing?
> > > 
> > > 
> > > I just tired to rebase with the master, but that did not fix the CI issues. @aryd could this be related to the new test vectors we're using?
> > 
> > 
> > Given that PR #219 passed the CI tests yesterday, and is apparently using the same download.sh as your PR, I'm dubious.
> 
> This PR is using
> 
> https://github.com/cms-L1TK/firmware-hls/blob/bd72f8787be43ac579e62b54b58d38a207cae6cf/emData/download.sh#L12-L18
> 
> while #219 is using
> 
> https://github.com/cms-L1TK/firmware-hls/blob/066f4fc4e1119a4ba1fea3ed0633fffaef056f15/emData/download.sh#L12-L18

I'm not sure about the origin of the new test vectors, but in any case, I think we want to leave the URLs in download.sh alone in this PR. The test vectors are properly updated in #220.

bryates:
@aehart I've commented out the new test vectors, and set the CI to allow the MP to fail again. This PR should be good to go.

PR #214, opened 2021-11-09
Chains for reduced config

aehart:
This PR is a essentially a rebase and restructure of the contents of `reduced_config_pr`. This includes two new subchains for the reduced config, MCTB and IRtoTB, as well as fixes to the IR and VMR.

Other minor changes include:
- A new Tcl command called `build_ip` has been added in IntegrationTests/common/script/build_ip.tcl that builds a single HLS IP given a top function name. All the compileHLS.tcl scripts now use it, which greatly simplifies them.
- All the compileHLS.tcl scripts now also create a symbolic link to IntegrationTests/common in the current working directory. This simplifies some paths, in particular, in the CI tests.
- Initialization of `data_` in all the default constructors for the memory classes has been removed. This helps avoid unintentional loops that HLS creates when an array of these objects is declared. If initialization really is needed, it should be done explicitly.
- A TPAR offset parameter has been added to the TrackBuilder. This is needed because the single input TPAR memory in the reduced config would normally be the sixth TPAR memory in the full config, so an offset is needed in order to index the memory correctly.

@meisonlikesicecream @sseifeln Could you take a look at your respective modules and make sure nothing got mistranslated during rebase?

Edit: All the compileHLS.tcl scripts now create a second symbolic link to the appropriate MemPrints directory (for either the full or the reduced config) in the current working directory. This again simplifies some paths, in particular, for the comparison script.

Edit: A symlink to the appropriate LUTs directory is also created now for the same reason.


TestBenches/InputRouter_test.cpp
```diff
@@ -546,75 +671,180 @@ int main(int argc, char * argv[])
...
+      case 16: 
+        InputRouterTop_IR_DTC_2S_4_B( hBx
+        , cLnkWrd // input link LUT 
+        , cPhBnWord  // n phi bins LUT 
+        , hInputStubs // input stub stream 
+        , hBx_o // output bx 
+        , hMemories); 
+        break;
+    } 
+
```
aehart, line 814: @sseifeln The IR currently fails C/RTL cosimulation because these lines cause a different top function to be called than what is specified in project/script_IR.tcl. Could you please fix?

meisonlikesicecream:
If you look at the CI pipeline overview, it looks like the IRVMR chain succeeds, but in reality it actually fails: https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/17453120 and https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/17453125. 

It should be an easy fix, just update the IRVMR VHDL files with new ones from the project generation scripts (I'll do that once I've checked that everything's OK), but shouldn't the CI catch this crash in the first place?

meisonlikesicecream:
I've checked the VMR and it looks OK. The IRVMR chain is fine as well now; we have the same number of errors as the current master branch.

aehart:
> If you look at the CI pipeline overview, it looks like the IRVMR chain succeeds, but in reality it actually fails: https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/17453120 and https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/17453125.
> 
> It should be an easy fix, just update the IRVMR VHDL files with new ones from the project generation scripts (I'll do that once I've checked that everything's OK), but shouldn't the CI catch this crash in the first place?

Yes, good catch. Thanks for updating the VHDL files and adding that check to the comparison script.

TestBenches/InputRouter_test.cpp
sseifeln: hi @aehart  - sure (sorry for the late response) - which top function should be called instead? I remember it was one of the PS10G IRs but not which one specifically..

tomalin:
The useful info that was in https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_pr/IntegrationTests/ReducedConfigPR/hdl/README.md and https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_pr/IntegrationTests/ReducedConfigPR/script/README.md has been lost in this PR.  (Although the former is perhaps not needed, as https://github.com/cms-L1TK/project_generation_scripts/blob/master/README.md gives a compatible recipe).

TestBenches/InputRouter_test.cpp
aehart: The default top function in project/script_IR.tcl is InputRouterTop_IR_DTC_PS10G_3_A, and that script passes a link ID of 8 to the test bench. That's unchanged from the master branch, but this switch statement, which came from the reduced_config_pr branch, seems to be inconsistent with those settings.

aehart:
> The useful info that was in https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_pr/IntegrationTests/ReducedConfigPR/hdl/README.md and https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_pr/IntegrationTests/ReducedConfigPR/script/README.md has been lost in this PR. (Although the former is perhaps not needed, as https://github.com/cms-L1TK/project_generation_scripts/blob/master/README.md gives a compatible recipe).

I didn't include the former because the project_generation_scripts repo feels like the place for such a recipe and yours is cleaner anyway. I didn't include the latter because that information is mostly already in the top level README:
https://github.com/cms-L1TK/firmware-hls/blob/master/README.md#running-chains-illustrated-for-pr-me-mc

But I have no problem adding back either README. Or I could add extra information to the top level README. Whatever you prefer.

IntegrationTests/ReducedConfig/IRtoTB/script/compileHLS.tcl
```diff
@@ -0,0 +1,98 @@
...
+add_files ../../../../TopFunctions/ReducedConfig/ProjectionRouterTop.cc -cflags "$CFLAGS"
+build_ip ProjectionRouterTop_L3PHIB ../../../../project
+build_ip ProjectionRouterTop_L4PHIB ../../../../project
+build_ip ProjectionRouterTop_L5PHIB ../../../../project
+build_ip ProjectionRouterTop_L6PHIB ../../../../project
+
+# Build MEs
+open_project -reset matchengine
+add_files ../../../../TopFunctions/ReducedConfig/MatchEngineTop.cc -cflags "$CFLAGS"
+build_ip MatchEngineTop_L3 ../../../../project
```
tomalin, line 80: This is a typo. MatchEngineTop_L3 has been repeated four times, when I suspect that L3, L4, L5 & L6 should be used?

aehart: Exactly right. Fixed.

IntegrationTests/IRVMR/script/compileHLS.tcl
```diff
@@ -1,28 +1,19 @@
 # Common settings
 source ../../../project/env_hls.tcl
+file link -symbolic common ../../common
+file link -symbolic MemPrints ../../../emData/MemPrints
+source common/script/build_ip.tcl
 set CFLAGS {-std=c++11 -I../../../TrackletAlgorithm -I../../../TopFunctions}
 set cwd [pwd]
 
 # Build IR_DTC_PS10G_3_A
 open_project -reset inputrouter
```
tomalin, line 10: It would be nice to provide option (normally disabled to save CPU) to run also csim & cosim, to check that the HLS module gives correct results. And to throw an error if they reveal problems ...

meisonlikesicecream:
I've changed a line in the gitlab yaml file so it does an additional check for the string "error" in the vivado output when we run runSim.tcl: https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_for_pr/.gitlab-ci.yml#L70

Do we need this extra safety check in other places where we call Vivado as well?

tomalin:
runSim.tcl fails complaining that the file ../../../../../../../../emData/LUTs/TE_L1PHID14_L2PHIB15_stubptinnercut.tab can't be opened. This is caused by this line https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_for_pr/IntegrationTests/ReducedConfig/IRtoTB/hdl/SectorProcessor.vhd#L1387 . It uses function getDirEMDATA() in tf_pkg.vhd to determine the correct relative path to this file to use with SIM or SYNTH. But getDirEMDATA() assumes the directory structure is similar to that in IntegrationTests/PRMEMC/ , so is confused by the extra directory layer in IntegrationTests/ReducedConfig/IRtoTB/ .

aehart:
> I've changed a line in the gitlab yaml file so it does an additional check for the string "error" in the vivado output when we run runSim.tcl: https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_for_pr/.gitlab-ci.yml#L70
> 
> Do we need this extra safety check in other places where we call Vivado as well?

Thanks, @meisonlikesicecream. I think we want this whenever vivado is called. So here
https://github.com/cms-L1TK/firmware-hls/blob/7c430ec1e2ddccca17f2a6960a6b8753526b299a/.gitlab-ci.yml#L69
and here as well
https://github.com/cms-L1TK/firmware-hls/blob/7c430ec1e2ddccca17f2a6960a6b8753526b299a/.gitlab-ci.yml#L101

However, I wonder if a simpler solution would be to just change `-mode tcl` → `-mode batch`. This seems to do the right think when I run it interactively (i.e., the exit code is nonzero when there is an error), but I haven't tested it in the CI. Did you try this in your testing?

aehart:
> runSim.tcl fails complaining that the file ../../../../../../../../emData/LUTs/TE_L1PHID14_L2PHIB15_stubptinnercut.tab can't be opened. This is caused by this line https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_for_pr/IntegrationTests/ReducedConfig/IRtoTB/hdl/SectorProcessor.vhd#L1387 . It uses function getDirEMDATA() in tf_pkg.vhd to determine the correct relative path to this file to use with SIM or SYNTH. But getDirEMDATA() assumes the directory structure is similar to that in IntegrationTests/PRMEMC/ , so is confused by the extra directory layer in IntegrationTests/ReducedConfig/IRtoTB/ .

I now have compileHLS.tcl create a symlink to the appropriate LUTs directory so that the SectorProcessor does not need to know where emData/LUTs/ is.

IntegrationTests/IRVMR/script/compileHLS.tcl
aehart: This would require a little more work. In particular, we need a way of switching the top function that is called during cosimulation. Currently, this is hard-coded in most of the C++ test benches. Also, all the test benches are configured to use the top functions generated from the full configuration, and we need to figure out how to switch to those generated from the reduced configuration from inside the Tcl script.

I would propose addressing this in a followup PR after this one is merged, because I agree that it would be very nice if this were possible.

meisonlikesicecream:
> > I've changed a line in the gitlab yaml file so it does an additional check for the string "error" in the vivado output when we run runSim.tcl: https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_for_pr/.gitlab-ci.yml#L70
> > Do we need this extra safety check in other places where we call Vivado as well?
> 
> Thanks, @meisonlikesicecream. I think we want this whenever vivado is called. So here
> 
> https://github.com/cms-L1TK/firmware-hls/blob/7c430ec1e2ddccca17f2a6960a6b8753526b299a/.gitlab-ci.yml#L69
> 
> 
> and here as well
> https://github.com/cms-L1TK/firmware-hls/blob/7c430ec1e2ddccca17f2a6960a6b8753526b299a/.gitlab-ci.yml#L101
> 
> However, I wonder if a simpler solution would be to just change `-mode tcl` → `-mode batch`. This seems to do the right thing when I run it interactively (i.e., the exit code is nonzero when there is an error), but I haven't tested it in the CI. Did you try this in your testing?

Woaw, it did work changing mode from tcl to batch,. I've pushed those changes now, without doing the awk magic thing.

aehart:
> > > I've changed a line in the gitlab yaml file so it does an additional check for the string "error" in the vivado output when we run runSim.tcl: https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_for_pr/.gitlab-ci.yml#L70
> > > Do we need this extra safety check in other places where we call Vivado as well?
> > 
> > 
> > Thanks, @meisonlikesicecream. I think we want this whenever vivado is called. So here
> > https://github.com/cms-L1TK/firmware-hls/blob/7c430ec1e2ddccca17f2a6960a6b8753526b299a/.gitlab-ci.yml#L69
> > 
> > and here as well
> > https://github.com/cms-L1TK/firmware-hls/blob/7c430ec1e2ddccca17f2a6960a6b8753526b299a/.gitlab-ci.yml#L101
> > 
> > However, I wonder if a simpler solution would be to just change `-mode tcl` → `-mode batch`. This seems to do the right thing when I run it interactively (i.e., the exit code is nonzero when there is an error), but I haven't tested it in the CI. Did you try this in your testing?
> 
> Woaw, it did work changing mode from tcl to batch,. I've pushed those changes now, without doing the awk magic thing.

As a Linux die-hard, it feels wrong encouraging someone to not use awk, but I'm glad the simpler solution works. Thanks for testing it.

IntegrationTests/ReducedConfig/IRtoTB/script/compileHLS.tcl
```diff
@@ -0,0 +1,100 @@
+# Common settings
+source ../../../../project/env_hls.tcl
+file link -symbolic common ../../../common
```
meisonlikesicecream, line 3: If I want to recompile the HLS, the tcl script complains about not being able to create new links until I've manually removed them. Could this be automised so it's fool-proof? (I'm lazy and stupid; managed several times to delete the actual files and not just the link...)

```Sourcing Tcl script 'compileHLS.tcl'
could not create new link "common": that path already exists
    while executing
"file link -symbolic common ../../../common"
    (file "compileHLS.tcl" line 3)
    invoked from within
"source compileHLS.tcl"
    ("uplevel" body line 1)
    invoked from within
"uplevel \#0 [list source $arg] "
```

meisonlikesicecream (commented):
@aehart In the previous branch, reduced_config_pr, the second template parameter `NBIT_MEMADDR` in had been removed as it seemed redundant: https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_pr/TrackletAlgorithm/VMStubMEMemory.h#L156

Is there a reason why this change isn't in this PR?
(I tried those changes for one VMR module in each layer/disk and it seemed to work. Can't remember why we had that extra parameter that in the first place...)

IntegrationTests/common/hdl/tf_pkg.vhd
```diff
@@ -171,17 +171,17 @@ package body tf_pkg is
...
     if IS_SIMULATION then
       -- Sim path specified relative to Vivado project's xsim directory. 
       -- e.g. IntegrationTests/PRMEMC/script/Work/Work.sim/sim_1/behav/xsim/
-      return "../../../../../../../../emData/";
+      return "../../../../../";
     else
       -- Synth path specified relative to dir where you run Vivado.
       -- e.g. IntegrationTests/PRMEMC/script/
-      return "../../../emData/";
+      return "./";
```
meisonlikesicecream, line 184: Should this one mayhaps be ```return "../../../";```? Or topTETC-synth will complain.

aehart:
> @aehart In the previous branch, reduced_config_pr, the second template parameter `NBIT_MEMADDR` in had been removed as it seemed redundant: https://github.com/cms-L1TK/firmware-hls/blob/reduced_config_pr/TrackletAlgorithm/VMStubMEMemory.h#L156
> 
> Is there a reason why this change isn't in this PR? (I tried those changes for one VMR module in each layer/disk and it seemed to work. Can't remember why we had that extra parameter that in the first place...)

It looks like that parameter was added back in PR #202, which was merged into the master branch back in September. Apparently, reduced_config_pr was rebased before then, so it doesn't include that change.

Maybe @aperloff can comment on why this extra parameter is needed.

TestBenches/InputRouter_test.cpp
sseifeln: ok thanks .. sorry we have a beam test on-going so didn't get around to it yet. But will make sure to commit a change tomorrow.

IntegrationTests/ReducedConfig/IRtoTB/script/compileHLS.tcl
aehart: I also wanted this to be able to create broken symlinks, which is useful for somebody running this interactively before emData/download.sh has been run. So I replaced "file link" with a function that just executes the "ln" shell command. This takes care of both issues.

IntegrationTests/common/hdl/tf_pkg.vhd
aehart: I don't really understand relative paths in Vivado, but yeah `../../../` seems to work. ¯\\\_(ツ)\_/¯

meisonlikesicecream: Looks like the CI didn't like it anyway :-(

aehart: OK I really don't understand what's happening. This worked interactively.

I'll have to try to debug this later…

TestBenches/InputRouter_test.cpp
sseifeln: @aehart  - just committed the changes and it now passes cosim : https://github.com/cms-L1TK/firmware-hls/commit/9c6438d2a5177ebcf931f59ab94ad2b9a386bb8b

aehart: Thank you!

IntegrationTests/common/hdl/tf_pkg.vhd
aehart: OK, I just had to add the download artifacts in the CI so that the link to the LUTs directory actually points to something. Everything works now!

aperloff:
> Maybe @aperloff can comment on why this extra parameter is needed.

Because this parameter of MemoryTemplateBinned needs to be configurable as per:
https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MatchEngine.h#L120
and
https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MatchEngine.h#L42-L46
This was needed to get the ME disks to work.

meisonlikesicecream (approved):
Looks fine to me, but maybe someone else should approve it as well?

aehart:
@pwittich and @tomalin, before we merge, are you happy with this PR?

tomalin:
> @pwittich and @tomalin, before we merge, are you happy with this PR?

Providing that:

1) linkCreate in compileHLS.sh allows this script to be run repeatedly without crashing due to already created symbolic links. (I assume it does, but haven't checked).
2) It is checked that it makes sense for linkCreate here to link to emData/LUTs/ instead of emData/LUTsReduced/ .
3) It is explained why we need a symbolic link to common/ rather than just running the scripts in the original location?

tomalin:
Could we change the default download.sh to download 100 events instead of 1 for reduced chain? Or at least use 100 events for the CI test? (I've checked its OK for 96 events out of 100 for IR to TB chain).

tomalin:
I've just made a PR to this branch https://github.com/cms-L1TK/firmware-hls/pull/216 to fix the issue with the START signal in HW reported by Ben at last HLS meeting. The corresponding script PR is https://github.com/cms-L1TK/project_generation_scripts/pull/40 .

aehart:
> > @pwittich and @tomalin, before we merge, are you happy with this PR?
> 
> Providing that:
> 
>     1. It is checked that it makes sense for linkCreate in compileHLS.tcl to link to emData/LUTs/ instead of emData/LUTsReduced/ . (The LUT tables are the same in both, so perhaps doesn't matter).
You're right that it doesn't matter, so I changed the link to point to LUTsReduced to avoid confusion.
> 
>     2. It is explained why we need a symbolic link to common/ rather than just running the scripts in the original location?
This is a matter of convenience. The check-results CI jobs call common/script/CompareMemPrintsFW.py:
https://github.com/cms-L1TK/firmware-hls/blob/b8339d0c9dee62de3c3958d4ddf6b03f8c30ddf5/.gitlab-ci.yml#L79-L91
In order for this to work from subdirectories of different depths, I created a symlink. The other option would be to add a variable to the CI that points to the location of the common/ directory, so we can call the script as
```yaml
    - python3 ${COMMON_DIR}/script/CompareMemPrintsFW.py -p -s
```
I have no opinion about which we do, and I can switch to the other option if you prefer.

aehart:
> Could we change the default download.sh to download 100 events instead of 1 for reduced chain? Or at least use 100 events for the CI test? (I've checked its OK for 96 events out of 100 for IR to TB chain).

I think this makes sense, so I changed the default to 100 events in download.sh.

tomalin (approved):
happy

tomalin:
@aehart does anything prevent this PR being merged today?

aehart:
> @aehart does anything prevent this PR being merged today?

Sorry for the delay. GitHub reports that the latest checks failed, but this seems to be due to a problem connecting to Gitlab. The tests actually passed:
https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/3274482
So I will merge now.

PR #216, opened 2021-11-19
Fix START signal creation to solve HW problem

tomalin:
Adds utility common/hdl/CreateStartSignal.vhd to create the START signal of an algo module from the DONE signal of the previous module in the chain.
Assumed to exist by cms-L1TK/project_generation_scripts#40 .
This should fix issues with START seen when running in HW.


aehart (approved):
This looks good to me. I'll approve and merge it.

Issue #188, opened 2021-08-06
ME reads the wrong stub in Vivado simulation


meisonlikesicecream:
The remaining errors in the PRMEMC chain is caused by the ME. It seems like the ME reads the wrong VMSME stubs if the previous event continues reading from the VMSME memories after the current event has started (thanks to rewind?). We don’t see this issue in the HLS simulation as it adds a few clock cycles in between events for some reason. 

I have attached a screenshot of the waveforms for one of the inconsistent events (event 27 in ME_L3PHIC19) to try and demonstrate this: 
The event starts where the **blue** vertical marker is positioned. 
The read address for the VMPROJ memories (red row) are correct and two clock cycles later, marked by the **yellow** vertical marker, it reads the correct VMPROJ stub (magenta row). 
In that same clock cycle (yellow vertical marker) the ME then uses this VMPROJ stub to figure out the read address for the VMSME stub (yellow rows; I have split it up into BX, BIN, STUB NUMBER = which stub we read within a bin). This address is incorrect. The BX and BIN bits are correct, but it should start with the 0th stub and continue up to the number of entries within that bin. In the events when it fails, STUB NUMBER is non-zero (7 in this case) and we end up reading the wrong ME stub. Instead of resetting the STUB NUMBER to 0 when reading from a new bin, it increments the stub number from the previous event as it had more than 7 stubs in that bin. 
This then leads to the wrong VMSME stub being read, which in turn leads to a missing stub in the output memory.

![image001](https://user-images.githubusercontent.com/56160327/128517342-067a7a9f-e131-4c0d-afca-68bc4e3689f2.png)

There is an ugly bodge to avoid this issue by not allowing the ME to read VMSME stubs after clock cycle 106 ( i.e. add `if (bufferNotEmpty && istep<kMaxProc-2) {` to this line: https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MatchEngine.cc#L140), but then it would of course truncate some stubs further down the chain unless we change the testvectors. Don't know if that bodge only works for our specific PRMEMC chain or in general.

I have added the ME VHDL testbench in my repo and it can be run using the same instructions as for the PRMEMC in the README for debugging purposes: https://github.com/meisonlikesicecream/firmware-hls/tree/me_vhdl_tb/IntegrationTests/ME

aehart:
Thanks for revisiting this @meisonlikesicecream. I at least had lost track of this issue.

[...]

PR #219, opened 2021-12-13
Reduce some VHDL top-level simulation errors

meisonlikesicecream:
Addresses some of the problems with resetting read/write addresses when running modules with the VHDL top-level in Vivado simulation (i.e. not Vivado HLS) mentioned in issue https://github.com/cms-L1TK/firmware-hls/issues/188. 

- Reduces errors in PRMEMC chain from 3 to 1. 

- Reduces errors in e.g. VMR_L2PHIC (not tested by default in CI). The remaining errors are due to missing stubs due to sacrificing the first loop iteration to clear the write address counters. Will do a PR to the emulation to include this change in the number of stubs the VMR can process.

- Updated SectorProcessor*.vhd & tb_tf_top.vhd using latest version of scripts for IR-TB & PR-ME-MC chains

### DIFF
--- a/TrackletAlgorithm/MatchEngine.h
+++ b/TrackletAlgorithm/MatchEngine.h
@@ -301,6 +301,9 @@ void MatchEngine(const BXType bx, BXType& bx_o,
 			  istub = 0;
 			  tail_readindex++;
 			}
+			else if(istep == kMaxProc - kMaxProcOffset(module::ME) - 1){ // The last loop iteration
+				istub = 0; // Reset stub index for next bx, needed for VHDL top-level Vivado simulation
+			}
 			else {
 			  istub++;
 			}
--- a/TrackletAlgorithm/MatchEngine.h
+++ b/TrackletAlgorithm/MatchEngine.h
@@ -301,8 +301,8 @@ void MatchEngine(const BXType bx, BXType& bx_o,
 			  istub = 0;
 			  tail_readindex++;
 			}
-			else if(istep == kMaxProc - kMaxProcOffset(module::ME) - 1){ // The last loop iteration
-				istub = 0; // Reset stub index for next bx, needed for VHDL top-level Vivado simulation
+			else if (istep == kMaxProc - kMaxProcOffset(module::ME) - 1) { // The last loop iteration
+				istub = 0; // Reset stub index for next bx. Needed for VHDL top-level Vivado simulation
 			}
 			else {
 			  istub++;
--- a/TrackletAlgorithm/VMRouter.h
+++ b/TrackletAlgorithm/VMRouter.h
@@ -657,26 +657,32 @@ void VMRouter(const BXType bx, BXType& bx_o, const int fineBinTable[], const int
 	ap_uint<kNBits_MemAddr> addrCountOL[nvmOL][MaxOLCopies]; // Writing of TE Overlap stubs
 	ap_uint<kNBits_MemAddr> addrCountTEO[nvmTE][MaxTEOCopies][nmaxbinsperpage]; // Writing of TE Outer stubs
 
-	if (maskME) {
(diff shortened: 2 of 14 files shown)


meisonlikesicecream:
Updated the VHDL top-level files as they were a bit outdated. Have not updated the MCTB chain as it seemed to have some manual edits?

TrackletAlgorithm/VMRouter.h
```diff
@@ -657,27 +657,33 @@ void VMRouter(const BXType bx, BXType& bx_o, const int fineBinTable[], const int
 	ap_uint<kNBits_MemAddr> addrCountOL[nvmOL][MaxOLCopies]; // Writing of TE Overlap stubs
 	ap_uint<kNBits_MemAddr> addrCountTEO[nvmTE][MaxTEOCopies][nmaxbinsperpage]; // Writing of TE Outer stubs
 
-	if (maskME) {
-		clear2DArray(nvmME, addrCountME);
```
pwittich, line 661: can you remind me what these arrays are? It seems like clearing a 2d array or a 3d array would take a long time (many clock cycles). I didn't think we needed to clear these arrays on every new BX.

TrackletAlgorithm/MatchEngine.h
```diff
@@ -301,6 +301,9 @@ void MatchEngine(const BXType bx, BXType& bx_o,
 			  istub = 0;
 			  tail_readindex++;
 			}
+			else if (istep == kMaxProc - kMaxProcOffset(module::ME) - 1) { // The last loop iteration
```
pwittich, line 304: is this logically equivalent to the clearing of the multi-D arrays in the VMRouter?

pwittich (dismissed):
Please see comments inline; I am concerned that adding two- or three-deep loops is pretty costly on every loop iteration?

TrackletAlgorithm/VMRouter.h
meisonlikesicecream: They keep track of the memory address the output stubs should be written to. Because we've got several possible memories to write to, and they have copies, and some of them have bins that also needs to be kept track of, they're multi-dimensional. They are partitioned fully so seems like it can handle it? It's just once every BX. 

If we don't clear it at the start of every BX, the stubs would get written to the wrong address, and not overwrite the stubs from the previous BX. As far as I understand, we still need to clear it every new BX inside the processing module. It was within the memories that we didn't need to clear it?

Before I cleared it before/outside the main loop, but sometimes it didn't clear it properly when looking at the Vivado simulation. Moving it inside the loop seemed to ensure that the arrays got cleared. Don't know if it has anything to do with the rewind option, as the documentation says:

_The code segment before the loop:_
- _Is considered as initialization._
- _Is executed only once in the pipeline._
- _Cannot contain any conditional operations (if-else)._

TrackletAlgorithm/MatchEngine.h
meisonlikesicecream: yes-ish... but here it keeps track of from which memory address the ME should READ from. Since it only reads from one memory at a time, it's not a multi-dimensional array.

The reason I clear it in the last loop-iteration (in contrast to the VMR where I do it in the first) is just pure trial and error... I don't understand what HLS seems to understand and not. To me it doesn't make much sense.

tomalin:
@aperloff since this fixes bugs in the MatchEngine, you might want to look at the changes too.

aperloff:
Hi @tomalin, I have taken a look and this looks fine to me from a purely technical standpoint. If it fixes the bugs reported in #188 then I'm all for it.

aehart (approved):
Looks good to me. Approved!

aehart:
@pwittich Could you please mark your comments as "resolved" if you're happy with them?

PR #221, opened 2022-01-14
Added constants needed by future SW to Constants.h

carriganm95:
Added three constants to TrackletAlgorithm/Constants.h

constexpr int nRegionsL1 = 8; number of regions in L1
constexpr int nRegions = 4; number of regions in other layers
constexpr int nBarrelPS = 3; number of barrel PS layers

These constants are needed to avoid hard coding template arguments in both the firmware_hls and the future SW

### DIFF
--- a/TrackletAlgorithm/Constants.h
+++ b/TrackletAlgorithm/Constants.h
@@ -156,4 +156,9 @@ typedef ap_uint<kNBits_BX> BXType;  // temporary definition. need to be revisite
 constexpr unsigned kBRAMwidth = 36; 
 constexpr int kNBits_DTC = 39; 
 
+//Future SW constants
+constexpr int nRegionsL1 = 8;
+constexpr int nRegions = 4;
+constexpr int nBarrelPS = 3;
+
 #endif
--- a/TrackletAlgorithm/Constants.h
+++ b/TrackletAlgorithm/Constants.h
@@ -157,8 +157,6 @@ constexpr unsigned kBRAMwidth = 36;
 constexpr int kNBits_DTC = 39; 
 
 //Future SW constants
-constexpr int nRegionsL1 = 8;
-constexpr int nRegions = 4;
 constexpr int nBarrelPS = 3;
 
 #endif


TrackletAlgorithm/Constants.h
```diff
@@ -156,4 +156,9 @@ typedef ap_uint<kNBits_BX> BXType;  // temporary definition. need to be revisite
 constexpr unsigned kBRAMwidth = 36; 
 constexpr int kNBits_DTC = 39; 
 
+//Future SW constants
+constexpr int nRegionsL1 = 8;
+constexpr int nRegions = 4;
```
aehart, line 161: Are these constants redundant with the information in `nbitsallstubs`?:
https://github.com/cms-L1TK/firmware-hls/blob/c2a7e7b633526d66db12d0349f58839a855d2cde/TrackletAlgorithm/Constants.h#L129
If so, could you just use `1 << nbitsallstubs[LAYER_OR_DISK]` in the emulation?

tomalin: Why is nBarrelPS being defined in Constants.h if no HLS code needs this number? Or does the number 3 appearing at various places in the HLS code need replacing with this constant?

carriganm95: I had talked with Brent about adding these numbers and he thought he could use these constants. If this is no longer true we can move it to the future SW. @bryates

bryates: Currently, this is not used in the HLS code, but I thought it might be a good thing for all the modules to have in the future.

tomalin: If you can identify specific lines in the HLS modules would could use these numbers, (and ideally make this substitution), then adding them to Consnants.h makes sense. If not, then I'd vote not to add them.
I agree with @aehart suggestion .  (Though define an array nRegions[nLayers] in emulation initialized as he suggests).

aehart (approved):
I'm happy with this now. Approved!

PR #220, opened 2022-01-07
FW synch 12_0_0_pre4

aehart:
This PR updates the tarballs downloaded in emData/download.sh to those generated with the ~~[fw_synch_220106](https://github.com/cms-L1TK/cmssw/tree/fw_synch_220106)~~ [fw_synch_220119](https://github.com/cms-L1TK/cmssw/tree/fw_synch_220119) tag.

PR #222, opened 2022-01-19
Fixed MP timing

bryates:
After adding another layer of manual pipelining, the MP timing is now just 4ns
`CP achieved post-implementation:    3.945`

We'll need to add these changes to the emulation, since a very small number of events now have more truncation.


aehart:
This seems like a good change. Can you open a PR for the emulation, so that we can include this change in the tarballs in #220?

PR #226, opened 2022-02-16
Barrel-only TEs

aehart:
This PR generalizes the test bench for the TE to support all barrel-only modules. The test vectors for all 412 barrel-only modules are set up by emData/download.sh, and all have been tested and pass the C-simulation, although project/script_TE.tcl only tests 20 modules by default, in order to save time during the CI tests.


emData/download.sh
```diff
@@ -78,8 +78,418 @@ declare -a processing_modules=(
   "VMRCM_D5PHIA"
 
   # TrackletEngine
+  "TE_L1PHIA1_L2PHIA1"
```
meisonlikesicecream, line 81: Maybe not for this PR, but the download script is starting to get a bit long. Since we're starting to test more modules, maybe we should just handle all processing modules automatically and not have to specify each one explicitly. (I sure hope you didn't do this by hand :-) )

aehart: I used the magic of awk! (… and grep… and sed).

And yeah, I was thinking the same thing. I just pushed a commit that gets rid of the manual list and just creates links for all modules. It takes longer to run (~1.5 minutes in my testing), but I think we're at the point where having the test vectors for all the modules is a useful thing.

Let me know what you think.

aehart:
Apparently, not all TEs work with the current set of test vectors, but they do with those on the fw_sync_12_0_0_pre4 branch, so I will switch the target branch to that.

emData/download.sh
meisonlikesicecream: Running the download for 1.5 minutes once in awhile doesn't sound too bad. We don't do a fw synch often enough for that to be a pain in my opinion.

meisonlikesicecream:
> Apparently, not all TEs work with the current set of test vectors, but they do with those on the fw_sync_12_0_0_pre4 branch, so I will switch the target branch to that.

Do you know why they don't work?

aehart:
> > Apparently, not all TEs work with the current set of test vectors, but they do with those on the fw_sync_12_0_0_pre4 branch, so I will switch the target branch to that.
> 
> Do you know why they don't work?

I do not. I'll try to track down which PR in the emulation did the trick.

meisonlikesicecream (approved):

PR #228, opened 2022-02-18
Added debug condition to prinout of nMemories in InputRouter.h

sseifeln:
@tomalin - change made to reduce the printout from the IR as requested.


tomalin (approved):
Approved. Thanks Sarah!

PR #229, opened 2022-02-18
Fw sync 12 0 0 pre4 - modified order of DTCs in IR

sseifeln:
@aehart - as requested, modification needed in test bench to take into account new ordering of DTCs

### DIFF
--- a/TestBenches/InputRouter_test.cpp
+++ b/TestBenches/InputRouter_test.cpp
@@ -128,14 +128,14 @@ CablingMap getMap( std::string pInputWiresMap )
             auto cLyrIter = std::find( cLyrs.begin(), cLyrs.end(), cLayerId ); 
             if( cLyrIter == cLyrs.end() )
             {
-              std::cout << " found a new layer readout by " <<  cDtcName << "\n";
+              // std::cout << " found a new layer readout by " <<  cDtcName << "\n";
               cIterator->second.second.first.push_back( cLayerId );
               BnWrd cBnWrd(0);cBnWrd.range( cPhyId, cPhyId) = 1; 
               cIterator->second.second.second.push_back( cBnWrd ); 
             }
             else
             {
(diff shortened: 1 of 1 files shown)


aehart:
@sseifeln I hadn't noticed this before, but it looks like you only test one event by default:
https://github.com/cms-L1TK/firmware-hls/blob/aea57524999a633206756f16929fa5eb415d8a13/TestBenches/InputRouter_test.cpp#L545
Can you increase this so that you test the full 100 events in the test vectors? Also, can you print the event number in base-10 here:
https://github.com/cms-L1TK/firmware-hls/blob/aea57524999a633206756f16929fa5eb415d8a13/TestBenches/InputRouter_test.cpp#L666
The `compareMem*` functions switch the output to base-16, so you have to explicitly switch back to decimal.

Finally, just to be clear, have you explicitly verified that all IRs pass the C-simulation with the current set of test vectors?

aehart (approved):
> @sseifeln I hadn't noticed this before, but it looks like you only test one event by default:
> 
> https://github.com/cms-L1TK/firmware-hls/blob/aea57524999a633206756f16929fa5eb415d8a13/TestBenches/InputRouter_test.cpp#L545
> 
> Can you increase this so that you test the full 100 events in the test vectors? Also, can you print the event number in base-10 here:
> 
> https://github.com/cms-L1TK/firmware-hls/blob/aea57524999a633206756f16929fa5eb415d8a13/TestBenches/InputRouter_test.cpp#L666
> 
> The `compareMem*` functions switch the output to base-16, so you have to explicitly switch back to decimal.
> 
> Finally, just to be clear, have you explicitly verified that all IRs pass the C-simulation with the current set of test vectors?

I understand you may be unavailable at the moment, so I took the liberty of making these two small changes to the test bench myself.

Everything else looks good to me. Approved!

Issue #181, opened 2021-07-12
download.sh LUT feature request


tomalin:
When "download.sh -t" is run, please ensure that it creates the directories emData/*/tables/ with symbolic links to the LUT tables. This is because the code in TrackletAlgorithms includes the LUT tables from these directories, so the code won't compile (e.g. in CMSSW) when option "-t" is used.

tomalin:
This is solved by https://github.com/cms-L1TK/firmware-hls/pull/230 .

PR #230, opened 2022-02-24
Modification to download.sh for Future emulation code

tomalin:
To help Future emulation code, the download.sh scripts "-t" option now creates and fills also the directories VMR/tables/ etc.
This addresses https://github.com/cms-L1TK/firmware-hls/issues/181

### DIFF
--- a/emData/download.sh
+++ b/emData/download.sh
@@ -239,22 +239,6 @@ do
   shift
 done
 
-# If the MemPrints directory exists, assume the script has already been run,
-# and simply exit.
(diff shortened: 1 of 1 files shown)


aehart (approved):
Looks good. Approved!

aehart:
I should say, the CI tests succeeded:
https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/3659434
There seems to have been a glitch with GitHub picking up the results, but since the tests succeeded, I will merge anyway.

PR #232, opened 2022-03-15
Sector processor generation

aehart:
emData/download.sh has been updated to automatically generate the HDL top files and test benches under IntegrationTests/ so that they are always in sync with the project generation scripts.

This is done via a submodule in emData/ that points to cms-L1TK/project_generation_scripts.


meisonlikesicecream:
Looks good to me :-) 

Are the files in https://github.com/cms-L1TK/firmware-hls/tree/sector_processor_generation/IntegrationTests/PRMEMC/hdl still needed?

meisonlikesicecream:
Or actually, for confused people like me, could you either something about having to clone the repo with the `--recursive` flag, or run the `git submodule update --init` command? Otherwise the `project_generation_script` submodule will be empty. Or maybe add the latter command to the download.sh script so one doesn't have to read the README...

meisonlikesicecream:
> Or actually, for confused people like me, could you add something about having to clone the repo with the `--recursive` flag, or run the `git submodule update --init` command in the README? Otherwise the `project_generation_script` submodule will be empty. Or maybe add the latter command to the download.sh script so one doesn't have to read the README...

Nvm, just saw that you're already doing it in the download.sh. I was getting a `You need to run this command from the toplevel of the working tree.` error when running the download due to the machine I'm running on having an old git version: https://stackoverflow.com/questions/21856214/git-submodule-update-why-from-toplevel

meisonlikesicecream:
Another issue with the software versions my machine is using, so maybe not relevant for anyone else... 

Python2 is deprecated (irl and on my machine), so can't install any new packages if running python2. However it doesn't have the `builtin` package used in https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L7 so the projgen scripts crashes as the shebangs in the projgen scripts uses python2. Would it make sense if I/you/someone changed the shebangs to use python3 in the project generation scripts repo? Or do we want to keep using python2, and I'll just have to create a hack for my machine?

aehart:
> Looks good to me :-)
> 
> Are the files in https://github.com/cms-L1TK/firmware-hls/tree/sector_processor_generation/IntegrationTests/PRMEMC/hdl still needed?

I'm pretty confident they're not needed anymore. I've removed them now, so we'll see if anyone notices that they're gone.

aehart:
> Another issue with the software versions my machine is using, so maybe not relevant for anyone else...
> 
> Python2 is deprecated (irl and on my machine), so can't install any new packages if running python2. However it doesn't have the `builtin` package used in https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L7 so the projgen scripts crashes as the shebangs in the projgen scripts uses python2. Would it make sense if I/you/someone changed the shebangs to use python3 in the project generation scripts repo? Or do we want to keep using python2, and I'll just have to create a hack for my machine?

I think we've tried to maintain compatibility with both Python 2 and 3, but it probably makes sense to go all in with Python 3 at this point. I'm OK with you updating the shebangs, but I wonder if @pwittich has an opinion?

meisonlikesicecream (approved):
I've approved this now, as the shebang discussion isn't technically related to this PR :-)

pwittich:
> > Another issue with the software versions my machine is using, so maybe not relevant for anyone else...
> > Python2 is deprecated (irl and on my machine), so can't install any new packages if running python2. However it doesn't have the `builtin` package used in https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L7 so the projgen scripts crashes as the shebangs in the projgen scripts uses python2. Would it make sense if I/you/someone changed the shebangs to use python3 in the project generation scripts repo? Or do we want to keep using python2, and I'll just have to create a hack for my machine?
> 
> I think we've tried to maintain compatibility with both Python 2 and 3, but it probably makes sense to go all in with Python 3 at this point. I'm OK with you updating the shebangs, but I wonder if @pwittich has an opinion?

I agree we should let py2 die. 
FWIW, @meisonlikesicecream, you can install anaconda with user permissions and then manage your py2 installation w/o needing special privileges.

meisonlikesicecream:
> > > Another issue with the software versions my machine is using, so maybe not relevant for anyone else...
> > > Python2 is deprecated (irl and on my machine), so can't install any new packages if running python2. However it doesn't have the `builtin` package used in https://github.com/cms-L1TK/project_generation_scripts/blob/master/TrackletGraph.py#L7 so the projgen scripts crashes as the shebangs in the projgen scripts uses python2. Would it make sense if I/you/someone changed the shebangs to use python3 in the project generation scripts repo? Or do we want to keep using python2, and I'll just have to create a hack for my machine?
> > 
> > 
> > I think we've tried to maintain compatibility with both Python 2 and 3, but it probably makes sense to go all in with Python 3 at this point. I'm OK with you updating the shebangs, but I wonder if @pwittich has an opinion?
> 
> I agree we should let py2 die. FWIW, @meisonlikesicecream, you can install anaconda with user permissions and then manage your py2 installation w/o needing special privileges.

Aaah that is good to know, thanks!

I can create the project generation shebang PR as I've already got the changes that made it possible for me to run it...

PR #233, opened 2022-03-15
Barrel-only IR→TB

aehart:
Preliminary but synthesizable version of a barrel-only IR→TB chain.

Corresponding test vectors were produced with an untagged version of the emulation. The commit is given in emData/download.sh, and I will create a tag once I have more confidence in the test vectors and the relevant code is merged into the emulation.


meisonlikesicecream:
Looks fine. I'm trying to run it myself before approving anything and it is taking awhile... I'll let you know if it crashes :-)

Did you write the IntegrationTests/BarrelConfig/IRtoTB/script/compileHLS.tcl and IntegrationTests/BarrelConfig/IRtoTB/script/makeProject.tcl by hand?

aehart:
> Looks fine. I'm trying to run it myself before approving anything but it is taking awhile... I'll let you know if it crashes :-)

Thanks for reviewing this! And yeah, everything takes a long time since the chain is so beefy now.

> 
> Did you write the IntegrationTests/BarrelConfig/IRtoTB/script/compileHLS.tcl and IntegrationTests/BarrelConfig/IRtoTB/script/makeProject.tcl by hand?

I did, using the magic of grep, sed, and awk. I guess it wouldn't be hard to make the project generation scripts write these, but I was lazy.

FYI, I just rebased this branch after merging #232, so now the VHDL files are automatically generated. The results *should* be identical.

meisonlikesicecream:
> 

Ah okay, yeah I was wondering if we want the projgen scripts to create those files in the future

meisonlikesicecream:
@aehart You don't happen to have seen the following crash when running the barrel-only simulation?
Screenshot 2022-04-14 at 16 02 37
Currently testing if it has to do with the "number of open files" limit or something...

meisonlikesicecream:
> @aehart You don't happen to have seen the following crash when running the barrel-only simulation? Screenshot 2022-04-14 at 16 02 37 Currently testing if it has to do with the "number of open files" limit or something...

Okay yes I think it's the number of open files that caused the crash. If I increased my default limit of 1024 files to 4096 it works. For "fun", I also tried changing the limit to just 200, but then Vivado didn't even crash, it just got stuck after "Starting static elaboration." (like completely stuck, it was running over the easter weekend).

Is it possible to add e.g. `ulimit -Sn 2048` to one of the scripts? Or write something about it in the README.

meisonlikesicecream:
Is the Barrel-only chain not in the CI because it would take too long?

meisonlikesicecream:
How many errors/inconsistent stubs do we expect after running the simulation? I seem to get 160465 accumulated errors after running CompareMemPrints.py, which seems like a lot :-)

tomalin:
CERN Linux supports both bash & tcsh scripts, so some people like myself may use tcsh. The command "ulimit" does not exist in tcsh, and I can't find any way in tcsh of setting the open file limit. I fear we'll therefore have to tell people they need to compile our code inside bash.

pwittich:
As a long time tcsh user I think the equivalent command is `limit descriptors N` but I can't test it in this context. Do our commands not specify which shell they are run under?

aehart:
I may have found the cause of both the too-many-files-open error and the inconsistencies in the output memories.

Looking at the automatically generated test bench, the following block writes the contents of the SP memories:
```vhdl
    SP_14_loop : for var in enum_SP_14 generate
    begin
      writeSP_14 : entity work.FileWriter
      generic map (
        FILE_NAME => FILE_OUT_SP_14&memory_enum_to_string(var)&outputFileNameEnding,
        RAM_WIDTH => 14,
        NUM_PAGES => 2
      )
      port map (
        CLK       => CLK,
        ADDR      => SP_14_mem_AV_writeaddr(var),
        DATA      => SP_14_mem_AV_din(var),
        WRITE_EN  => SP_14_mem_A_wea(var),
        START     => TC_DONE,
        DONE      => TE_DONE
      );
    end generate SP_14_loop;
```
What doesn't make sense is for the `START` port of the `FileWriter` to be connected to `TC_DONE`. It should be connected to `VMR_DONE`. There is similar weirdness with the `START` ports of the other intermediate memory writers.

I tracked this down to makeBarrelConfig.py in the project_generation_scripts repo outputting processingmodules.dat in alphabetical order. It turns out that generator_hdl.py assumes the processing module categories are in the same order they are run in, i.e., all the IRs are first in processingmodules.dat, all the VMRs are second, etc.

I have fixed makeBarrelConfig.py in https://github.com/cms-L1TK/project_generation_scripts/pull/44. I suggest we merge that, then I can update the submodule on this branch and @meisonlikesicecream can re-test. Hopefully this will allow the output files to be properly closed by the FileWriter and we won't have to worry about this ulimit nonsense.

aehart:
> Is the Barrel-only chain not in the CI because it would take too long?

Yeah that was why I didn't include it. I don't think we want to wait so many hours for the CI tests to finish.

aehart:
> I have fixed makeBarrelConfig.py in [cms-L1TK/project_generation_scripts#44](https://github.com/cms-L1TK/project_generation_scripts/pull/44). I suggest we merge that, then I can update the submodule on this branch and @meisonlikesicecream can re-test. Hopefully this will allow the output files to be properly closed by the FileWriter and we won't have to worry about this ulimit nonsense.

The submodule has been updated.

meisonlikesicecream:
> > I have fixed makeBarrelConfig.py in [cms-L1TK/project_generation_scripts#44](https://github.com/cms-L1TK/project_generation_scripts/pull/44). I suggest we merge that, then I can update the submodule on this branch and @meisonlikesicecream can re-test. Hopefully this will allow the output files to be properly closed by the FileWriter and we won't have to worry about this ulimit nonsense.
> 
> The submodule has been updated.

Now the accumulated errors have gone down to 5951, so that seems more reasonable!

However, I still need to set ulimit to 4096... :-(

aehart:
> > > I have fixed makeBarrelConfig.py in [cms-L1TK/project_generation_scripts#44](https://github.com/cms-L1TK/project_generation_scripts/pull/44). I suggest we merge that, then I can update the submodule on this branch and @meisonlikesicecream can re-test. Hopefully this will allow the output files to be properly closed by the FileWriter and we won't have to worry about this ulimit nonsense.
> > 
> > 
> > The submodule has been updated.
> 
> Now the accumulated errors have gone down to 5951, so that seems more reasonable!

That's progress!

> 
> However, I still need to set ulimit to 4096... :-(

I still don't understand why so many files are being opened. And I don't know if our test bench or the Vivado simulation itself is the culprit.

Let me think about it a little and, worst case scenario, I can add instructions to a README for increasing the file descriptor limit in Bash and CSH.

meisonlikesicecream:
> > > > I have fixed makeBarrelConfig.py in [cms-L1TK/project_generation_scripts#44](https://github.com/cms-L1TK/project_generation_scripts/pull/44). I suggest we merge that, then I can update the submodule on this branch and @meisonlikesicecream can re-test. Hopefully this will allow the output files to be properly closed by the FileWriter and we won't have to worry about this ulimit nonsense.
> > > 
> > > 
> > > The submodule has been updated.
> > 
> > 
> > Now the accumulated errors have gone down to 5951, so that seems more reasonable!
> 
> That's progress!
> 
> > However, I still need to set ulimit to 4096... :-(
> 
> I still don't understand why so many files are being opened. And I don't know if our test bench or the Vivado simulation itself is the culprit.
> 
> Let me think about it a little and, worst case scenario, I can add instructions to a README for increasing the file descriptor limit in Bash and CSH.

Don't they all have to be open if we want to continuously write the stubs to the files? In the FileWriter*.vhd the files are not closed until all events have been processed, so I guess it's the testbench?

aehart:
> > > > > I have fixed makeBarrelConfig.py in [cms-L1TK/project_generation_scripts#44](https://github.com/cms-L1TK/project_generation_scripts/pull/44). I suggest we merge that, then I can update the submodule on this branch and @meisonlikesicecream can re-test. Hopefully this will allow the output files to be properly closed by the FileWriter and we won't have to worry about this ulimit nonsense.
> > > > 
> > > > 
> > > > The submodule has been updated.
> > > 
> > > 
> > > Now the accumulated errors have gone down to 5951, so that seems more reasonable!
> > 
> > 
> > That's progress!
> > > However, I still need to set ulimit to 4096... :-(
> > 
> > 
> > I still don't understand why so many files are being opened. And I don't know if our test bench or the Vivado simulation itself is the culprit.
> > Let me think about it a little and, worst case scenario, I can add instructions to a README for increasing the file descriptor limit in Bash and CSH.
> 
> Don't they all have to be open if we want to continuously write the stubs to the files? In the FileWriter*.vhd the files are not closed until all events have been processed, so I guess it's the testbench?

Ah you're right. I was thinking that the FileWriters closed their files after each event.

And by my count, there are 2429 FileWriters and 36 FileReaders in the test bench, so it makes sense that the default limit of 1024 does not work. Maybe we could close the files after each event and reopen them in append mode after the first event? I don't know how much that would slow down the simulation, but it would be more scalable than what we're doing now.

meisonlikesicecream:
> Ah you're right. I was thinking that the FileWriters closed their files after each event.
> 
> And by my count, there are 2429 FileWriters and 36 FileReaders in the test bench, so it makes sense that the default limit of 1024 does not work. Maybe we could close the files after each event and reopen them in append mode after the first event? I don't know how much that would slow down the simulation, but it would be more scalable than what we're doing now.

But if all modules work in parallel, wouldn't they all open and close at the same time? 

How stupid would it be to wait with the file opening until all events are processed...? Then all IR writers would open and close their files first, then VMRs etc etc. But then each FileWriter would have to store many lines of text before it can write so that's probably also bad... 

(or have I completely misunderstood how the VHDL works)

aehart:
> > Ah you're right. I was thinking that the FileWriters closed their files after each event.
> > And by my count, there are 2429 FileWriters and 36 FileReaders in the test bench, so it makes sense that the default limit of 1024 does not work. Maybe we could close the files after each event and reopen them in append mode after the first event? I don't know how much that would slow down the simulation, but it would be more scalable than what we're doing now.
> 
> But if all modules work in parallel, wouldn't they all open and close at the same time?

Crap, yeah I think you're right. So it would be more complicated than that.

> 
> How stupid would it be to wait with the file opening until all events are processed...? Then all IR writers would open and close their files first, then VMRs etc etc. But then each FileWriter would have to store many lines of text before it can write so that's probably also bad...

That's an interesting idea. I'm not sure how much additional RAM that would use, but I feel like it can't be _that_ much.

> 
> (or have I completely misunderstood how the VHDL works)

In any case, we can think more about how to make the FileWriters more scalable. For now, I have updated runSim.tcl to increase the limit on the number of open files. This seems to work in my testing, but maybe you can test it too, just to make sure.

Then I would propose merging while we continue to debug the discrepancies with this preliminary set of test vectors.

tomalin:
> > > Ah you're right. I was thinking that the FileWriters closed their files after each event.
> > > And by my count, there are 2429 FileWriters and 36 FileReaders in the test bench, so it makes sense that the default limit of 1024 does not work. Maybe we could close the files after each event and reopen them in append mode after the first event? I don't know how much that would slow down the simulation, but it would be more scalable than what we're doing now.
> > 
> > 
> > But if all modules work in parallel, wouldn't they all open and close at the same time?
> 
> Crap, yeah I think you're right. So it would be more complicated than that.
> 
> > How stupid would it be to wait with the file opening until all events are processed...? Then all IR writers would open and close their files first, then VMRs etc etc. But then each FileWriter would have to store many lines of text before it can write so that's probably also bad...
> 
> That's an interesting idea. I'm not sure how much additional RAM that would use, but I feel like it can't be _that_ much.
> 
> > (or have I completely misunderstood how the VHDL works)
> 
> In any case, we can think more about how to make the FileWriters more scalable. For now, I have updated runSim.tcl to increase the limit on the number of open files. This seems to work in my testing, but maybe you can test it too, just to make sure.
> 
> Then I would propose merging while we continue to debug the discrepancies with this preliminary set of test vectors.

If we modify the code to store in memory all the output data from all events, so that all files can be opened & written at end of job, I'm concerned about the memory use. In future, we may want to run over tens of thousands of events to search for rare errors.

My suspicion is that if we modified the FileWriter so that it opened & closed the file every time it's called, this may work. As it is called from the test-bench, and the file open & close commands are not synthesisable, there's a chance that they are executed sequentially for different modules according to a simulation time, rather than real clock time. One would have to try it to see ...

pwittich:
do we have to run the file writer every time? Maybe we should disable it for e.g. people doing HDL simulation, if that's not already happening.

aehart:
> do we have to run the file writer every time? Maybe we should disable it for e.g. people doing HDL simulation, if that's not already happening.

I don't think we have to, but right now, the FileWriters open their files immediately on the first clock cycle. So they would need to be tweaked.

I'm not sure how much of the time that the simulation takes is due to writing to these files, but it might not be insignificant, so maybe it's worth doing at some point.

aehart:
> > do we have to run the file writer every time? Maybe we should disable it for e.g. people doing HDL simulation, if that's not already happening.
> 
> I don't think we have to, but right now, the FileWriters open their files immediately on the first clock cycle. So they would need to be tweaked.
> 
> I'm not sure how much of the time that the simulation takes is due to writing to these files, but it might not be insignificant, so maybe it's worth doing at some point.

Actually, I guess we could just add a constant to the test benches that controls whether the FileWriters are called at all. That should be pretty easy to add, but I think it can be independent of this PR.

meisonlikesicecream:
Sorry for being slow, this seems to work now! :-)

meisonlikesicecream (approved):

PR #235, opened 2022-04-26
Ian ci

tomalin:
When a PR is made to the HLS repo, triggering CI to run in github, the generic script used to trigger CI in the mirror gitlab repo has been replaced. It is no longer:
https://github.com/aperloff/gitlab-mirror-and-ci-action
but now:
https://github.com/cms-L1TK/gitlab-mirror-and-ci-action

The new version is the same one used in the CMSSW L1Trk CI tests, and although it is based heavily on the old one, includes mods to allow its use both with CMSSW & HLS. It also prints out more clearly the web address of the gitlab CI tests, making it easier to find them.


.github/workflows/github_CI.yml
```diff
@@ -0,0 +1,67 @@
...
+
+    - name: Redefine branch on push
+      if: github.event_name == 'push'
+      run: |
+        echo "current_branch=$(echo ${{ github.ref }} | sed -E 's|refs/[a-zA-Z]+/||')" >> $GITHUB_ENV
+
+    - name: Trigger gitlab CI run
+      # Communicate between gitlab & github, and trigger gitlab CI run.
+      # (Forked from https://github.com/SvanBoxel/gitlab-mirror-and-ci-action).
+      #uses: aperloff/gitlab-mirror-and-ci-action@master
```
aperloff, line 42: Remove unnecessary comments

.github/workflows/github_CI.yml
```diff
@@ -0,0 +1,67 @@
...
+      uses: cms-L1TK/gitlab-mirror-and-ci-action@master
+      #with:
```
aperloff, line 44: Same here

.github/workflows/github_CI.yml
tomalin: done

.github/workflows/github_CI.yml
tomalin: I've removed the obsolete reference to  aperloff/gitlab-mirror-and-ci-action@master at L42. Is this what you are referring to?

tomalin: done

aperloff (approved):

PR #239, opened 2022-05-23
Synchronization with emulation

aehart:
This PR updates the test vectors to those generated with the latest version of the emulation. Some of the instructions in the README were also corrected.


tomalin:
The instructions for running the emulation in https://github.com/cms-L1TK/firmware-hls/blob/fw_synch/README.md look incomplete. To run the summer chain emulation, the reduced wiring map data files would need copying into TrackFindingTracklet/data/ . To get them, one presumably has to:
1) Run emulation for the full Hybrid configuration, with writeConfig_=true, so the full wiring map .txt files are written.
2) Checkout the python project generation scripts, and run https://github.com/cms-L1TK/project_generation_scripts/blob/master/makeReducedConfig.py , to make the reduced wiring files from the full ones.

Also, the instructions should explain how to run the emulation for barrel-only.

aehart:
@tomalin I have updated the README with your suggestions and reorganized it a bit to make the configuration-specific instructions clearer.

PR #240, opened 2022-05-27
First version of code for a combined (TP, MP) module chain

aryd:
This pull request contains first version of code for a 'skinny' chain for combined modules. Code runs the IR, VMRCM, TP, MP, and TB modules. Tracks are produced - but there are still some issues to be resolved. (Likely by fixing start signals of the TP from the VMR.)

This code enables the rewind option for the TP, and MP modules. We need to look more at timing performance in these modules after enabling the rewind option. 

A separate pull request to the project_generation_scripts repository will be made for generating the configuration.


IntegrationTests/ReducedCombinedConfig/script/MemPrints
```diff
@@ -0,0 +1 @@
+../../../emData/MemPrintsReducedCM
```
aehart, line 1: @aryd Please remove this symlink, since it is generated by compileHLS.tcl.

IntegrationTests/ReducedCombinedConfig/script/makeProject.tcl
```diff
@@ -0,0 +1,65 @@
+# Create Vivado project, with user HDL files & IP.
+# Run this in IntegrationTests/xyz/script/ 
+
+# Create project
+set projName "Work"
+set FPGA "xcvu13p-fhgb2104-2-e"
```
aehart, line 6: @aryd Does this not cause problems with the part for the HLS modules being a VU7P?:
https://github.com/cms-L1TK/firmware-hls/blob/a3e5decfe1ed903ab9c66c948ba16e3811c6fc68/project/settings_hls.tcl#L22

IntegrationTests/common/hdl/tf_mem.vhd
```diff
@@ -114,6 +135,7 @@ process(clka)
   variable addr_in_page : integer := 0;
 begin
   if rising_edge(clka) then -- ######################################### Start counter initially
+    report "tm_mem "&NAME&" nent(0) nent(1) "&to_bstring(nent_o(0))&" "&to_bstring(nent_o(1));
```
aehart, line 138: @aryd These reports, and the similar ones in tf_mem_bin_cm5, add a _lot_ of output to the simulation. Would it be possible to have a flag for turning this on or off?

IntegrationTests/common/hdl/tf_mem_bin_cm.vhd
```diff
@@ -0,0 +1,211 @@
...
+      elsif (regceb='1') then
+        for imem in (NCOPY-1) downto 0 loop
+          doutb(imem) <= sv_RAM_row(imem);
+        end loop;
+      end if;
+    end if;
+  end process;
+end generate MODE;
+
+end rtl;
```
aehart, line 211: @aryd Is this the module that doesn't work because of some unsupported VHDL feature in Vivado? Since it isn't used anywhere, can we just remove it?

IntegrationTests/common/hdl/tf_mem_bin_cm5.vhd
```diff
@@ -0,0 +1,291 @@
...
+  for idx in 63 downto 0 loop
+    sl_str_v := to_bstring(slv(idx));
+    res_v(idx*5) := sl_str_v(1);
+    res_v(idx*5+1) := sl_str_v(2);
+    res_v(idx*5+2) := sl_str_v(3);
+    res_v(idx*5+3) := sl_str_v(4);
+    res_v(idx*5+4) := '-';
+  end loop;
+  return res_v;
+end function;
```
aehart, line 119: @aryd Since these functions are of general use, and two of them are essentially duplicated in tf_mem.vhd, can they be moved to a common place, like tf_pkg.vhd?

IntegrationTests/common/hdl/tf_pkg.vhd
```diff
@@ -61,13 +63,22 @@ package tf_pkg is
 
   -- ########################### Types ###########################
 
+  -- type std_logic_aoa is array(natural range <>) of std_logic_vector;
```
aehart, line 66: @aryd Can you remove this comment?

TestBenches/TrackletProcessor_test.cpp
```diff
@@ -28,13 +28,13 @@
...
+  #define ITC_ C_
 #endif
 #if !defined MODULE_
-  #define MODULE_ TP_L1L2D_
+  #define MODULE_ TP_L1L2C_
 #endif
 #if !defined TOP_FUNC_
-  #define TOP_FUNC_ TrackletProcessor_L1L2D
+  #define TOP_FUNC_ TrackletProcessor_L1L2C
 #endif
```
aehart, line 38: @aryd These macros should correspond to this module here:
https://github.com/cms-L1TK/firmware-hls/blob/a3e5decfe1ed903ab9c66c948ba16e3811c6fc68/project/script_TP.tcl#L15-L17
Otherwise, the C/RTL cosimulation will fail. It doesn't fail right now because the `module_to_export` is not in the list of `modules_to_test` here:
https://github.com/cms-L1TK/firmware-hls/blob/a3e5decfe1ed903ab9c66c948ba16e3811c6fc68/project/script_TP.tcl#L9-L14

Can you set the `module_to_export` in script_TP.tcl to `TP_L1L2C` to match these macros and so that the cosimulation runs?

emData/download.sh
```diff
@@ -11,6 +11,12 @@ luts_url_reduced="https://cernbox.cern.ch/index.php/s/GJZA1zLnWg3hP4y/download"
 # Combined modules
 memprints_url_cm="https://cernbox.cern.ch/index.php/s/YcqX3KUgFdZyMG6/download"
 luts_url_cm="https://cernbox.cern.ch/index.php/s/lKrxzKJ0XmelE0j/download"
+# Reduced Combined modules                                                                                                                     
+#memprints_url_reducedcm="https://aryd.web.cern.ch/aryd/MemPrints_CombinedReduced_220309.tgz"
+#luts_url_reducedcm="https://aryd.web.cern.ch/aryd/LUTs_CombinedReduced_220309.tgz"
+#memprints_url_reducedcm="https://aryd.web.cern.ch/aryd/MemPrints_CombinedReduced_220411.tgz"
+memprints_url_reducedcm="https://aryd.web.cern.ch/aryd/MemPrints_CombinedReduced_220503.tgz"
+luts_url_reducedcm="https://aryd.web.cern.ch/aryd/LUTs_CombinedReduced_220411.tgz"
```
aehart, line 19: @aryd Can you add a recipe for reproducing these files to the README? There are now various sections for the different configurations, so you can add another section for this configuration.

TrackletAlgorithm/VMRouterCM.h
```diff
@@ -280,9 +280,11 @@ void VMRouterCM(const BXType bx, BXType& bx_o,
 		AllStub<OutType> allstub = (disk2S) ? stubDisk2S.raw() : stub.raw();
 
 		// Write stub to all memory copies
-		for (int n = 0; n < nAllCopies; n++) {
+		if (nAllCopies > 0) {
```
aehart, line 283: @aryd I don't understand why this `if` block is needed. Isn't the `for` loop skipped anyway if `nAllCopies` is nonpositive?

IntegrationTests/ReducedCombinedConfig/script/LUTs
```diff
@@ -0,0 +1 @@
+../../../emData/LUTsCMReduced
```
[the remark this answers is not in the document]
aryd: Removed

IntegrationTests/ReducedCombinedConfig/script/MemPrints
aryd: Removed

IntegrationTests/ReducedCombinedConfig/script/makeProject.tcl
aryd: Had not noticed any problems; but we should probably have this consistent - does it need to be set in more than one place?

IntegrationTests/common/hdl/FileWriter.vhd
```diff
@@ -22,6 +22,7 @@ entity FileWriter is
     FILE_NAME  : string;   --! Name of .txt file to be written
     RAM_WIDTH  : natural := 18;    --! RAM data width
     NUM_PAGES  : natural := 2;     --! Number of pages in RAM memory
+    PAGE_LENGTH : natural := 128;     --! Page length
```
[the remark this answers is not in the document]
aryd: Can you just use PAGE_LENGTH (in C++ you would do something like tf_pkg::PAGE_LENGTH).

IntegrationTests/common/hdl/tf_lutdat.vhd
```diff
@@ -0,0 +1,85 @@
...
+  process(clk)
+  begin
+    if rising_edge(clk) then
+      dout <= dout_tmp;
+    end if;
+  end process;
+end generate MODE;
+
+
+end architecture behavioral;
```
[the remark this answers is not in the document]
aryd: Andrew, this was discussed in an email exchange with the subject "Unhelpful error message -> New error" around April 8. The 'old' LUT were written out in C++ format, but I have now change the LUT writing such that it produces a .dat file which is the numbers in hex. I think this is simpler and safer to parse when reading in the data from vcdl. So I think we should remove the old file reading. But we need to update all the test vectors before removing the old code.

IntegrationTests/common/hdl/tf_mem.vhd
aryd: I guess these should just be removed - though I found them very helpful in debugging... I'll see if I can protect them with a flag

TestBenches/TrackletProcessor_test.cpp
aryd: Fixed.

TrackletAlgorithm/VMRouterCM.h
aryd: I agree - I don't see why this should be needed - but if I leave it out I get:

ERROR: [SYNCHK 200-61] ../../../TrackletAlgorithm/AllStubMemory.h:86: unsupported memory access on variable 'memoriesAS.dataarray_.data_.V' which is (or contains) an array with unknown size at compile time.
INFO: [SYNCHK 200-10] 1 error(s), 1 warning(s).
ERROR: [HLS 200-70] Synthesizability check failed.

If you see a better fix let me know.

IntegrationTests/common/hdl/tf_pkg.vhd
aryd: Done

IntegrationTests/common/hdl/tf_mem_bin_cm5.vhd
aryd: Done - these are now in the tf_pkg.vhd

IntegrationTests/common/hdl/tf_mem_bin_cm.vhd
aryd: Removed.

IntegrationTests/common/hdl/tf_mem.vhd
aryd: Done - I added a flag, DEBUG, that can be set to true if you want this printout.

emData/download.sh
aryd: This description has been added. This was tested in CMSSW_12_5_0_pre2. With the code on the branch ReducedModuleConfigurationFixes (PR#182).

IntegrationTests/ReducedCombinedConfig/script/makeProject.tcl
aehart: Consistency would be good. For the HLS modules, it needs to be set in project/settings_hls.tcl, and for each of the chains, it needs to be set in makeProject.tcl in that chain's script/ directory.

Here, I would suggest setting `FPGA` to the VU7P for now, to match the other chains and since this chain should easily fit on one. Then we can think about how to make the parts consistent as we transition our testing to the VU13P.

IntegrationTests/common/hdl/tf_lutdat.vhd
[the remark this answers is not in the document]
aehart: Ah OK. Sorry I forgot about that exchange. Makes sense to have this separate version for now.

IntegrationTests/common/hdl/FileWriter.vhd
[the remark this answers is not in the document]
aehart: Yes, I believe you can just use `PAGE_LENGTH`.

aehart:
@aryd FYI, I modified .gitlab-ci.yml so that the C-simulation for the MP can fail until the test vectors are regenerated.

I also updated the project_generation_scripts/ submodule so that backwards compatibility is maintained for the chains of non-combined modules. Those changes are in cms-L1TK/project_generation_scripts#46.

IntegrationTests/common/hdl/FileWriter.vhd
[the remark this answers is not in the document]
aryd: OK - I finally manage to test this an it was OK. SO I have pushed this change now.

aehart (approved):
Looks good to me. Approved!

PR #241, opened 2022-07-08
Fixed gcc10 compilation error from unsigned int comparison

nzipper:
This pull request fixes a compilation error in gcc10 where an unsigned int is compared to be less than the NCOPY variable, which is 0 for layer 1, implying the comparison will always return false. A check is performed to ensure that the memories are only looped over and written if NCOPY != 0.

### DIFF
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -114,7 +114,7 @@ class MemoryTemplateBinnedCM{
     if(!NBIT_BX) ibx = 0;
     if (nentry_ibx < getNEntryPerBin()-1) { // Max 15 stubs in each memory due to 4 bit nentries
       // write address for slot: getNEntryPerBin() * slot + nentry_ibx
-      writememloop: for (signed int icopy=0; icopy< (signed) NCOPY;icopy++) {
+      writememloop: for (signed int icopy=0; icopy< (signed) NCOPY;icopy++) { // Casting to signed int to avoid unsigned int comparison compilation error
 #pragma HLS unroll
         dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry_ibx] = data;
       }
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -115,13 +115,10 @@ class MemoryTemplateBinnedCM{
     if (nentry_ibx < getNEntryPerBin()-1) { // Max 15 stubs in each memory due to 4 bit nentries
       // write address for slot: getNEntryPerBin() * slot + nentry_ibx
   
-    writememloop:
-    if constexpr (NCOPY != 0) {
-      for (unsigned int icopy=0; icopy< NCOPY;icopy++) {
+    writememloop: for (signed int icopy=0; icopy< (signed) NCOPY;icopy++) {
 #pragma HLS unroll
     dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry_ibx] = data;
         }
-    }
       
       #ifdef CMSSW_GIT_HASH
       ap_uint<kNBitsRZBinCM> ibin;
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -114,11 +114,10 @@ class MemoryTemplateBinnedCM{
     if(!NBIT_BX) ibx = 0;
     if (nentry_ibx < getNEntryPerBin()-1) { // Max 15 stubs in each memory due to 4 bit nentries
       // write address for slot: getNEntryPerBin() * slot + nentry_ibx
-  
-    writememloop: for (signed int icopy=0; icopy< (signed) NCOPY;icopy++) {
+      writememloop: for (signed int icopy=0; icopy< (signed) NCOPY;icopy++) {
 #pragma HLS unroll
-    dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry_ibx] = data;
-        }
+        dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry_ibx] = data;
+      }
       
       #ifdef CMSSW_GIT_HASH
       ap_uint<kNBitsRZBinCM> ibin;
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -115,11 +115,14 @@ class MemoryTemplateBinnedCM{
     if (nentry_ibx < getNEntryPerBin()-1) { // Max 15 stubs in each memory due to 4 bit nentries
       // write address for slot: getNEntryPerBin() * slot + nentry_ibx
   
-    writememloop:for (unsigned int icopy=0;icopy<NCOPY;icopy++) {
+    writememloop:
+    if constexpr (NCOPY != 0) {
+      for (unsigned int icopy=0; icopy< NCOPY;icopy++) {
 #pragma HLS unroll
-	dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry_ibx] = data;
-      }
-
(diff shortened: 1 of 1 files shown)


tomalin:
This PR needs a description at the top to explain what it is about.

nzipper:
This pull request fixes a compilation error in gcc10 where an unsigned int is compared to be less than the NCOPY variable, which is 0 for layer 1, implying the comparison will always return false. A check is performed to ensure that the memories are only looped over and written if NCOPY != 0.

From MemoryTemplateBinnedCM.h, "NCOPY: is the number of memory copies - same as number of TE used". This will be 0 for layer 1, as defined in L18 of TrackletAlgorithm/Constants.h.

nzipper:
The failure in the CI pipeline to build is do to hls interpreting "if constexpr...", is there a pragma or different expression to handle this at compile time?

pwittich:
> The failure in the CI pipeline to build is do to hls interpreting "if constexpr...", is there a pragma or different expression to handle this at compile time?

This is a c++17 construct and vivado hls supports c++14 at best...  a cursory web search suggests that `std::enable_if` might be similar in c++14 but I don't know if it is supported in HLS.

nzipper:
I found a different method to fix this compilation error, simply by casting NCOPY to a signed int and matching the type of the iterator. This is maybe the most minimalist way to avoid the error, but let me know if you think this isn't the best way to handle it.

nzipper:
Is this what you mean by changing the label position?

pwittich:
> Is this what you mean by changing the label position?

that'll do yes thanks

meisonlikesicecream:
> What is the physical meaning of variable NCOPY? And what is the significance of it sometimes being 0? At the line
> 
> https://github.com/cms-L1TK/firmware-hls/blob/9d9ad27db384c4fc0c6e4b0eae608b44153a2839/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L45
> 
> , this causes array dataarray_ to have one dimension that is of size 0, which isn't good C++.
> @meisonlikesicecream generate_VMRCM.py instantiates memory VMStubTEOuterMemoryCM which uses MemoryTemplateBinnedCM, and sets its NCOPY template parameter to kNTEUnitsLayerDisk[layerdisk], which as Noah points out is 0 for some layers. I don't understand why we're bothering to instantiate an object representing 0 memories?

Where do you see that it instantiates a VMStubTEOuterMemoryCM with size 0? Which top function? 

It should only instantiate it if that VMR uses that memory, but maybe my values are outdated: https://github.com/cms-L1TK/firmware-hls/blob/9d9ad27db384c4fc0c6e4b0eae608b44153a2839/emData/generate_VMRCM.py#L363

tomalin:
@nzipper given that the when compiled with Vivado_HLS, we don't have a problem with NCOPY = 0, apparently because of the "if has vmste_outer" statement that @meisonlikesicecream mentions above, why do we have one when VMRCM is called from CMSSW?

tomalin:
I'm nervous that the gcc10 warning may indicate a genuine bug in the VMR Future SW. Is it possible that this is instantiating memories with NCOPY=0, whereas in FW compilation the VMR HLS code & VMR HLS test-benches avoid doing instantiating memories with NCOPY=0, because of their "if has vmste_outer" statement? 
I've not checked, so this fear may be unjustified. But if you're not sure either, please check.

tomalin:
According to LUTsCM/wires.dat, VMR_L2PHIA writes to 3 VMSTE memories, whereas VMR_L1PHIA writes to none. The reason the compilation with Vivado_HLS gives no errors is that whilst TopFunctions/VMRouterCMTop_L2PHIA.cc passes the VMRouterCM constructor an array of VMStubTEOuterMemoryCM of dimension 3.; TopFunctions/VMRotuerCMTop_L1PHIA.cc instead passes the VMRouterCM constructor a nullptr.
The reason our CMSSW code gives compiler errors is that it always calls the VMRouterCM constructor with an array of VMStubTEOuterMemoryCM of dimension 1 https://gitlab.cern.ch/ejclemen/cmssw_trackfinding_hlsframework/-/blob/master/TrackFindingTrackletHLS/plugins/VMRProducer.cc#L295 . If we could find a way of passing a nullptr or array of size 0 from here instead, then the compilation error would go away, without needing to change the HLS code.
We could compile CMSSW without errors, and without needing any changes to our HLS repo, if we could do something similar

tomalin:
If we wish to avoid instantiating a teMemories array of size 1 at https://gitlab.cern.ch/ejclemen/cmssw_trackfinding_hlsframework/-/blob/master/TrackFindingTrackletHLS/plugins/VMRProducer.cc#L295 when the object inside it contains NCOPY=0 copies of the memory, a possible solution is to replace this code by:

    unsigned int nTE = (kNTEUnitsLayerDisk[layerToUse-1] == 0) ? 0 : 1;
    VMStubTEOuterMemoryCM<rt, 3, 3, kNTEUnitsLayerDisk[layerToUse-1]>* teMemori
es = nullptr;
    if (nTE) {
      auto myPtr = std::make_unique<VMStubTEOuterMemoryCM<rt, 3, 3, kNTEUnitsLayerDisk[layerToUse-1]>[]>(nTE);
      teMemories = myPtr.get();
    }

nzipper:
Ahh ok I see what you're saying, thanks for the clarification here. I was trying to fix it exclusively from the HLS side because of the second instantiation. So I think if we combine the small fix I made in this repo with your proposed change in the emulation code we can fix possible errors from both instantiations. 

I added the code you have above to VMRProducer.cc and everything compiles/runs properly, and I've double-checked that we are now successfully creating a null pointer rather than an unnecessary array. All the changes here have already passed the CI pipeline, so should I just submit a separate PR in the other repo? Or is there some procedure for referencing the other PR as well?

tomalin:
I'll approve your HLS PR. Though add a comment to the sign(NCOPY) line explaining what is going on, since otherwise someone from the HLS group might remove it again!

The change I propose to the CMSSW repo is a matter of style, imitating how the HLS top-level avoids problems. Though the C++ compiler seems happy without this fix, I think. So it's not essential. Also, I think my proposal is buggy. I put "auto myPtr = " inside a bracket, so the unique_ptr will cease to exist as soon as the code exits the bracket. It would need rewriting to avoid this ...

nzipper:
I've added the comment to address the fix. 

Yes error-wise it now has all the info it needs at compile time so it should work correctly. It seems like a matter of if we prefer this superfluous array being instantiated or keeping the existing code simpler and less future error-prone. I'll think about how to clean up these additional lines a little more, but it seems functional either way.

tomalin:
@meisonlikesicecream & @nzipper I just noticed that the VMR HLS now fails the CI test https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/4258526 . The cause is apparently that it produces so much debug printout that it crashes the system. Which of you added this printout? Can we get it under control?

meisonlikesicecream:
> @meisonlikesicecream & @nzipper I just noticed that the VMR HLS now fails the CI test https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/4258526 . The cause is apparently that it produces so much debug printout that it crashes the system. Which of you added this printout? Can we get it under control?

I don't think the job log crashes the system, I think there's an actual error somewhere but we just can't see it because it stopped writing to the job log. Another recent PR has the same warning but it passes the simulation: https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/23132729

And to who added the printout, it's been there from before my time and has been very helpful when debugging :-)

meisonlikesicecream:
How did you manage to merge a branch that did not pass the CI? I thought that was blocked?

tomalin:
The previous merged PR https://github.com/cms-L1TK/firmware-hls/pull/239 passed CI, so something must have changed since then.

tomalin:
I just checked out "master" and ran script_VMR.tcl. It gave no errors, so I don't know why CI fails. If @meisonlikesicecream can make another PR to reduce the debug printout, we can understand what is going on.

meisonlikesicecream:
It is strange that the previous commit passed the CI (https://github.com/cms-L1TK/firmware-hls/pull/241/commits/48b0aac2c5d4dfa02d322d23f01d3c18fee0950d) but the only difference between that one and the newest one is a comment so it shouldn't matter?!

PR #243, opened 2022-07-21
Vmr debug printout fix

meisonlikesicecream:
Reduces the number of printouts when running CSim unless a VMR_DEBUG/VMRCM_DEBUG flag is specified, hopefully that will stop the CI from exceeding the job log output limit ( https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/23421481) as discussed here https://github.com/cms-L1TK/firmware-hls/pull/241.

### DIFF
--- a/TrackletAlgorithm/VMRouter.h
+++ b/TrackletAlgorithm/VMRouter.h
@@ -676,11 +676,10 @@ void VMRouter(const BXType bx, BXType& bx_o, const int fineBinTable[], const int
 				memoriesAS[n].write_mem(bx, allstub, i);
 			}
 
-// For debugging
-#ifndef __SYNTHESIS__
+#ifndef VMR_DEBUG
 			edm::LogVerbatim("L1trackHLS") << std::endl << "Stub index no. " << i << std::endl << "Out put stub: " << std::hex << allstub.raw() << std::dec
 					<< std::endl;
-#endif // DEBUG
+#endif // VMR_DEBUG
 
 
 		/////////////////////////////////////////////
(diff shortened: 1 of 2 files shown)


tomalin:
1) Please add a commented out
#define VMR_DEBUG
statement somewhere.
2) If someone uncomments this line, then VMR_DEBUG should automatically be set undefined again if __SYNTHESIS__ is defined.

tomalin:
Can we remove the "endl" statements at the end of the LogVerbatim lines, to reduce debug printout even if VMR_DEBUG is defined. (See my Skype message to HLS forum)?

meisonlikesicecream:
Ah we can see why it fails now, it's not the VMR/VMRCM itself: https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/23434698

tomalin:
True, the VMR CI archive was too big. The successful one from an earlier PR https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/22324732/artifacts/browse/project/vmrouter/ was 1.3GB, whereas https://forum.gitlab.com/t/uploading-artifacts-to-coordinator-too-large-archive/41441 implies the default limit is 1GB, so we've always been on the edge.

tomalin:
The largest file in the archive is csim.exe, one copy of which is created for each VMR that is simulated. If all these are deleted, it reduces the size of the VMR CI archive from 1.4GB to 0.9GB.

tomalin (approved):
This solves the "log file too long" CI error. Approved.

PR #244, opened 2022-09-07
VMR move constants

meisonlikesicecream:
As discussed with @tomalin (and @nzipper??) moving and changing some constants in the VMR might simplify the inclusion of the VMR into CMSSW. 

- The VMR(CM) used an enum called phiRegions defined in the parameters file, but this has now been moved Constants.h and is now an enum of int instead of char.
- MatchCalculator and TrackletCalculator now also uses this phiRegions enum in Constants.h
- Removed trailing whitespaces...*

**Why draft??!**
The use of getNum-memory functions for defining the size of the memory arrays was also mentioned. I have not changed that yet as I was not sure if Ian and Noah wanted me to do this, or why this seemed to cause issues?

If there is anything else that should be moved/removed/changed let me know.

_\* I noticed that I forgot to turn off the "remove trailing whitespaces" tool in my editor, so it looks like I've changed more than I actually have... I can add back the trailing whitespaces whenever I've got time (as people have complained about removing them before, even though I think they shouldn't be there), OR the people reviewing can just press the "hide whitespace" button and all the ugly whitespaces will be removed but you won't see the whitespace changes when comparing:_
Screenshot 2022-09-07 at 15 04 27
:-)))))


aehart:
For what it's worth, I love that you removed trailing whitespace, but then, I'm kind of zealous about this. Every day is a struggle to not open a PR where I just remove trailing whitespace (and tabs) from every file in the repo.

So I vote to keep those changes.

tomalin:
This looks a nice simplification. Can https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MatchProcessor.h#L308 also be replaced with your new phiRegions enum in Constants.h?

tomalin:
I've changed this from "Draft" to "Ready for review". The question of whether functions such as getNum-memory should be modified to give a more uniform style requires thought, and depends also on use in the Future CMSSW code. If necessary, we can fix this later as a separate PR. But the current PR is needed now to simplify the VMR Future CMSSW code.

meisonlikesicecream:
> This looks a nice simplification. Can https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MatchProcessor.h#L308 also be replaced with your new phiRegions enum in Constants.h?

Should be possible, yes :-)

tomalin (approved):
Looks good. Approved.

PR #245, opened 2022-09-08
LUT path update

aehart:
This PR introduces pre-simulation and pre-synthesis scripts to copy the appropriate LUTs/ directory into the Vivado project directory. This makes it easier to reference the LUTs/ directory, especially when working in a framework such as EMP, since there are no relative paths to deal with.

This PR is dependent on cms-L1TK/project_generation_scripts#47 , which is already merged.


bylsma (approved):
The changes for the LUT path fix look good to me and are ready to merge.  I'm not familiar with the requirements for the CI yaml file, but the additions look reasonable.

PR #246, opened 2022-09-20
Trackmerger v2

kwalkingshaw:
The TrackMerger compares and merges tracks by removing duplicate tracks (stubs are not yet merged as currently there is only one stub per layer on each track). This version uses 16 Comparison Modules to compare tracks and merges tracks if they have > 3 stubs in common.  Each CM has its own master track and each input track is compared against the master, if there is no match found, the input track is passed to the next CM, and so on.

- Will need to edit the code to add > 1 stub per layer
- Will need to implement q/pT bins


project/script_TM.tcl
```diff
@@ -0,0 +1,29 @@
...
+source settings_hls.tcl
+
+# data files
+add_files -tb ../emData/PD/
+config_dataflow -scalar_fifo_depth 50
+csim_design -compiler gcc -mflags "-j8"
+#csynth_design
+#cosim_design 
+#-trace_level all -rtl verilog -verbose
+#export_design -format ip_catalog
```
aehart, line 26: Please uncomment `csynth_design`, `cosim_design`, and `export_design` (without the `-flow impl` option), so they are also tested by default.

project/settings_hls.tcl
```diff
@@ -19,8 +19,9 @@ switch -glob -- $exe {
     }
     default {
         # Set the FPGA part number
-        set_part {xcvu7p-flvb2104-1-e} -tool vivado
+        #set_part {xcvu7p-flvb2104-1-e} -tool vivado
         #set_part {xcku115-flvb2104-1-e} -tool vivado
+        set_part {xcvu9p-flgc2104-1-e} -tool vivado
```
aehart, line 24: Is there a reason for switching to a VU9P?

TestBenches/TrackMerger_test.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+    // Comparing outputs
+    err_count += compareMemWithFile<TrackFitMemory<NBarrelStub, NDiskStub>>(outputTracks, fout_outputTracks.at(0), ievt, "Tracks", truncation);
+    
+    
+  }
+
+  // Handling case of err%256 == 0 
+  // if (err_count > 255) err_count = 255;
+  // return err_count;
```
aehart, line 134: Is there a reason for having this commented out? Is it just because it fails with the current set of test vectors?

TrackletAlgorithm/TrackHandler.h
```diff
@@ -0,0 +1,50 @@
+#ifndef TrackletAlgorithm_TrackHandler_h
+#define TrackletAlgorithm_TrackHandler_h
+#include "TrackFitMemory.h"
+#include "hls_stream.h"
+const int NBarrelStub = 4;
+const int NDiskStub = 4;
```
aehart, line 6: I think some `typedefs` would improve the readability of the code throughout. At least an alias for `TrackFit<NBarrelStub, NDiskStub>`, and maybe others that appear frequently. Not strictly necessary, but I think some of the long lines would be easier to digest.

TrackletAlgorithm/TrackHandler.h
```diff
@@ -0,0 +1,50 @@
...
+const unsigned int kFullBarrelStubSize = TrackFit<NBarrelStub, NDiskStub>::kBarrelStubSize * 4;
+const unsigned int kFullDiskStubSize = TrackFit<NBarrelStub, NDiskStub>::kDiskStubSize * 4;
```
aehart, line 8: Please use `NBarrelStub` and `NDiskStub` instead of the magic number `4`.

TrackletAlgorithm/TrackHandler.h
```diff
@@ -0,0 +1,50 @@
...
+const unsigned int kBarrelStubIndexSizeLSB = TrackFit<NBarrelStub, NDiskStub>::kTFBarrelStubRSize + TrackFit<NBarrelStub, NDiskStub>::kTFPhiResidSize + TrackFit<NBarrelStub, NDiskStub>::kTFZResidSize;
+const unsigned int kDiskStubIndexSizeMSB = TrackFit<NBarrelStub, NDiskStub>::kTFDiskStubRSize + TrackFit<NBarrelStub, NDiskStub>::kTFPhiResidSize + TrackFit<NBarrelStub, NDiskStub>::kTFRResidSize + TrackFit<NBarrelStub, NDiskStub>::kTFStubIndexSize;
+const unsigned int kDiskStubIndexSizeLSB = TrackFit<NBarrelStub, NDiskStub>::kTFDiskStubRSize + TrackFit<NBarrelStub, NDiskStub>::kTFPhiResidSize + TrackFit<NBarrelStub, NDiskStub>::kTFRResidSize;
+const unsigned int layerStubIndexSize = 1; // 7 stubs per layer
+
+struct TrackStruct {
+  
+  TrackFit<NBarrelStub, NDiskStub>::TrackWord _trackWord = 0; 
+  TrackFit<NBarrelStub, NDiskStub>::BarrelStubWord _barrelStubArray[4][layerStubIndexSize] = {0}; 
+  TrackFit<NBarrelStub, NDiskStub>::DiskStubWord _diskStubArray[4][layerStubIndexSize] = {0};
```
aehart, line 20: Again, please use `NBarrelStub` and `NDiskStub` instead of the magic number `4`.

TrackletAlgorithm/TrackHandler.h
```diff
@@ -0,0 +1,50 @@
...
+    void mergeTrack(TrackStruct& trk, TrackStruct& masterTrk);
+
+    void setDebugFlag(unsigned int debugFlag){
+      debug = debugFlag;
+    }
+
+    
+  private:
+    ap_uint<1> matchesFoundBarrel[4][layerStubIndexSize];
+    ap_uint<1> matchesFoundDisk[4][layerStubIndexSize];
```
aehart, line 40: Here as well, replace `4` with `NBarrelStub` and `NDiskStub`.

TrackletAlgorithm/TrackHandler.h
```diff
@@ -0,0 +1,50 @@
...
+    ap_uint<1> stubPadding = 0;
+    unsigned int debug = 0;
+    ap_uint<12> mergedBarrelStubsMap;
+    ap_uint<12> mergedDiskStubsMap;
+    ap_uint<24> totalHitMap;
```
aehart, line 45: Please replace the magic numbers in the types with named constants. In this case, I think they can be calculated from `TrackFit<NBarrelStub, NDiskStub>::kTFHitCountSize`, `NBarrelStub`, and `NDiskStub`.

TrackletAlgorithm/TrackHandler.cc
```diff
@@ -0,0 +1,86 @@
...
+    } else {
+      matchesFoundBarrel[layerIndex][0] = 0;
+    }
+    
+    if ((masterDiskStubIndex == inputDiskStubIndex) && (masterDiskStubIndex > 0)){
+      matchesFoundDisk[layerIndex][0] = 1;
+    } else {
+      matchesFoundDisk[layerIndex][0] = 0;
+    }
+  }
```
aehart, line 25: To be as general as possible, I think this should be broken up into two loops, one for the barrel stubs with an upper limit of `NBarrelStub`, and one for the disk stubs with an upper limit of `NDiskStub`.

This comment also applies to other loops in this file.

TrackletAlgorithm/TrackHandler.cc
```diff
@@ -0,0 +1,86 @@
...
+  #pragma HLS inline
+  // update master track
+  // check whether the stub word is non-zero in the compared track
+  // then add stub into master track if share > 3 stubs in common
+  LOOP_SetInputStubToMaster:
+  for(unsigned int layerIndex = 0; layerIndex < 4; layerIndex++){
+    #pragma HLS unroll
+    if((masterTrk._barrelStubArray[layerIndex][0] == 0) && (trk._barrelStubArray[layerIndex][0] != 0)){
+      masterTrk._barrelStubArray[layerIndex][0] = trk._barrelStubArray[layerIndex][0];
+      mergedBarrelStubsMap += (1 << 3*layerIndex);
```
aehart, line 53: Please replace the `3` with `TrackFit<NBarrelStub, NDiskStub>::kTFHitCountSize` (or a constant derived from this).

TrackletAlgorithm/TrackHandler.cc
```diff
@@ -0,0 +1,86 @@
...
+    }
+    if((masterTrk._diskStubArray[layerIndex][0] == 0) && (trk._diskStubArray[layerIndex][0] !=0 )){
+      masterTrk._diskStubArray[layerIndex][0] = trk._diskStubArray[layerIndex][0];
+      mergedDiskStubsMap += (1 << 3*layerIndex);
```
aehart, line 57: Here as well, replace `3` with `TrackFit<NBarrelStub, NDiskStub>::kTFHitCountSize`.

TrackletAlgorithm/TrackMerger.h
```diff
@@ -0,0 +1,99 @@
...
+#define TrackletAlgorithm_TrackMerger_h
+
+#include "hls_stream.h"
+#include <cassert>
+#include "TrackFitMemory.h"
+#include "TrackHandler.h"
+
+const unsigned int kNComparisonModules = 16;
+const unsigned int kNBuffers = kNComparisonModules + 1;
+const unsigned int kMaxTrack = 50;
```
aehart, line 11: Maybe I missed it, but where did 50 come from?

TrackletAlgorithm/TrackMerger.cc
```diff
@@ -0,0 +1,176 @@
...
+  hls::stream<TrackFit<NBarrelStub, NDiskStub>::BarrelStubWord> &barrelStubWords_1_o, 
+  hls::stream<TrackFit<NBarrelStub, NDiskStub>::BarrelStubWord> &barrelStubWords_2_o, 
+  hls::stream<TrackFit<NBarrelStub, NDiskStub>::BarrelStubWord> &barrelStubWords_3_o, 
+  hls::stream<TrackFit<NBarrelStub, NDiskStub>::DiskStubWord> &diskStubWords_0_o, 
+  hls::stream<TrackFit<NBarrelStub, NDiskStub>::DiskStubWord> &diskStubWords_1_o, 
+  hls::stream<TrackFit<NBarrelStub, NDiskStub>::DiskStubWord> &diskStubWords_2_o, 
+  hls::stream<TrackFit<NBarrelStub, NDiskStub>::DiskStubWord> &diskStubWords_3_o 
+  )
+{
+  #pragma HLS inline
```
aehart, line 113: Would it make sense to use the `recursive` option here? It looks like you're inlining a lot of your subfunctions, and this would allow you to just have one instance of this pragma.

If the `recursive` option makes sense, maybe you could even move it to `TrackMergerTop` in TrackMergerTop.cc.

TrackletAlgorithm/TrackMerger.cc
```diff
@@ -0,0 +1,176 @@
...
+  static ComparisonModule comparisonModule[kNComparisonModules];
+  #pragma HLS array_partition variable=comparisonModule complete dim=0
+
+  // TrackStruct unmergedTracks[kMaxTrack];
+  // #pragma HLS array_partition variable=unmergedTracks complete dim=0
+  LOOP_Input:
+  for (unsigned int i = 0; i < kMaxTrack; i++){ 
+    #pragma HLS pipeline II=1
```
aehart, line 121: Have you tried adding the `rewind` option here? It will be necessary eventually so that the pipeline doesn't have to drain before starting to process the next event.

TrackletAlgorithm/TrackMerger.cc
```diff
@@ -0,0 +1,176 @@
...
+                barrelStubWords_0_o,
+                barrelStubWords_1_o,
+                barrelStubWords_2_o,
+                barrelStubWords_3_o,
+                diskStubWords_0_o,
+                diskStubWords_1_o,
+                diskStubWords_2_o,
+                diskStubWords_3_o);
+
+  }
```
aehart, line 172: Can you explain what this loop is doing?

aehart:
@kwalkingshaw Thank you for the well-written code. In addition to the inline comments, I had just a few general ones:
- I was told to not use `hls::stream` because that header file is not available in CMSSW, which will cause problems when this code is ultimately integrated into the HLS-based emulation. That's why I used the `stream` pragma instead in the TrackBuilder. I'm fine with this conversion happening in a later PR, but if you want to do it in this one, all the better.
- For consistency, please move TrackMergerTop.{h,cc} to the TopFunctions/ directory.
- Can you please add the TM to .gitlab-ci.yml, so CI tests are run automatically? You just need to add a `quality-check` and `vivado-hls-build` test, which you can mostly copy and paste from, e.g., the TB. For the `vivado-hls-build` test, you can add `allow_failure: true` for now.

project/settings_hls.tcl
kwalkingshaw: I just wanted to try with a different FPGA to see if the resource usage was a bit better, but it was about the same as a VU7P.

TestBenches/TrackMerger_test.cpp
kwalkingshaw: Yes I commented it out as it currently fails with the current set of test vectors, by uncommenting I can force the CSIM to pass to be able to look at the resource usage etc.

TrackletAlgorithm/TrackHandler.h
kwalkingshaw: Sure, will implement those!

TrackletAlgorithm/TrackMerger.cc
kwalkingshaw: This loop is sending out the master track of each ComparisonModule after it has finished processing tracks.

kwalkingshaw:
Thanks for the comments Andrew, will make a commit as soon as I am done. Regarding using the stream class, I am working on another version that is very similar to this code but uses arrays with the stream pragma instead of the stream class.

TestBenches/TrackMerger_test.cpp
aehart: OK, I think this makes sense for now, given the state of the test vectors.

TrackletAlgorithm/TrackMerger.cc
aehart: OK, I think I get what it's doing now. Thanks.

TrackletAlgorithm/TrackMerger.h
aehart: OK that seems reasonable. Thanks.

aehart (approved):
I think this looks good, and I'll approve it.

Followup is needed to address the rewind/pipelining issue and the use of hls::stream, but these can be done in separate PRs.

PR #247, opened 2022-10-26
URL fix

aehart:
This PR updates the URLs in emData/download.sh so they are compatible with the updated CERNBox.

The files pointed to by the URLs are unchanged.


tomalin (approved):
Thanks @aehart for the rapid fix. I've checked this now works.

PR #249, opened 2022-11-07
CompareMemWithFile Fix for FutureSW CMSSW

carriganm95:
This change edits the FileReadUtility.h function compareMemWithFile to remove the range function from the reference and computed bitstrings. In CMSSW the range function has been seen to cause odd behavior and leads to memories appearing blank during memory checking. This is solved by using an ifdef CMSSW_GIT_HASH statement to remove the range when working in CMSSW.

### DIFF
--- a/TestBenches/FileReadUtility.h
+++ b/TestBenches/FileReadUtility.h
@@ -142,13 +142,10 @@ unsigned int compareMemWithFile(const MemType& memory, std::ifstream& fout,
   constexpr int msb = (LSB >= 0 && MSB >= LSB) ? MSB : MemType::getWidth() - 1;
 
   for (unsigned int i = 0; i < memory_ref.getDepth(); ++i) {
-#ifdef CMSSW_GIT_HASH
-    auto data_ref = memory_ref.read_mem(ievt,i).raw();
-    auto data_com = memory.read_mem(ievt,i).raw();
-#else
-    auto data_ref = memory_ref.read_mem(ievt,i).raw().range(msb,lsb);
-    auto data_com = memory.read_mem(ievt,i).raw().range(msb,lsb);
-#endif
+    auto data_ref_raw = memory_ref.read_mem(ievt,i).raw();
+    auto data_com_raw = memory.read_mem(ievt,i).raw();
+    auto data_ref = data_ref_raw.range(msb,lsb);
+    auto data_com = data_com_raw.range(msb,lsb);
     if (i==0) {
       // If both reference and computed memories are completely empty, skip it
       if (data_com == 0 && data_ref == 0) break;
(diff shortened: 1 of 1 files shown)


tomalin:
@carriganm95 from a quick test, replacing the dodgy lines instead by the following, which work both for HLS & CMSSW, so require no pragma, is also an option.

    auto data_ref_raw = memory_ref.read_mem(ievt,i).raw();
    auto data_com_raw = memory.read_mem(ievt,i).raw();
    auto data_ref = data_ref_raw.range(msb,lsb);
    auto data_com = data_com_raw.range(msb,lsb);

carriganm95:
> @carriganm95 from a quick test, replacing the dodgy lines instead by the following, which work both for HLS & CMSSW, so require no pragma, is also an option.
> 
> ```
> auto data_ref_raw = memory_ref.read_mem(ievt,i).raw();
> auto data_com_raw = memory.read_mem(ievt,i).raw();
> auto data_ref = data_ref_raw.range(msb,lsb);
> auto data_com = data_com_raw.range(msb,lsb);
> ```

tomalin:
> > @carriganm95 from a quick test, replacing the dodgy lines instead by the following, which work both for HLS & CMSSW, so require no pragma, is also an option.
> > ```
> > auto data_ref_raw = memory_ref.read_mem(ievt,i).raw();
> > auto data_com_raw = memory.read_mem(ievt,i).raw();
> > auto data_ref = data_ref_raw.range(msb,lsb);
> > auto data_com = data_com_raw.range(msb,lsb);
> > ```
> 
> If this is preferred to the ifdef I can change the PR. Just let me know

Hi @carriganm95, if you confirm my proposal also fixes the problem, I think it is prettier. (1) Avoiding pragmas simplifies the code. (2) If our FW uses the range function and CMSSW does not, then the danger exists that they may sometimes given different results (and it will confuse people reading the code).

carriganm95:
> > > @carriganm95 from a quick test, replacing the dodgy lines instead by the following, which work both for HLS & CMSSW, so require no pragma, is also an option.
> > > ```
> > > auto data_ref_raw = memory_ref.read_mem(ievt,i).raw();
> > > auto data_com_raw = memory.read_mem(ievt,i).raw();
> > > auto data_ref = data_ref_raw.range(msb,lsb);
> > > auto data_com = data_com_raw.range(msb,lsb);
> > > ```
> > 
> > 
> > If this is preferred to the ifdef I can change the PR. Just let me know
> 
> Hi @carriganm95, if you confirm my proposal also fixes the problem, I think it is prettier. (1) Avoiding pragmas simplifies the code. (2) If our FW uses the range function and CMSSW does not, then the danger exists that they may sometimes given different results (and it will confuse people reading the code).

@tomalin this fix worked for me so I have committed your suggestion.

tomalin (approved):
Approved, (though shame we don't understand why the fix helps).

PR #250, opened 2022-11-08
Makefiles

aehart:
This PR replaces the subchain-specific compileHLS.tcl with a Makefile. This allows the HLS modules to be C-synthesized and exported as IPs using multiple CPU cores, as illustrated in the README:
https://github.com/cms-L1TK/firmware-hls/blob/61cfe2e9a9219eb6669b38b666a6aa5dea787dcb/README.md?plain=1#L22-L25

The synthesis and implementation can also be run using the Makefile:
https://github.com/cms-L1TK/firmware-hls/blob/61cfe2e9a9219eb6669b38b666a6aa5dea787dcb/README.md?plain=1#L29-L30


.gitlab-ci.yml
```diff
@@ -65,8 +65,7 @@ stages: # ---------------------------------------------------------------------
   script:
     - cd IntegrationTests/${PROJ_NAME}/script
     - pwd; ls -la; #debug
-    - vivado_hls -f ./compileHLS.tcl
-    - vivado -mode batch -source ./makeProject.tcl | awk 'BEGIN{IGNORECASE=1} /^error/ {exit_code=1;} // {print $0;} END{exit exit_code}'
+    - make -j4 Work
```
tomalin, line 68: The CMSSW scram command lets one specify "-j4". But it also lets one specify "-j" meaning "use all available cores". Can you provide this option? On linux, the number of available cores can be determined with the "nproc" command.

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
```diff
@@ -0,0 +1,511 @@
...
+
+MemPrints:
+	ln -sf $(EMDATA)/MemPrintsBarrel MemPrints
+
+LUTs:
+	ln -sf $(EMDATA)/LUTsBarrel LUTs
+
+### Make HLS IPs ###
+
+InputRouterTop_IR_DTC_2S_1_A: $(TOP_FUNCS)/InputRouter_parameters.h $(TOP_FUNCS)/InputRouterTop.cc $(TOP_FUNCS)/InputRouterTop.h $(TRK_ALGO)/InputRouter.h
```
tomalin, line 36: Does the repetitive code in this Makefile, specifying how to compile each specific processing module, such as InputRouterTop_IR_DTC_2S_1_B, have to be written by hand? This will be painful to maintain, and means that we'll need several copies of this Makefile for different chains. Is there no way for the Makefile to determine this info itself, or for these parts of the Makefile to be written by a python script or similar?

.gitlab-ci.yml
aehart: According to the man page for scram and gmake, "-j" without an argument does not limit the number of processes that run simultaneously. So it's essentially equivalent to something like "-j999".

But I think we want to limit it to the number of CPU cores, so I now use "nproc" as you recommend.

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
aehart: I'm looking into automatically generating prerequisites using "gcc", and it is definitely possible. I just need to do further tests and decide on the best way to organize things.

tomalin:
Instruction 4 in the "running chains" section of https://github.com/cms-L1TK/firmware-hls/blob/makefile/README.md is missing a "../" in "../../common/script/CompareMemPrintsFW.py".

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
pwittich: A cursory glance suggests that the RHS of the dependency is the same on every line. If that is true, you could just create one set of dependencies and use that to reduce the copypasta.

i.e.,
```
IRDEPS=  $(TOP_FUNCS)/InputRouter_parameters.h \
                 $(TOP_FUNCS)/InputRouterTop.cc \                 
                 $(TOP_FUNCS)/InputRouterTop.h \
                 $(TRK_ALGO)/InputRouter.h
IputRouterTop_IR_DTC_2S_1_A: $(IRDEPS)
InputRouterTop_IR_DTC_PS10G_1_A: $(IRDEPS)
```
etc etc.

pwittich:
To me this looks pretty good, I'd be happy to merge it and improve it going forward.

.gitlab-ci.yml
pwittich: is this resolved, @tomalin ?

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
pwittich: Also, the argument to GCC to generate the dependencies is `-MD`; this spits out a .d file which you can then include into your makefile with something like
```
#
# Include the automatically generated dependency files.
#
ifneq (${MAKECMDGOALS},clean)
-include ${wildcard ${COMPILER}/*.d} __dummy__
endif
```
here ${COMPILER} is the directory that the .d and .o files are stored.

pwittich:
FWIW I just tried this and following the instructions literally under 'running chains' didn't work for me from a freshly checked-out copy of the repo.
```
(base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >make -j `nproc` Work
ln -sf ../../../common common
ln -sf ../../../../emData/MemPrintsReduced MemPrints
ln -sf ../../../../emData/LUTsReduced LUTs
make: *** No rule to make target `../../../../TopFunctions/ReducedConfig/InputRouter_parameters.h', needed by `InputRouterTop_IR_DTC_PS10G_1_A'.  Stop.
make: *** Waiting for unfinished jobs....
(base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >which vivado_hls
/nfs/opt/Xilinx/Vivado/2018.2/bin/vivado_hls
(base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >ls -l ../../../../TopFunctions/ReducedConfig/
total 0
lrwxrwxrwx 1 pw94 cms 23 Jan  4 14:38 TrackletEngineTop.cc -> ../TrackletEngineTop.cc
lrwxrwxrwx 1 pw94 cms 22 Jan  4 14:38 TrackletEngineTop.h -> ../TrackletEngineTop.h
(base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >
```

There is also a mild issue that the make command listed in the readme is a bit confusing; I'll fix it in the PR.

pwittich:
> Instruction 4 in the "running chains" section of https://github.com/cms-L1TK/firmware-hls/blob/makefile/README.md is missing a "../" in "../../common/script/CompareMemPrintsFW.py".

I tried to fix this in the commit to the README

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
aehart: Yes, I have learned a lot from the gmake manual over the past few days, and I think utilizing these automatically generated dependency files is the way to go.

In fact, I'm almost at the point where there are ~2 lines that are unique to the Makefile for each subchain, with all the actual rules being common. I just need a little more tinkering to get everything working.

aehart:
> FWIW I just tried this and following the instructions literally under 'running chains' didn't work for me from a freshly checked-out copy of the repo.
> 
> ```
> (base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >make -j `nproc` Work
> ln -sf ../../../common common
> ln -sf ../../../../emData/MemPrintsReduced MemPrints
> ln -sf ../../../../emData/LUTsReduced LUTs
> make: *** No rule to make target `../../../../TopFunctions/ReducedConfig/InputRouter_parameters.h', needed by `InputRouterTop_IR_DTC_PS10G_1_A'.  Stop.
> make: *** Waiting for unfinished jobs....
> (base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >which vivado_hls
> /nfs/opt/Xilinx/Vivado/2018.2/bin/vivado_hls
> (base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >ls -l ../../../../TopFunctions/ReducedConfig/
> total 0
> lrwxrwxrwx 1 pw94 cms 23 Jan  4 14:38 TrackletEngineTop.cc -> ../TrackletEngineTop.cc
> lrwxrwxrwx 1 pw94 cms 22 Jan  4 14:38 TrackletEngineTop.h -> ../TrackletEngineTop.h
> (base) lnx4108:/mnt/scratch1/wittich/2018.2/firmware-hls/IntegrationTests/ReducedConfig/IRtoTB/script >
> ```
> 
> There is also a mild issue that the make command listed in the readme is a bit confusing; I'll fix it in the PR.

I think the issue is that InputRouter_parameters.h doesn't exist until emData/download.sh is run. In both my local testing and in the CI tests, that script is already run before make is.

I'll try to fix this in the next version I push.

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
pwittich: Let me know if you need a beta tester.

aehart: @tomalin @pwittich I'm pretty sure I damaged my brain in the process, but I have greatly simplified the Makefiles, with all of the dependencies determined automatically. There are now just two lines that need to be updated when copying a Makefile to a new chain: the location of the top functions and the list of modules (top functions) in the chain.

Everything seems to work beautifully in my testing, and we'll see if the CI tests remain happy, but please feel free to try to break things yourself.

README.md
```diff
@@ -21,14 +21,13 @@ This would create a project directory \<project> ("projrouter" in case of the ab
 
 ## Running chains (illustrated for IR to TB skinny chain)
 
-1) cd IntegrationTests/ReducedConfig/IRtoTB/script/
-2) vivado_hls -f compileHLS.tcl (makes HLS IP cores).
-3) vivado -mode batch -source makeProject.tcl (creates Vivado project).
-4) vivado -mode batch -source runSim.tcl (runs Vivado simulation,
+1) `cd IntegrationTests/ReducedConfig/IRtoTB/script/`
+2) `make -j $(nproc) Work` (makes HLS IP cores; run as many jobs as you have CPU cores).
```
aehart, line 25: @pwittich Are the backticks needed? I'm not sure I understand their purpose here.

pwittich: backtick renders it as inline code in markdown

aehart: Now that makes sense, as does the previous confusion with the backticks that I added.

Any reason to not give the other lines of commands the same treatment?

pwittich: Go for it.

pwittich: I couldn't figure out how to render actual backticks so I went with the $() construction instead. ¯\_(ツ)_/¯

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
pwittich: I was able to run it successfully; it might be worth in the future exploring if we can make the download script (which now does a lot) faster too (i.e., parallelize parts of it...), maybe via `xargs` or something.

README.md
aehart: > Go for it.

Done.

> I couldn't figure out how to render actual backticks so I went with the $() construction instead. ¯_(ツ)_/¯

\\\` (one backslash, one backtick) seems to work, but $() is works just as well.

IntegrationTests/BarrelConfig/IRtoTB/script/Makefile
aehart: Thanks for testing. I've made a few additional tweaks so that hopefully `make` plays nicely with the CI tests now. We'll see if my understanding was correct later this evening.

And agreed about making download.sh faster. I've had really good results using `xargs` to parallelize scripts in other projects.

IntegrationTests/common/script/Makefile.mk
```diff
@@ -77,7 +77,8 @@ $(DEPS)/functions/%.d: $(EMDATA)/MemPrints
 	 TOP_FUNC=`echo $@ | sed 's,.*\/\([^/]*\)\.d,\1,g'`; \
 	 TOP_FILE=`grep -l $${TOP_FUNC} $(TOP_FUNCS)/*.cc | sed 's,.*\/\([^/]*\)\.cc,\1,g'`; \
 	 echo "$${TOP_FUNC}: $(DEPS)/files/$${TOP_FILE}.d $(EMDATA)/MemPrints" > $@; \
-	 echo "	vivado_hls -f $(COMPILE_HLS) $(TOP_FUNCS)/$${TOP_FILE}.cc $${TOP_FUNC}" >> $@; \
+	 echo "	@rm -rfv $${TOP_FUNC}; \\" >> $@; \
+	 echo "	 vivado_hls -f $(COMPILE_HLS) $(TOP_FUNCS)/$${TOP_FILE}.cc $${TOP_FUNC}" >> $@; \
```
pwittich, line 81: Not to nit-pick, but shouldn't this be a rule in the makefile for the appropriate file endings rather than generating a rule  for each dependency explicitly?  something like 
```
#
# The rule for building the object file from each C source file.
#
${TOP_FUNCS}/${SUFFIX}/%.o: %.cc
	vivado_hls -f $(COMPILE_HLS) $(TOP_FUNCS)/$${TOP_FILE}.cc $${TOP_FUNC}

```

aehart: The issue is determining the file that a given top function is contained in. Either this mapping between function and file has to be hard-coded in the Makefile itself, or it has to be determined dynamically, which is what is done now with `grep` (N.B., TOP_FUNC and TOP_FILE are shell variables of the recipe, not Makefile variables).

However, I did just realize that the two pattern rules in Makefile.mk can be collapsed into just one, which I think is easier to understand.

pwittich (approved):
nice work

PR #252, opened 2023-01-02
Update to VU13P

aehart:
This PR updates the default part number to a VU13P.

It also moves the part number to a single file, to be used by both HLS and Vivado, for consistency.

### DIFF
--- a/IntegrationTests/BarrelConfig/IRtoTB/script/makeProject.tcl
+++ b/IntegrationTests/BarrelConfig/IRtoTB/script/makeProject.tcl
@@ -3,7 +3,7 @@
 
 # Create project
 set projName "Work"
-set FPGA "xcvu7p-flvb2104-1-e"
+source common/script/fpga.tcl
 create_project -force ${projName} ./${projName} -part $FPGA
 set_property target_language VHDL [current_project]
 
--- a/IntegrationTests/IRVMR/script/makeProject.tcl
+++ b/IntegrationTests/IRVMR/script/makeProject.tcl
@@ -3,7 +3,7 @@
 
 # Create project
 set projName "Work"
-set FPGA "xcvu7p-flvb2104-1-e"
+source common/script/fpga.tcl
 create_project -force ${projName} ./${projName} -part $FPGA
 set_property target_language VHDL [current_project]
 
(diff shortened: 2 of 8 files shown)


aryd (commented):
According to Charlie Strohman we are going to use the part: xcvu13p-flga2577-2-e. Any reason not to use the actually part we plan to have on the boards?

aehart:
> According to Charlie Strohman we are going to use the part: xcvu13p-flga2577-2-e. Any reason not to use the actually part we plan to have on the boards?

I did add that as a commented-out line, so that it can be used for testing, but I left the -1 part as the default.

Maybe @pwittich can comment on which speed grade should be the default for now?

tomalin:
Given we have timing problems, and purchasing speed 2 grade for production is being seriously considered (unless @pwittich contradicts that?), I'd recommend making speed 2 the default. If in future, we succeed in improving the FW to the point where speed 1 runs at 240 MHz, we can change the default and offer to save CMS some money.

tomalin:
This PR looks good to me, except that it seems ugly that project/settings_hls.tcl is dependent on IntegrationTests/common/script/fpga.tcl . i.e. To compile any individual HLS module, we now need code from the IntegrationTests/ directory. I'd suggest moving fpga.tcl to project/.

aehart:
> Given we have timing problems, and purchasing speed 2 grade for production is being seriously considered (unless @pwittich contradicts that?), I'd recommend making speed 2 the default. If in future, we succeed in improving the FW to the point where speed 1 runs at 240 MHz, we can change the default and offer to save CMS some money.

Makes sense to me, and I've changed the default to -2.

> This PR looks good to me, except that it seems ugly that project/settings_hls.tcl is dependent on IntegrationTests/common/script/fpga.tcl . i.e. To compile any individual HLS module, we now need code from the IntegrationTests/ directory. I'd suggest moving fpga.tcl to project/.

Fair point, and I've moved it to project/.

pwittich:
> > According to Charlie Strohman we are going to use the part: xcvu13p-flga2577-2-e. Any reason not to use the actually part we plan to have on the boards?
> 
> I did add that as a commented-out line, so that it can be used for testing, but I left the -1 part as the default.
> 
> Maybe @pwittich can comment on which speed grade should be the default for now?

It makes sense to have the -2 speed grade as the default.

pwittich (approved):
Looks good to me

PR #254, opened 2023-01-12
Trackmerger nostream

kwalkingshaw:
Updated version of track merger to use arrays for trackWord, barrelStubWords, and diskStubWords. This branch no longer uses the hls::stream class.

### DIFF
--- a/TestBenches/TrackMerger_test.cpp
+++ b/TestBenches/TrackMerger_test.cpp
@@ -39,7 +39,7 @@ int main(){
 
   auto &fout_outputTracks = tb.files("old_CleanTrack_CT_L1L2*"); // use CleanTrack_CT_L1L2_04.dat when tracks have been merged or if no merge make a copy e.g. output_TrackFit_TF_L1L2*
   // Loop over events
-  for (unsigned int ievt = 0; ievt < nevents; ++ievt) { //nevents
+  for (unsigned int ievt = 0; ievt < nevents; ++ievt) { 
     cout << "Event: " << dec << ievt << endl;
     
     for (unsigned short i = 0; i < kMaxTrack; ++i){
--- a/TestBenches/TrackMerger_test.cpp
+++ b/TestBenches/TrackMerger_test.cpp
@@ -30,14 +30,14 @@ int main(){
   TrackFitType::DiskStubWord diskStubWords_o[kMaxTrack][4];
   TrackFitMemory<NBarrelStub, NDiskStub> outputTracks;
   // int outputNumber;
-
+ 
   TBHelper tb("../../../../../emData/PD/PD/");
 
   // Open input files
-  auto &fin_inputTracks = tb.files("old_TrackFit_TF_L1L2*");
+  auto &fin_inputTracks = tb.files("TrackFit_TF_L1L2*");
 
 
-  auto &fout_outputTracks = tb.files("old_CleanTrack_CT_L1L2*"); // use CleanTrack_CT_L1L2_04.dat when tracks have been merged or if no merge make a copy e.g. output_TrackFit_TF_L1L2*
+  auto &fout_outputTracks = tb.files("CleanTrack_CT_L1L2*"); // use CleanTrack_CT_L1L2_04.dat when tracks have been merged or if no merge make a copy e.g. output_TrackFit_TF_L1L2*
   // Loop over events
   for (unsigned int ievt = 0; ievt < nevents; ++ievt) { 
     cout << "Event: " << dec << ievt << endl;
@@ -111,10 +111,12 @@ int main(){
       
       outputTracks.write_mem(bx, track, nTracks );
       ++nTracks;
+      
     } 
 
     // Comparing outputs
     err_count += compareMemWithFile<TrackFitMemory<NBarrelStub, NDiskStub>>(outputTracks, fout_outputTracks.at(0), ievt, "Tracks", truncation);
+    outputTracks.clear();
     
   }
 
--- a/TestBenches/TrackMerger_test.cpp
+++ b/TestBenches/TrackMerger_test.cpp
@@ -37,8 +37,7 @@ int main(){
   auto &fin_inputTracks = tb.files("TrackFit_TF_L1L2*");
 
 
-  auto &fout_outputTracks = tb.files("CleanTrack_CT_L1L2*"); // use CleanTrack_CT_L1L2_04.dat when tracks have been merged or if no merge make a copy e.g. output_TrackFit_TF_L1L2*
-  // Loop over events
+  auto &fout_outputTracks = tb.files("CleanTrack_CT_L1L2*"); 
   for (unsigned int ievt = 0; ievt < nevents; ++ievt) { 
     cout << "Event: " << dec << ievt << endl;
     
@@ -116,10 +115,9 @@ int main(){
 
     // Comparing outputs
     err_count += compareMemWithFile<TrackFitMemory<NBarrelStub, NDiskStub>>(outputTracks, fout_outputTracks.at(0), ievt, "Tracks", truncation);
-    outputTracks.clear();
     
   }
-
+  // Commented as f/w does not match the s/w - implementations are different so test vectors are different, TB fails.
   // Handling case of err%256 == 0 
   // if (err_count > 255) err_count = 255;
   // return err_count;
--- a/TestBenches/TrackMerger_test.cpp
+++ b/TestBenches/TrackMerger_test.cpp
@@ -67,13 +67,11 @@ int main(){
     for(unsigned short i = 0; i < kMaxTrack; i++){
       TrackFitType track;
       if (i < inputTracks.getEntries(bx)) {
-      track = inputTracks.read_mem(bx, i+1); //i+1 for merge
+      track = inputTracks.read_mem(bx, i);
       } else {
         track = TrackFitType();
       } 
-      // std::cout << "tbIn TrkWrd: " << std::hex << track.getTrackWord() << std::endl;
       trackWord[i] = track.getTrackWord();
-      // std::cout << "tbTrkSet: " << std::hex << trackWord[i] << std::endl;
       barrelStubWords[i][0] = track.getBarrelStubWord<0>();
       barrelStubWords[i][1] = track.getBarrelStubWord<1>();
       barrelStubWords[i][2] = track.getBarrelStubWord<2>();
--- a/TestBenches/TrackMerger_test.cpp
+++ b/TestBenches/TrackMerger_test.cpp
@@ -13,35 +13,22 @@
 const int nevents = 100;
 
 using namespace std;
-using namespace hls;
 
 int main(){
   //Error counter
   int err_count = 0;
     
   // Input memories
-  stream<TrackFitType::TrackWord> trackWord("input_stream_1");
-  static stream<TrackFitType::BarrelStubWord> barrelStubWords_0("input_stream_2");
-  static stream<TrackFitType::BarrelStubWord> barrelStubWords_1("input_stream_3");
-  static stream<TrackFitType::BarrelStubWord> barrelStubWords_2("input_stream_4");
-  static stream<TrackFitType::BarrelStubWord> barrelStubWords_3("input_stream_5");
-  static stream<TrackFitType::DiskStubWord> diskStubWords_0("input_stream_6");
-  static stream<TrackFitType::DiskStubWord> diskStubWords_1("input_stream_7");
-  static stream<TrackFitType::DiskStubWord> diskStubWords_2("input_stream_8");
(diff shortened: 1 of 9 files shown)


project/settings_hls.tcl
```diff
@@ -12,17 +12,17 @@ config_compile -name_max_length 100
 # Set clock frequency
 create_clock -period 250MHz -name default
 
-# Set FPGA variable with part number
```
pwittich, line 15: maybe this is just WIP but please don't commit this change as we want to move to the VU13P ...

kwalkingshaw:
- Have kept a single loop over layers and not split into barrel and disk stubs since there will always be 4 layers of barrel stubs and 4 layers of disk stubs
- Have not used dataflow as .process() function contains many conditionals which is unsuitable for a dataflow region
- Used data_pack pragma for arrays of structs for easier handling of info
- Now meets timing post-implementation using a VU13P FPGA

TrackletAlgorithm/TrackHandler.cc
```diff
@@ -1,82 +1,81 @@
...
-
-    if((masterBarrelStubIndex == inputBarrelStubIndex) && (masterBarrelStubIndex > 0)){
-      matchesFoundBarrel[barrelStubNum][0] = 1;
-    } else {
-      matchesFoundBarrel[barrelStubNum][0] = 0;
-    }
+    ap_uint<TrackFitType::kTFStubIndexSize> masterBarrelStubIndex = ap_uint<TrackFitType::kTFStubIndexSize>(masterTrk._barrelStubArray[barrelStubNum][0].range(kBarrelStubIndexSizeMSB, kBarrelStubIndexSizeLSB));
+    ap_uint<TrackFitType::kTFStubIndexSize> inputBarrelStubIndex = ap_uint<TrackFitType::kTFStubIndexSize>(trk._barrelStubArray[barrelStubNum][0].range(kBarrelStubIndexSizeMSB, kBarrelStubIndexSizeLSB));
+    
+    ((masterBarrelStubIndex == inputBarrelStubIndex) && (masterBarrelStubIndex > 0)) ? matchesFoundBarrel[barrelStubNum][0] = 1 : matchesFoundBarrel[barrelStubNum][0] = 0;
```
aehart, line 17: This can be written slightly more compactly, with less duplication:
```suggestion
    matchesFoundBarrel[barrelStubNum][0] = ((masterBarrelStubIndex == inputBarrelStubIndex) && (masterBarrelStubIndex > 0)) ?  1 : 0;
```

TrackletAlgorithm/TrackHandler.cc
```diff
@@ -1,82 +1,81 @@
...
-    if ((masterDiskStubIndex == inputDiskStubIndex) && (masterDiskStubIndex > 0)){
-      matchesFoundDisk[diskStubNum][0] = 1;
-    } else {
-      matchesFoundDisk[diskStubNum][0] = 0;
-    }
+    #pragma HLS unroll
+    ap_uint<TrackFitType::kTFStubIndexSize> masterDiskStubIndex = ap_uint<TrackFitType::kTFStubIndexSize>(masterTrk._diskStubArray[diskStubNum][0].range(kDiskStubIndexSizeMSB, kDiskStubIndexSizeLSB));
+    ap_uint<TrackFitType::kTFStubIndexSize> inputDiskStubIndex = ap_uint<TrackFitType::kTFStubIndexSize>(trk._diskStubArray[diskStubNum][0].range(kDiskStubIndexSizeMSB, kDiskStubIndexSizeLSB));
+
+    ((masterDiskStubIndex == inputDiskStubIndex) && (masterDiskStubIndex > 0)) ? matchesFoundDisk[diskStubNum][0] = 1 : matchesFoundDisk[diskStubNum][0] = 0;
```
aehart, line 25: Likewise, this can be more compact:
```suggestion
    matchesFoundDisk[diskStubNum][0] = ((masterDiskStubIndex == inputDiskStubIndex) && (masterDiskStubIndex > 0)) ? 1 : 0;
```

TrackletAlgorithm/TrackHandler.cc
```diff
@@ -1,82 +1,81 @@
...
     matchesFound += matchesFoundBarrel[layerIndex][0] + matchesFoundDisk[layerIndex][0];
   }
-  if (matchesFound >= mergeCondition){
-    matchFound = 1;
-    return matchFound; 
-    } else {
-      matchFound = 0;
-      return matchFound;
-    }
+  (matchesFound >= mergeCondition) ? matchFound = 1 : matchFound = 0;
```
aehart, line 36: Here as well:
```suggestion
  matchFound = (matchesFound >= mergeCondition) ? 1 : 0; 
```

TrackletAlgorithm/TrackHandler.cc
```diff
@@ -1,82 +1,81 @@
...
+  // Update master track
+  // Check whether the stub word is non-zero in the compared track
+  // Then add stub into master track if share > 3 stubs in common
   LOOP_SetInputBarrelStubToMaster:
   for(unsigned int barrelStubNum = 0; barrelStubNum < NBarrelStub; barrelStubNum++){
     #pragma HLS unroll
     if((masterTrk._barrelStubArray[barrelStubNum][0] == 0) && (trk._barrelStubArray[barrelStubNum][0] != 0)){
       masterTrk._barrelStubArray[barrelStubNum][0] = trk._barrelStubArray[barrelStubNum][0];
-      mergedBarrelStubsMap += (1 << TrackFitType::kTFHitCountSize*barrelStubNum);
+      mergedBarrelStubsMap | (1 << TrackFitType::kTFHitCountSize*barrelStubNum);
```
aehart, line 54: @kwalkingshaw This line and the corresponding one for `mergedDiskStubsMap` don't actually do anything. They calculate a bitwise OR, but it isn't stored anywhere…

TrackletAlgorithm/TrackMerger.cc
```diff
@@ -6,167 +6,162 @@ TrackStruct& ComparisonModule::getMasterTrackStruct(){
...
-  #pragma HLS array_partition variable=comparisonModule complete dim=0
-
-  // TrackStruct unmergedTracks[kMaxTrack];
-  // #pragma HLS array_partition variable=unmergedTracks complete dim=0
+  #pragma HLS array_partition variable=comparisonModule dim=1
+
+  static TrackStruct unmergedTracks[kNLastTracks];
+  #pragma HLS data_pack variable=unmergedTracks
+
+  static TrackStruct cmTracks[kNComparisonModules];
```
aehart, line 98: @kwalkingshaw Do these actually need to be `static`? Does removing the `static` keyword change the results of the C-simulation or C-synthesis?

project/settings_hls.tcl
aehart: @kwalkingshaw Can you revert this file to what's on the master branch? The FPGA part number should now be set only in project/fpga.tcl.

aehart:
@kwalkingshaw Sorry for not noticing this sooner, but if you check the results of the CI here:
https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/jobs/27301147
you see that the C-simulation fails with an exception from `std::vector`:
```
…
Event: 0
terminate called after throwing an instance of 'std::out_of_range'
  what():  vector::_M_range_check: __n (which is 0) >= this->size() (which is 0)
@E Simulation failed.
ERROR: [SIM 211-100] CSim failed with errors.
INFO: [SIM 211-3] *************** CSIM finish ***************
…
```

Can you debug this before we merge?

kwalkingshaw:
> 
Hi Andrew, it's because I am using the old test vectors and not the current ones to debug the module - I had to copy the old files across so should I just set the test bench to a false pass?

TrackletAlgorithm/TrackMerger.cc
kwalkingshaw: Removing the static doesn't update the arrays properly (even in simulation) and the static keyword also reduces the latency as extra loops appeared in the synthesis report which had a latency of the size of the array, not sure if this is an HLS issue?

aehart: The constructor for `ComparisonModule` initializes members of the class:
https://github.com/cms-L1TK/firmware-hls/blob/e794563ac2d4289c9471265ecdfebb974b87cd93/TrackletAlgorithm/TrackMerger.h#L17-L18

And you also have default member initializations in both `ComparisonModule` (which obviate the initializations in the constructor, BTW) and `TrackStruct`:
https://github.com/cms-L1TK/firmware-hls/blob/e794563ac2d4289c9471265ecdfebb974b87cd93/TrackletAlgorithm/TrackMerger.h#L37-L41
https://github.com/cms-L1TK/firmware-hls/blob/e794563ac2d4289c9471265ecdfebb974b87cd93/TrackletAlgorithm/TrackHandler.h#L22-L24

When declaring an array of `ComparisonModule` or `TrackStruct` objects, the compiler will generate an implicit loop to initialize all these members. Unfortunately, being implicit means you cannot unroll the loop as you normally would with the HLS unroll pragma.

So you could try moving all these initializations to separate methods (assuming they are all truly needed), and then calling these methods in explicit loops after declaring your arrays of `ComparisonModule`/`TrackStruct`. Then you can just add the unroll pragma to these loops, e.g., as is done in the TrackletProcessor:
https://github.com/cms-L1TK/firmware-hls/blob/e794563ac2d4289c9471265ecdfebb974b87cd93/TrackletAlgorithm/TrackletProcessor.h#L569-L575

However, I'm more confused that the arrays don't update properly without `static`. Does that mean the results of one event depend on something happening in the previous event? Or am I missing something?

aehart:
> > 
> 
> Hi Andrew, it's because I am using the old test vectors and not the current ones to debug the module - I had to copy the old files across so should I just set the test bench to a false pass?

Can you remind me how the current test vectors differ from the old ones? And does the C-simulation pass when using the old test vectors?

kwalkingshaw:
> > > 
> > 
> > 
> > Hi Andrew, it's because I am using the old test vectors and not the current ones to debug the module - I had to copy the old files across so should I just set the test bench to a false pass?
> 
> Can you remind me how the current test vectors differ from the old ones? And does the C-simulation pass when using the old test vectors?

I've just run with the latest test vectors, and it works okay. It doesn't pass due to the software emulation being different from the f/w, so in the test bench I force it to pass so it's able to synthesise.

TestBenches/TrackMerger_test.cpp
```diff
@@ -66,67 +67,56 @@ int main(){
...
       
       outputTracks.write_mem(bx, track, nTracks );
       ++nTracks;
+      
     } 
 
     // Comparing outputs
     err_count += compareMemWithFile<TrackFitMemory<NBarrelStub, NDiskStub>>(outputTracks, fout_outputTracks.at(0), ievt, "Tracks", truncation);
-    
+    outputTracks.clear();
```
aehart, line 119: @kwalkingshaw I think this is superfluous, since you have the same thing here:
https://github.com/cms-L1TK/firmware-hls/blob/e7b4815cb98c9303643b60dd2f91a77bd0b45040/TestBenches/TrackMerger_test.cpp#L57

aehart:
> > > > 
> > > 
> > > 
> > > Hi Andrew, it's because I am using the old test vectors and not the current ones to debug the module - I had to copy the old files across so should I just set the test bench to a false pass?
> > 
> > 
> > Can you remind me how the current test vectors differ from the old ones? And does the C-simulation pass when using the old test vectors?
> 
> I've just run with the latest test vectors, and it works okay. It doesn't pass due to the software emulation being different from the f/w, so in the test bench I force it to pass so it's able to synthesise.

OK, this all makes sense to me… again. Thank you for refreshing my memory about why the test bench is set up this way.

Could you maybe add a brief comment near the commented-out return statement in the test bench that explains this? Just so that I and anyone else looking at this isn't confused in the future.

kwalkingshaw:
> > > > > 
> > > > 
> > > > 
> > > > Hi Andrew, it's because I am using the old test vectors and not the current ones to debug the module - I had to copy the old files across so should I just set the test bench to a false pass?
> > > 
> > > 
> > > Can you remind me how the current test vectors differ from the old ones? And does the C-simulation pass when using the old test vectors?
> > 
> > 
> > I've just run with the latest test vectors, and it works okay. It doesn't pass due to the software emulation being different from the f/w, so in the test bench I force it to pass so it's able to synthesise.
> 
> OK, this all makes sense to me… again. Thank you for refreshing my memory about why the test bench is set up this way.
> 
> Could you maybe add a brief comment near the commented-out return statement in the test bench that explains this? Just so that I and anyone else looking at this isn't confused in the future.

Sure, have just pushed to the repo.

aehart (approved):
Thank you for the well-written code, @kwalkingshaw. Approved!

PR #255, opened 2023-01-30
Barrel-only test-bench fix

aehart:
This PR just updates the project_generation_scripts submodule, which now includes a fix to the test-bench for the barrel-only project.


bylsma (approved):
Looks good.

PR #242, opened 2022-07-19
MC/MP disk

bryates:
This PR adds the disk implementation for the MC and MP.

Current status:
MC - Full agreement for all barrel and disk layers, _except_ a single FullMatch in D2 (investigating)
MP - Work in progress. Initial code ported from MC, additional changes required (disk sometimes us an extra `zbin` bit)

PR #256, opened 2023-01-31
second reduced combined chain, new VMSME memory module

jasonfan393:
PR adds second combined module chain (larger than reduced but not full barrrel).  Adds a new vhdl binned memory module for the VMSME memories, and the necessary changes to MP to interface with it. Also addresses some small bugs with reduced CM chain.

corresponds to [#49](https://github.com/cms-L1TK/project_generation_scripts/pull/49) in project_generation_scripts


aehart:
@jasonfan393 Once https://github.com/cms-L1TK/project_generation_scripts/pull/49 is merged, I don't see anywhere in firmware-hls or project_generation_scripts were tf_mem_bin_cm4 is used. Can tf_mem_bin_cm4_new simply replace tf_mem_bin_cm4 instead of being a separate module?

I wanted to ask before merging https://github.com/cms-L1TK/project_generation_scripts/pull/49 and locking in the name.

jasonfan393:
> @jasonfan393 Once [cms-L1TK/project_generation_scripts#49](https://github.com/cms-L1TK/project_generation_scripts/pull/49) is merged, I don't see anywhere in firmware-hls or project_generation_scripts were tf_mem_bin_cm4 is used. Can tf_mem_bin_cm4_new simply replace tf_mem_bin_cm4 instead of being a separate module?
> 
> I wanted to ask before merging [cms-L1TK/project_generation_scripts#49](https://github.com/cms-L1TK/project_generation_scripts/pull/49) and locking in the name.


I think you're right, looks like this is just a relic of development and I've edited both this and the pull request in project generation scripts to match. hopefully it works as intended

aehart:
Now that https://github.com/cms-L1TK/project_generation_scripts/pull/49 has been merged, please update the submodule to point to the HEAD of the master branch.

jasonfan393:
@aehart submodule updated now.

jasonfan393:
@aehart Any idea what is causing CI to fail randomly? seems to regularly randomly fail on TopPRMEMC-synth, but passes sometimes on re-run. Looks like something to do with makefiles
[passing example](https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/5182604)
[failing example](https://gitlab.cern.ch/cms-l1tk/firmware-hls/-/pipelines/5194548)

IntegrationTests/common/hdl/tf_pkg.vhd
```diff
@@ -98,6 +99,10 @@ package tf_pkg is
   subtype t_arr8_7b is t_arr_7b(0 to 7);
   type t_arr8_2_7b is array(0 to 7) of t_arr2_7b;
 
+  type t_arr_8_4b  is array(integer range<>) of t_arr8_4b;
+  subtype t_arr2_8_4b is t_arr_8_4b(0 to 1);
+  subtype t_arr4_8_4b is t_arr_8_4b(0 to 3);
```
aehart, line 104: @jasonfan393 Likewise, these don't seem to be used anywhere.

IntegrationTests/common/script/Makefile.mk
```diff
@@ -44,7 +44,10 @@ MemPrints:
 	   ln -sf $(EMDATA)/MemPrintsReducedCM MemPrints; \
 	 elif [[ $${BASE_DIR} == "ReducedConfig" ]]; \
 	 then \
-	   ln -sf $(EMDATA)/MemPrintsReduced MemPrints; \
+	   ln -sf $(EMDATA)/ReducedConfig MemPrints; \
```
aehart, line 47: @jasonfan393 Did you mean to change the target of this link?

emData/download.sh
```diff
@@ -14,6 +14,9 @@ luts_url_cm="https://cernbox.cern.ch/remote.php/dav/public-files/LoqjBIg2ARZav8u
 # Reduced Combined modules                                                      
 memprints_url_reducedcm="https://aryd.web.cern.ch/aryd/MemPrints_CombinedReduced_220807.tgz"
 luts_url_reducedcm="https://aryd.web.cern.ch/aryd/LUTs_CombinedReduced_220807.tgz"
+# Reduced Combined modules2
+memprints_url_cm2="https://aryd.web.cern.ch/aryd/MemPrints_CombinedReduced2_220829.tgz"
+luts_url_cm2="https://jafan.web.cern.ch/jafan/HLSFiles/LUTs_CombinedReduced2_230201.tgz"
```
aehart, line 19: @jasonfan393 Can you add instructions to the README explaining how these files were produced?

IntegrationTests/common/hdl/tf_pkg.vhd
jasonfan393: removed them

IntegrationTests/common/hdl/tf_pkg.vhd
```diff
@@ -85,6 +85,7 @@ package tf_pkg is
   type t_arr8_8_1b is array(0 to 7) of t_arr8_1b;
   type t_arr8_8_4b is array(0 to 7) of t_arr8_4b;
   -- 4D
+  type t_arr4_8_8_4b is array(0 to 3) of t_arr8_8_4b;
```
aehart, line 88: @jasonfan393 I meant to comment on this line earlier as well (hence, the stray "Likewise" on the previous comment on this file).

Anyway, this type also doesn't seem to be used anywhere.

TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -256,7 +262,7 @@ inline void step(const VMStubMECM<VMSMEType> stubmem[4][1024]) {
 inline bool processing() {
 #pragma HLS inline  
   return !idle_||good__||good__t||good___;
```
aehart, line 264: @jasonfan393 Please indent these lines to be consistent with where the closing curly brace is now.

TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -373,7 +379,7 @@ inline void advance() {
  ap_uint<4> stubmask_;
  ap_uint<1> second_;
  ap_uint<1> phiPlus_;
- ap_int<2> shift_;
+ ap_uint<1> phiProjBin_;
```
aehart, line 382: @jasonfan393 I think we could eliminate a magic number by taking the type from `ProjectionRouterBuffer`:
```suggestion
 ProjectionRouterBuffer<BARREL, AllProjectionType>::PHIPROJBIN phiProjBin_;
```

And this might apply to other types within this class…

TrackletAlgorithm/MatchProcessor.h
```diff
@@ -803,51 +820,104 @@ void MatchProcessor(BXType bx,
 
       unsigned int ivmPlus = iphi;
 	
-      ap_int<2> shift = 0;
+      //ap_int<2> shift = 0;
       
+      //If the projection goes to the first (ivmMinus) phiProjBin is zero but
+      //if the projection goes to the second bin ivmMinus+1 phiProjBin is one
+      ap_uint<1> phiProjBin = 0;
```
aehart, line 827: @jasonfan393 I guess we can elimimate a magic number by taking the type from `ProjectionRouterBuffer`:
```suggestion
      ProjectionRouterBuffer<BARREL, APTYPE>::PHIPROJBIN phiProjBin = 0;
```
Maybe other magic numbers could be eliminated in a similar way…

TrackletAlgorithm/MatchProcessor.h
```diff
@@ -803,51 +820,104 @@ void MatchProcessor(BXType bx,
...
+      //if (ivmMinus==ivmPlus) {
+      //	nstubfirstPlus = 0;
+      //	nstublastPlus = 0;
+      //}
+      //if (zfirst==zlast) {
+      //	nstublastMinus = 0;
+      //	nstublastPlus = 0;
+      //}
+
+      ap_uint<4> maskstubs=(stublastPlus, stubfirstPlus, stublastMinus, stubfirstMinus);
```
aehart, line 890: @jasonfan393 There are a lot of 4s and 8s floating around here, as well as in MemoryTemplateBinnedCM.h. Could you please replace these with named constants, ideally sharing them between here and MemoryTemplateBinnedCM.h if appropriate?

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -46,6 +46,8 @@ class MemoryTemplateBinnedCM{
 
   ap_uint<8> binmask8_[kNBxBins][8];
   ap_uint<32> nentries8_[kNBxBins][8];
+  ap_uint<32> nentries8A_[kNBxBins*8];
+  ap_uint<32> nentries8B_[kNBxBins*8];
```
aehart, line 50: @jasonfan393 Is there a reason these are not two-dimensional arrays, like the other data members are?

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -114,16 +132,20 @@ class MemoryTemplateBinnedCM{
     if(!NBIT_BX) ibx = 0;
     if (nentry_ibx < getNEntryPerBin()-1) { // Max 15 stubs in each memory due to 4 bit nentries
       // write address for slot: getNEntryPerBin() * slot + nentry_ibx
-      writememloop: for (signed int icopy=0; icopy< (signed) NCOPY;icopy++) { // Casting to signed int to avoid unsigned int comparison compilation error
+  
+    writememloop:for (unsigned int icopy=0;icopy<NCOPY;icopy++) {
```
aehart, line 136: @jasonfan393 Please add two spaces to the indent of this line.

IntegrationTests/common/hdl/tf_mem_bin_cm4.vhd
```diff
@@ -33,6 +33,7 @@ entity tf_mem_bin_cm4 is
     NUM_PAGES       : natural := 2;                --! Specify no. Pages in RAM memory
     RAM_DEPTH       : natural := NUM_PAGES*PAGE_LENGTH_CM; --! Leave at default. RAM depth (no. of entries)
     NUM_MEM_BINS    : natural := 64;                --! Specify number of memory bins
+    BIN_ADDR_WIDTH  : natural := 4;                 --! Bits for address
```
aehart, line 36: @jasonfan393 I think this parameter can be derived from other constants as `clogb2(NUM_ENTRIES_PER_MEM_BINS)`. Could you use this in the body of the code, or at least define `BIN_ADDR_WIDTH` to have this value?

IntegrationTests/common/hdl/tf_mem_bin_cm4.vhd
```diff
@@ -69,6 +75,7 @@ architecture rtl of tf_mem_bin_cm4 is
 
 -- ########################### Types ###########################
 type t_arr_1d_slv_mem is array(0 to RAM_DEPTH-1) of std_logic_vector(RAM_WIDTH-1 downto 0); --! 1D array of slv
+type t_arr_1d_slv_mem_nent is array(0 to 31) of std_logic_vector(3 downto 0); --! 1D array of slv
```
aehart, line 78: @jasonfan393 Same comment here regarding the hard-coded 31 and 3.

IntegrationTests/common/hdl/tf_mem_bin_cm4.vhd
aryd: I have modified the code such that we now derive this number.

IntegrationTests/common/hdl/tf_mem_bin_cm4.vhd
```diff
@@ -60,7 +61,12 @@ entity tf_mem_bin_cm4 is
     doutb2     : out std_logic_vector(RAM_WIDTH-1 downto 0);         --! RAM output data
     doutb3     : out std_logic_vector(RAM_WIDTH-1 downto 0);         --! RAM output data
     sync_nent : in  std_logic;                                      --! Synchronize nent counter; Connect to start of reading module
-    nent_o    : out t_arr_64_4b(0 to NUM_PAGES-1) := (others => (others => (others => '0'))); --! entries(page)(bin)
+    enb_nentA  : in  std_logic;                                      --! Read Enable, for additional power savings, disable when not in use
+    enb_nentB  : in  std_logic;                                      --! Read Enable, for additional power savings, disable when not in use
+    addr_nentA : in std_logic_vector(4 downto 0); --! Addres for nentries 
+    addr_nentB : in std_logic_vector(4 downto 0); --! Addres for nentries 
+    dout_nentA    : out std_logic_vector(31 downto 0); --! entries output data
+    dout_nentB    : out std_logic_vector(31 downto 0); --! entries output data
```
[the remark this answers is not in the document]
aryd: These are now derived from existing constants.

IntegrationTests/common/hdl/tf_mem_bin_cm4.vhd
aryd: Also derived from existing constants.

TrackletAlgorithm/MatchProcessor.h
jasonfan393: @aehart working on replacing magic 4s and 8s now. Just wondering if you have any suggestion on how to handle removing the reference to 8 in line 848. I guess this can be replaced with a loop, but unsure if that will be implemented different by HLS. Curious if there is a way to still do this with the parentheses without explicit reference to 6,5,4 etc.

`
      (entries_zfirst[7], entries_zfirst[6], entries_zfirst[5],
       entries_zfirst[4], entries_zfirst[3], entries_zfirst[2],
       entries_zfirst[1],entries_zfirst[0]) = 
	//instubdata.get_mem_entries8A()[bx][zfirst];      
	instubdata.get_mem_entries8A()[(bx&3)*8+zfirst];      
`
`
      (entries_zlast[7], entries_zlast[6], entries_zlast[5],
       entries_zlast[4], entries_zlast[3], entries_zlast[2],
       entries_zlast[1],entries_zlast[0]) = 
	//instubdata.get_mem_entries8B()[bx][zlast];      
	instubdata.get_mem_entries8B()[(bx&3)*8+zlast];      
`

bryates: A loop with the unroll pragma might work well here.

jasonfan393:
note that checks are 'passing'  but that's just because I've adopted the allowed to fail tag from master, need to clear up some bug with my cleanups in the memory vhdl module.

TrackletAlgorithm/MatchProcessor.h
jasonfan393: @aehart -removed most of the 4s and 8s highlighted . There are a lot of them in MemoryTemplateBinnedCM prior to any changes I've made, which will also need changes in TP if I want to rename all these 'nentries8*' things. Should we just put this on a separate PR/issue for now?

aehart: I think it's fine to have a limited amount of hard-coded numbers, especially in the names of data members, and especially while things are under such intensive development. We can think about how to make it prettier when everything works.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
aehart: Sounds good to me.

IntegrationTests/common/hdl/tf_mem_bin_cm4.vhd
```diff
@@ -196,6 +214,16 @@ begin
...
       sv_RAM_row3 <= sa_RAM_data3(to_integer(unsigned(addrb3)));
     end if;
+    if (enb_nentA='1') then
+      for i in 0 to NUM_PHI_BINS-1 loop
+        dout_nentA(BIN_ADDR_WIDTH*(i+1)-1 downto BIN_ADDR_WIDTH*i) <= sa_RAM_nentA(i)(to_integer(unsigned(addr_nentA)));
+      end loop;
+    end if;
+    if (enb_nentB='1') then
+      for i in 0 to NUM_PHI_BINS-1 loop
+        dout_nentB(BIN_ADDR_WIDTH*(i+1)-1 downto BIN_ADDR_WIDTH*i) <= sa_RAM_nentB(i)(to_integer(unsigned(addr_nentA)));
```
aehart, line 224: @jasonfan393 Should the `addr_nentA` on this line be `addr_nentB`?

TrackletAlgorithm/ProjectionRouterBuffer.h
```diff
@@ -141,8 +145,8 @@ template<int VMProjType, int AllProjectionType>
     return data_.range(kPRBufferTCIDMSB,kPRBufferTCIDLSB);
   }
   
-  SHIFT shift() const {
-    return data_.range(kPRBufferShiftMSB,kPRBufferShiftLSB);
+  PHIPROJBIN phiProjBin() const {
```
aehart, line 148: @jasonfan393 For consistency with the other getters, can you start the name of this method with `get`?
```suggestion
  PHIPROJBIN getPhiProjBin() const {
```

emData/generate_MP.py
```diff
@@ -202,6 +202,8 @@ def FMRegion(region):
         topFile.write(
             "#pragma HLS resource variable=allstub->get_mem() latency=2\n"
             "#pragma HLS resource variable=instubdata.get_mem() latency=2\n"
+            "#pragma HLS resource variable=instubdata.get_mem_entries8A() latency=0\n"
+            "#pragma HLS resource variable=instubdata.get_mem_entries8B() latency=0\n"
```
aehart, line 206: @jasonfan393 During C-synthesis, I see the following warnings:
```
WARNING: [TECH 200-23] The specified latency 0 on variable ('instubdata_nentries8A_V_load', /home/ahart/TrackFinding/jason/firmware-hls/TrackletAlgorithm/MatchProcessor.h:858->/home/ahart/TrackFinding/jason/firmware-hls/TopFunctions/ReducedCombinedConfig2/MatchProcessorTop.cc:29) is smaller than the min latency 1 of core 'RAM'. The min latency 1 is selected.
WARNING: [TECH 200-23] The specified latency 0 on variable ('instubdata_nentries8B_V_load', /home/ahart/TrackFinding/jason/firmware-hls/TrackletAlgorithm/MatchProcessor.h:859->/home/ahart/TrackFinding/jason/firmware-hls/TopFunctions/ReducedCombinedConfig2/MatchProcessorTop.cc:29) is smaller than the min latency 1 of core 'RAM'. The min latency 1 is selected.
```

Could you set the latency to one here, to avoid these warnings? I'm not sure if this affects the logic of the SectorProcessors that include the MP.

aryd:
Andrew,

 The binned memories are segmented as 8x8 in phi x z (or phi x r in disks). The entries8_ contains the 8 entries for all phi bins. When we process projections in the ME we need to look either one or two bins in phi and z if the projection is near a boundary between two bins. (Could be four bins if the projection is in the corner between four regions.)

 We used to have nentries8_ which was a fully partitioned array and this allows us to ready multiple slots on each clock. But I’m changing these to be implemented as DRAM and then we can only do one read and one write per clock, hence now we need two memories, nentries8A_ and nentries8B_. (I have worked on a better version of the memory module that implements this as one array that returns 16 entries. Hence, I think we should not further change the current version.)

 nentries8A_ and nentries8B_ can not be zeroed out as there is no way of zeroing all elements in a DRAM as far as I understand it - only entries where binmask8_ is non-zero will we have valid entries in nenetries8{A,B}.

 Is nentries8_ actually used?

 -Anders



Anders Ryd
***@***.******@***.***>



On Jun 12, 2023, at 7:38 PM, Andrew Hart ***@***.******@***.***>> wrote:



@aehart commented on this pull request.

________________________________

In TrackletAlgorithm/MemoryTemplateBinnedCM.h<https://github.com/cms-L1TK/firmware-hls/pull/256#discussion_r1227026956>:

> +  ap_uint<32> nentries8A_[kNBxBins*kNPhiBins];
+  ap_uint<32> nentries8B_[kNBxBins*kNPhiBins];


@jasonfan393<https://github.com/jasonfan393> Can you remind me why these extra nentries are needed? Unless I missed something, nentries8_, nentries8A_, and nentries8B_ are always equal.

Also, should nentries8A_ and nentries8B_ be set to zero in MemoryTemplateBinnedCM::clear, along with binmask8_ and nentries8_?
https://github.com/cms-L1TK/firmware-hls/blob/9c6192830760809619171e7207acac27973c09ef/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L184-L188

jasonfan393:
@aryd this version has regular nentries8_ still used in the trackletprocessor, the plan is to remove this but this version needs both versions of nentries

aehart:
> We used to have nentries8_ which was a fully partitioned array and this allows us to ready multiple slots on each clock. But I’m changing these to be implemented as DRAM and then we can only do one read and one write per clock, hence now we need two memories, nentries8A_ and nentries8B_. (I have worked on a better version of the memory module that implements this as one array that returns 16 entries. Hence, I think we should not further change the current version.)

@aryd Thank you for clarifying. I thought something like this was probably the reason. But when you're just looking at the C++, it looks a little strange.

> nentries8A_ and nentries8B_ can not be zeroed out as there is no way of zeroing all elements in a DRAM as far as I understand it - only entries where binmask8_ is non-zero will we have valid entries in nenetries8{A,B}.

I noticed that @bryates does this in his PR:
https://github.com/cms-L1TK/firmware-hls/blob/56629526fc9e92ec3c0a401412628f49ecc20f70/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L202-L205
I don't think this changes the behavior of the firmware, since `MemoryTemplateBinnedCM::clear` is never called in the HLS code, either directly or via the constructor. It's only ever called in the test benches, where the memories are constructed.

That being said, maybe it isn't stricly needed, but it is consistent with the existing code.

bryates:
> > We used to have nentries8_ which was a fully partitioned array and this allows us to ready multiple slots on each clock. But I’m changing these to be implemented as DRAM and then we can only do one read and one write per clock, hence now we need two memories, nentries8A_ and nentries8B_. (I have worked on a better version of the memory module that implements this as one array that returns 16 entries. Hence, I think we should not further change the current version.)
> 
> @aryd Thank you for clarifying. I thought something like this was probably the reason. But when you're just looking at the C++, it looks a little strange.
> 
> > nentries8A_ and nentries8B_ can not be zeroed out as there is no way of zeroing all elements in a DRAM as far as I understand it - only entries where binmask8_ is non-zero will we have valid entries in nenetries8{A,B}.
> 
> I noticed that @bryates does this in his PR:
> 
> https://github.com/cms-L1TK/firmware-hls/blob/56629526fc9e92ec3c0a401412628f49ecc20f70/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L202-L205
> 
> 
> I don't think this changes the behavior of the firmware, since `MemoryTemplateBinnedCM::clear` is never called in the HLS code, either directly or via the constructor. It's only ever called in the test benches, where the memories are constructed.
> That being said, maybe it isn't stricly needed, but it is consistent with the existing code.

I found i had to zero them out or I saw inconsistencies. I haven't checked it this causes any issues when running the firmware, but the CI seems to pass fine.

aehart:
> I found i had to zero them out or I saw inconsistencies. I haven't checked it this causes any issues when running the firmware, but the CI seems to pass fine.

I could see it being necessary in the test bench because of how the memories are compared.

FWIW, regarding this question, I don't think anything needs to change in this PR, and I've marked the original comment "resolved". There are just four unresolved comments now, which I think should be straightforward to address. Then we can merge this and move remaining synchronization/rebasing issues to #242.

aryd:
Brent, Andrew,

 I agree that this would be an issue for the validation of the VMRouterCM - but as the MP only reads these memories they should not need to be zeroed. Brent - just to confirm - you needed to zero these to get agreement in the MP? If so we should check why this is.

 -Anders

Anders Ryd
***@***.******@***.***>



On Jun 19, 2023, at 3:49 PM, Brent R. Yates ***@***.******@***.***>> wrote:



We used to have nentries8_ which was a fully partitioned array and this allows us to ready multiple slots on each clock. But I’m changing these to be implemented as DRAM and then we can only do one read and one write per clock, hence now we need two memories, nentries8A_ and nentries8B_. (I have worked on a better version of the memory module that implements this as one array that returns 16 entries. Hence, I think we should not further change the current version.)

@aryd<https://github.com/aryd> Thank you for clarifying. I thought something like this was probably the reason. But when you're just looking at the C++, it looks a little strange.

nentries8A_ and nentries8B_ can not be zeroed out as there is no way of zeroing all elements in a DRAM as far as I understand it - only entries where binmask8_ is non-zero will we have valid entries in nenetries8{A,B}.

I noticed that @bryates<https://github.com/bryates> does this in his PR:

https://github.com/cms-L1TK/firmware-hls/blob/56629526fc9e92ec3c0a401412628f49ecc20f70/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L202-L205

I don't think this changes the behavior of the firmware, since MemoryTemplateBinnedCM::clear is never called in the HLS code, either directly or via the constructor. It's only ever called in the test benches, where the memories are constructed.
That being said, maybe it isn't stricly needed, but it is consistent with the existing code.

I found i had to zero them out or I saw inconsistencies. I haven't checked it this causes any issues when running the firmware, but the CI seems to pass fine.

aehart (approved):
This looks good to me now, so I am approving it. Since it touches some of the same files, rebasing #242 will be nontrivial, but I hope @jasonfan393 and @aryd can help with it.

PR #264, opened 2023-03-16
Memory fixes

aehart:
While working on the simulation of the barrel-only project, I noticed seemingly random differences related to the `nent_o` ports on the memory modules. Looking at the logic, I see that the output is not always well-defined, in particular, before the first write-enable signal is received.

I fix this by adding an integer variable that is initialized to zero and is assigned one when the first write-enable is received. Zeroes are output on `nent_o` while this variable is zero. Something similar is done for the `mask_o` ports on the binned memories for the combined module chains.

This seems to have no effect on the results from the skinny chain, but greatly improves the agreement for the barrel-only project:
|   | `master` branch | `mem_fix` branch |
| ------------- | ------------- | ------------- |
| bad events in TF_L1L2 | 100 | 12 |
| bad events in TF_L2L3 | 58 | 0 |
| bad events in TF_L3L4 | 100 | 10 |
| bad events in TF_L5L6 | 100 | 17 |

I've also added the unbinned memory module based on Thomas Schuh's work, which is currently used on the barrel_config branch. This is just for reference, and it is not yet used here.

### DIFF
--- a/IntegrationTests/BarrelConfig/IRtoTB/script/makeProject.tcl
+++ b/IntegrationTests/BarrelConfig/IRtoTB/script/makeProject.tcl
@@ -179,6 +179,7 @@ add_files -fileset sources_1 [glob ../hdl/SectorProcessor*.vhd]
 add_files -fileset sources_1 [glob ../hdl/memUtil_pkg.vhd]
 add_files -fileset sources_1 [glob common/hdl/*.vhd]
 remove_files -fileset sources_1 [glob common/hdl/latency_monitor.vhd]
+remove_files -fileset sources_1 [glob common/hdl/tf_mem_new.vhd]
 
 # Add HDL for TB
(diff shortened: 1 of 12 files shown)


pwittich (commented):
This looks good to me except I think giving a file a name with `_new` in it will look a bit silly in 5 years.... maybe a better name should be considered.

aehart:
> This looks good to me except I think giving a file a name with `_new` in it will look a bit silly in 5 years.... maybe a better name should be considered.

I only add that file here because I think I promised to make it available once it was debugged on the barrel_config branch, even before it's actually used anywhere on the master branch. Then, once the project generation scripts are updated to use it, it will replace the current tf_mem.vhd file.

So it's provided now for convenience, and the "_new" suffix is temporary.

pwittich (approved):
Approve with the understanding that the name will be changed once this starts being used.

PR #265, opened 2023-03-17
Changes for the vitis migration

KathrynColdham:
Changes for the Vitis migration for the Match Engine module. Also some changes to prevent output folder names for other modules being overwritten when switching between HLS Vivado and Vitis.


tomalin:
You can see from the git CI output https://github.com/cms-L1TK/firmware-hls/pull/265/checks that this failed to rebase your branch on top of Peter's. And if I look at the differences between your branch and Peter's in this PR, it says that your PR would delete for example all the TrackQuality code. 
I suspect that you created your branch from the latest "master", whereas Peter created his branch from an old master. So try in a new empty directory, rebasing Peters branch to the latest master without any of your code changes, and make a 2nd PR which just does this. If we merge that PR, it should stop CI complaining about this one.

README.md
```diff
@@ -29,34 +29,6 @@ This would create a project directory \<project> ("projrouter" in case of the ab
 5) `make Work/Work.runs/synth_1` (runs synthesis, writes utilization & timing reports to current directory).
 6) `make Work/Work.runs/impl_1` (runs implementation, writes utilization & timing reports to current directory). N.B. This step is optional, and not required for code validation.
 
-## Track Quality Specific Instructions
```
pwittich, line 32: Did you intend to delete this section on Track Quality?

pwittich:
hi @KathrynColdham it looks like this PR deletes a bunch of Track Quality files. Is this intentional?

tomalin:
> hi @KathrynColdham it looks like this PR deletes a bunch of Track Quality files. Is this intentional?

Hi @pwittich as mentioned in my earlier comment on this PR, I suspect that Kathryn created her branch from "master", where the TQ was indeed deleted by Chris Brown, whereas you created your branch from "master" before Chris did that.

pwittich:
> 

I missed that. I guess I should rebase this branch on master.

emData/generate_PR.py
```diff
@@ -74,7 +74,7 @@ def writeSourceModuleInstance(module, wiresFileName):
     strSource += "#pragma HLS inline off\n"
     strSource += "#pragma HLS interface mode=ap_vld register port=bx_o\n"
     for i in range(TProjCount):
-        strSource += "#pragma HLS interface mode=ap_memory port=projin[" + str(i) + "].get_mem() latency=2\n"
```
pwittich, line 77: if I am reading this diff correctly, there is no longer any pragma for these interfaces; I think we do need that?

pwittich:
If I am understanding the diffs, the only change that affect the source code in this PR is in `emData/generate_PR.py`. The other changes allow us to handle Vitis and Vivado at the same time which is great. So I am ok to merge it once i understand my comment in the change above.

emData/generate_PR.py
pwittich: also it looks like with this change the loop is empty, so if the line should be removed the `for` statement can also be removed.

tomalin:
@pwittich & @KathrynColdham I fear this merge was made prematurely. The git CI is complaining of conflicts, probably because git is confused by Kathryn having created her branch from master instead of vitis_migration. I just did a test of the new "vitis_migration" by checking it out and then trying to rebase it to "master" and got git errors.
I think the solution is to rebase revert this merge, rebase vitis_migration to master, and then try merge again. I can try this if you like.

pwittich:
The CI was never tested on this branch. I'll fix it when it gets pushed to master.

pwittich:
@tomalin I rebased the `vitis_migration` branch onto master

PR #266, opened 2023-03-17
Add future-emulation-specific numInputs function in VMRCM

meisonlikesicecream:
Added two functions in the VMRCM parameters file that will be used by the VMRProducer only, as requested. Does not affect the HLS.


aehart (approved):
Looks good to me. Approved!

PR #269, opened 2023-03-23
EMP for SC

FilMarini:
Added files for EMP build of Skinny Chain.

The only modification I've done to the files already present is in

- [`IntegrationTests/common/hdl/tf_pkg.vhd`](https://github.com/cms-L1TK/firmware-hls/blob/emp_for_sc/IntegrationTests/common/hdl/tf_pkg.vhd#L398)
The modification is due to a conflict between the function declaration and its body, where in the first `N_PAGES` is declared as `natural` while in the latter is declared as `integer`. 
This conflict was throwing an error in the EMP Questa simulation phase.


README.md
```diff
@@ -29,6 +29,34 @@ This would create a project directory \<project> ("projrouter" in case of the ab
 5) `make Work/Work.runs/synth_1` (runs synthesis, writes utilization & timing reports to current directory).
 6) `make Work/Work.runs/impl_1` (runs implementation, writes utilization & timing reports to current directory). N.B. This step is optional, and not required for code validation.
 
+## Track Quality Specific Instructions
```
pwittich, line 32: I think all this TQ stuff was deleted form the repo, it probably is not your intention to add it back in....

pwittich (changes requested):
This looks good except I think the readme reverts some text that was intentionally removed.

README.md
FilMarini: Hi Peter, I've fixed the README and commited

pwittich (approved):
Looks fine to me; I guess the only question is what the appropriate layout of files is in the project (the EMP-specific ones). @aehart what do you think?

PR #272, opened 2023-04-02
Duplicate Removal

tschuh:
This PR adds a Duplicate Removal algorithm (no merging, just killing) which corresponds to the VHDL version found here https://github.com/cms-L1TK/l1tk-for-emp/tree/dr/dr/firmware and is in sync with the cmssw emulation found here https://github.com/cms-L1TK/cmssw/blob/L1TK-dev-12_6_0_pre5/L1Trigger/TrackFindingTracklet/plugins/ProducerDR.cc.

Compared with the vhdl version it has an increased latency by one clock tick (extreme marginal) and uses the double amount of LUTs (not extreme marginal, but not really an issue) and has a max logic level of 5 instead of 4 (also fine).

It took me a week (and help from Mei and Katie) to get this "10 liner" algorithm to work. HLS is a terrible language. The VHDL version was written and compared with s/w in 20 minutes...

Not sure if we actually want to merge this since I am not very motivated to maintain or develop this further. On the other hand side this might be helpful for Katie or a good starting point for whomever will overtake Katie's work.


pwittich:
hey @tschuh is this draft on purpose?

tschuh:
Actually yes since I am not sure if we want to merge it. But since I see now that it states not ready for review it makes no sense.

project/script_DR_hls.tcl
pwittich, line 1: could you rename this just `script_DR.tcl` to make it the same as all the other files. And could you also please source the common setup file used by other HLS modules, and follow the conventions therein for project names etc.

It looks like you are using vitis; is that right? Have you tested on vivado?

tschuh: I started with vivado_hls, but it failed cosim without any warnings. So I tried moving to vitis_hls where it just works.

TestBenches/DuplicateRemoval_test.cpp
```diff
@@ -0,0 +1,81 @@
...
+  for (int tick = 0; tick < numTracks + 1 + numComparisonModules - 1; tick++) {
+    const TrackIn& trackIn = tick < numTracks + 1 ? tracksIn[tick] : null;
+    TrackOut& trackOut = tick >= numComparisonModules - 1 ? tracksOut[tick - numComparisonModules + 1] : tmp;
+    DuplicateRemovalTop(trackIn, trackOut);
+  }
+
+  cout << "Expectation:" << endl;
+  print(tracksExp);
+  cout << "Simulation:" << endl;
+  print(tracksOut);
```
aehart, line 79: @tschuh Can `tracksExp` and `tracksOut` be compared in the code, as well as printed, and used to set the return value of `main`? The C-simulation and C/RTL cosimulation expect the test bench to return zero on success and non-zero if there are errors.

project/script_DR.tcl
```diff
@@ -0,0 +1,29 @@
+# script to generate half ip core DuplicateRemoval
+# excecute in firmware-hls/project
+# excecute with vitis_hls -f script_DR_hls.tcl
```
aehart, line 3: @tschuh
```suggestion
# excecute with vitis_hls -f script_DR.tcl
```

project/script_DR.tcl
```diff
@@ -0,0 +1,29 @@
...
+source settings_hls.tcl
+
+config_array_partition -complete_threshold 9999
+#config_unroll -tripcount_threshold 9999
+#config_interface -register_io  scalar_out
+config_rtl -reset all
+
+csynth_design
+#csim_design
+#cosim_design
```
aehart, line 26: @tschuh Assuming the C/RTL cosimulation works in Vitis, and assuming the test bench can be made to return an integer, I would suggest making the default behavior of the script like the other modules:
```suggestion
csim_design
csynth_design
cosim_design
```

project/script_DR_vivado.tcl
```diff
@@ -0,0 +1,16 @@
+# script to finalize ip core DuplicateRemoval
+# excecute in firmware-hls/project
+# excecute after vitis_hls -f script_DR_hls.tcl
```
aehart, line 3: @tschuh
```suggestion
# excecute after vitis_hls -f script_DR.tcl
```

aehart:
@tschuh In addition to the above inline comments, it looks like TopFunctions/DuplicateRemovalTop.h is missing.

Also, can you explain why the second step with project/script_DR_vivado.tcl is needed?

project/script_DR.tcl
tschuh: done

project/script_DR.tcl
tschuh: done

project/script_DR_vivado.tcl
tschuh: done

tschuh:
> @tschuh In addition to the above inline comments, it looks like TopFunctions/DuplicateRemovalTop.h is missing.
> 
> Also, can you explain why the second step with project/script_DR_vivado.tcl is needed?

Upsi, I added them. The second script is needed to produce all files (the .xci is missing) needed by IPBB to integrate this IP with the emp framework. I do not understand why vivado_hls or vitis_hls are not able to produce it, I guess that would be too handy...

aehart (approved):
Thanks, @tschuh. This looks good to me.

PR #273, opened 2023-04-17
Test (almost) all IRs

meisonlikesicecream:
Needed a little break from thesis-writing so added all* modules to the csim test for the IR.

\* all non-neg modules; the "neg" modules are not defined in the testbench. They could be added though my break is over for now :-)

### DIFF
--- a/project/script_IR.tcl
+++ b/project/script_IR.tcl
@@ -20,18 +20,6 @@ array set modules_to_test {
     9 {PS_1_B}
     10 {PS_2_A}
     11 {PS_2_B}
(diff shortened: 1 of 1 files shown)


aehart (approved):
Very nice update, @meisonlikesicecream. Could you also create an issue in GitHub for adding the "neg" IRs to the test bench so they can be tested as well?

meisonlikesicecream:
> Very nice update, @meisonlikesicecream. Could you also create an issue in GitHub for adding the "neg" IRs to the test bench so they can be tested as well?

sure!

PR #274, opened 2023-04-19
VMRCM avoid floating point core

meisonlikesicecream:
Some (one??) synthesised VMRCMs created an unnecessary floating point core for calculating the AllStub Inner r/z cuts which should've been computed at compile-time. Moved the cuts outside the main function solved the issue.


aehart (approved):
Thank you for the rapid fix, @meisonlikesicecream. Approved!

PR #279, opened 2023-06-21
revert accdiental change in MemoryTemplateteBinnedCM.h

jasonfan393:
PR reverts removal of '(signed)' in a integer comparison that is needed for future SW

### DIFF
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -133,7 +133,7 @@ class MemoryTemplateBinnedCM{
     if (nentry_ibx < getNEntryPerBin()-1) { // Max 15 stubs in each memory due to 4 bit nentries
       // write address for slot: getNEntryPerBin() * slot + nentry_ibx
 
-      writememloop:for (unsigned int icopy=0;icopy<NCOPY;icopy++) {
+      writememloop:for (unsigned int icopy=0;icopy< (signed) NCOPY;icopy++) {
 #pragma HLS unroll
 	dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry_ibx] = data;
       }


aehart:
@jasonfan393 I'm confused by this change. Normally we avoid comparing signed and unsigned values, but here we're forcing such a comparison?

Do you know what issue this is solving in the future SW?

mlarson02:
@aehart I requested this change for the future SW as it's causing a compilation error when initializing a VMStubTE memory with NCOPY = 0. This is affecting all of our CIs, as you can see here: https://gitlab.cern.ch/ejclemen/cmssw_trackfinding_hlsframework/-/jobs/30483265. I believe an alternative way to fix this would be to replace the 0's in the array [here ](https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/Constants.h#L18) (which we are using to get this NCOPY value) to be 1's.

ivanreid:
Has this been tested?  I get:

/home/eesridr/CMSSW_12_6_0_pre5/../firmware-hls/TrackletAlgorithm/MemoryTemplateBinnedCM.h:137:53: error: comparison of unsigned expression in '< 0' is always false [-Werror=type-limits]
 137 |         writememloop:for (unsigned int icopy=0;icopy< (signed) NCOPY;icopy++) {
         |                                                ~~~~~^~~~~~~~~~~~~~~~

As I understand it, when comparing unsigned and signed ints, the signed int is first promoted to unsigned, thus negating the cast to signed.  Would a better solution be
writememloop:for (unsigned int icopy=0; icopy != NCOPY; icopy++)
 ?  Then, if NCOPY is zero, the loop won't be executed, otherwise it will be executed NCOPY times.

mlarson02:
@ivanreid Sorry, my mistake, it looks like either that would work or both icopy and NCOPY would have to be signed. Just tested and either of these solutions would work.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -133,7 +133,7 @@ class MemoryTemplateBinnedCM{
     if (nentry_ibx < getNEntryPerBin()-1) { // Max 15 stubs in each memory due to 4 bit nentries
       // write address for slot: getNEntryPerBin() * slot + nentry_ibx
 
-      writememloop:for (unsigned int icopy=0;icopy<NCOPY;icopy++) {
+      writememloop:for (unsigned int icopy=0;icopy< (signed) NCOPY;icopy++) {
```
tomalin, line 136: This PR is not reverting to this code to the old version. The old version https://github.com/cms-L1TK/firmware-hls/blob/cacc74b170cebaa0e9ba48366e96ca007dfba9d8/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L117 declared icopy to be of type "signed int", not "unsigned int", which makes more sense given that NCOPY is now signed int.

tomalin:
> @aehart I requested this change for the future SW as it's causing a compilation error when initializing a VMStubTE memory with NCOPY = 0. This is affecting all of our CIs, as you can see here: https://gitlab.cern.ch/ejclemen/cmssw_trackfinding_hlsframework/-/jobs/30483265. I believe an alternative way to fix this would be to replace the 0's in the array [here ](https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/Constants.h#L18) (which we are using to get this NCOPY value) to be 1's.

@mlarson02 there is a logical difference between needing 0 or 1 TE per layer. So if we really don't need any in some layers, wouldn't it would be confusing to set this to 1?

TrackletAlgorithm/MemoryTemplateBinnedCM.h
ivanreid: If NCOPY really can be zero, then the < test will cause an unwanted loop iteration in that case, n'est-ce pas?

ivanreid: Sorry, the heat has scrambled my brain...

mlarson02:
@tomalin Yes I agree this would be confusing and should be avoided, but I believe in the VMRProducer we still need to declare the VMStubTE memories for each layer for scope purposes, as they're not used for these layers declaring with 0 vs. 1 copy wouldn't affect functionality. I think now that it's clear what the code would be reverted to this reversion would be the easiest and best solution.

tomalin (approved):
Approved, since it is simply a 1 line change that return that line to how it used to be. Though @jasonfan393 , please add a comment above this line explaining why "signed int" is needed. Otherwise someone may change it back in future.

PR #285, opened 2023-07-27
MP Refactoring/Cleanup

bryates:
This PR adds some additional updates to the MP disks (merge after #242). It improves the timing (currently down to 3.58 ns) and removes some unused variables.

### DIFF
--- a/TrackletAlgorithm/MatchEngineUnit.h
+++ b/TrackletAlgorithm/MatchEngineUnit.h
@@ -49,6 +49,7 @@ class MatchEngineUnit : public MatchEngineUnitBase<VMProjType> {
 #pragma HLS inline
     writeindex_ = 0;
     readindex_ = 0;
+    readindexnext = 1;
     stubmask_ = 0;
     nstubs_ = 0;
     idle_ = true;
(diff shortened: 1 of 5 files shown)


TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -448,6 +436,10 @@ inline void advance() {
  ProjectionRouterBuffer<VMProjType, AllProjectionType> projbuffer_;
  ap_uint<kNBits_MemAddr> projseq_;
  bool isPSseed_;
+ ap_uint<1> usefirstMinus;
```
aryd, line 439: Use trailing _ for member data.

aryd: Actually - as far as I can tell these are only used in one method, init(), so they should not be class variables, but local variables in the init method.

bryates: Good catch, I've moved them.

aryd (approved):
I'm happy and set this as approved so that we can merge this.

aryd:
We have an issue with MP_L2PHIC not giving the right result, but it was caused in an earlier PR with new test vectors so will leave this as is and merge this PR.

PR #288, opened 2023-08-08
Cm barrel chain with delay ryd

aryd:
This PR is a trial to include Jason's CM Barrel Chain with Delay code after updating to the latest project generation script and rebasing on master.


aryd:
The updates in this PR were included in PR 293 that was merged so I will close this PR.

PR #289, opened 2023-08-13
Speeded up download.sh

tomalin:
This runs the slowest parts of download.sh as background processes, so they run in parallel. On my computer, this reduces the time for download.sh to run from 16 minutes to 5 minutes.

The output files are identical, except for barrel_memorymodules.dat and barrel_processingmodules.dat, where the order of some elements changes.  But I think this is just because makeBarrelConfig.py uses a python "set", which doesn't preserve the order of the elements.


aryd (approved):
This looks fine to me. (In looking at the code I don't think we should rerun the scripts to generate the reduced configurations - but I will enter this as an issue.)

PR #290, opened 2023-08-14
Declare LUTs in MEU

bryates:
This PR addresses the MEU CPU usage issue found in CMSSW (future software). The LUTs for the MEU are now declared as `const` in the class definition, instead of inside the `step()` method.

Some spacing and tabs were also fixed.

### DIFF
--- a/TrackletAlgorithm/MatchEngineUnit.h
+++ b/TrackletAlgorithm/MatchEngineUnit.h
@@ -211,6 +211,11 @@ inline void step(const VMStubMECM<VMSMEType> stubmem[4][1<<(kNbitsrzbinMP+kNbits
     constexpr unsigned int kRInvSteps = 32;
     constexpr unsigned int kRInvBits = BITS_TO_REPRESENT(kRInvSteps - 1);
     
+    static const ap_uint<1 << 2*projfinephibits> phiLUT = isLessThanSize<projfinephibits,StubPhiPositionConsistency::kMax,false,projfinephibits,stubfinephibits>();
+    static const ap_uint<1 << 2*projfinephibits> stubZBarrelPS = isLessThanSize<projfinephibits,StubZPositionBarrelConsistency::kBarrelPSMax,true,stubfinephibits,projfinephibits>();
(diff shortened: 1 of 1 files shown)


TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -457,18 +451,27 @@ inline void advance() {
...
+ };
+  enum StubPhiPositionConsistency {
+    kMax = 3
  };
 
+ static constexpr int projfinephibits = VMProjectionBase<VMProjType>::kVMProjFinePhiWideSize;
+ static constexpr int projfinezbits = VMProjectionBase<VMProjType>::kVMProjFineZSize;
+ static constexpr int stubfinephibits = VMStubMECMBase<VMSMEType>::kVMSMEFinePhiSize;
+ static constexpr int stubfinezbits = VMStubMECMBase<VMSMEType>::kVMSMEFineZSize;
+ const ap_uint<1 << 2*projfinephibits> phiLUT = isLessThanSize<projfinephibits,StubPhiPositionConsistency::kMax,false,projfinephibits,stubfinephibits>();
```
tomalin, line 470: To make it clear to the compiler that constants are known at compile time, I find it necessary to use "static const" (in cases where constexpr unavailable). Are you able to add the "static" keyword? e.g. https://stackoverflow.com/questions/34026555/how-to-define-a-static-const-variable-of-a-template-class .

bryates: Using `static const` was causing issues in the HLS, but I can look a bit more into it.

bryates: Ok, I've moved the definitions back into the function, which allows `static const` to be used.

tomalin (approved):
Looks good.

TrackletAlgorithm/MatchEngineUnit.h
bryates: Should we go ahead and merge this, or wait for something to rebase the Future SW first to test?

tomalin: Yes, let's merge it.

PR #291, opened 2023-08-15
Future sw update

carriganm95:
Update to the generateMP script to put max full match and max tracklet projection arrays into different order. Previously was done alphabetically and is now changed to start with L1PHIA and end with D5PHID. These arrays are only used by the future SW so the ordering should not affect firmware.

### DIFF
--- a/emData/generate_MP.py
+++ b/emData/generate_MP.py
@@ -131,7 +131,7 @@ with open(os.path.join(dirname, arguments.outputDirectory, "MatchProcessor_param
     maxFMMems += str(len(TPMems.keys())) + "] = {"
 
     # Calculate parameters and print out parameters and top function for each MP.
-    for mpName in sorted(TPMems.keys()):
+    for mpName in sorted(TPMems.keys(), key = lambda x: x.startswith('L')):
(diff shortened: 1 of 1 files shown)


bryates (approved):
Looks good to me. @carriganm95 I think we'll want a rebase instead of a merger from the master, but otherwise this PR is good to merge once the CI is done.

bryates:
> Looks good to me. @carriganm95 I think we'll want a rebase instead of a merger from the master, but otherwise this PR is good to merge once the CI is done.

Maybe this isn't a big deal for this small PR, since we always squash before merging back into the master anyway.

carriganm95:
> Looks good to me. @carriganm95 I think we'll want a rebase instead of a merger from the master, but otherwise this PR is good to merge once the CI is done.

I think this should be fixed now

bryates:
> > Looks good to me. @carriganm95 I think we'll want a rebase instead of a merger from the master, but otherwise this PR is good to merge once the CI is done.
> 
> I think this should be fixed now

Perfect, thanks!

aryd:
I think this looks fine so I will go ahead and squash and merge.

PR #296, opened 2023-08-25
TP template cleanups and fixes for future SW

jasonfan393:
Removes some unnecessary templates on main TP function, moving them to be calculated with Seed/iTC template parameters. 

Fixes mismatch between NCOPY template parameter on VMSTE memories between VMRCM and TP. Moves some variable declarations outside main TP function loop (needed for future SW).


jasonfan393:
CI checks appear to be in unrelated integration test chains

TestBenches/TrackletProcessor_test.cpp
```diff
@@ -96,6 +96,12 @@ int main()
 #else
   #error "Undefined seed"
 #endif
+#if SEED_ == L2D1_
```
aryd, line 99: We should see if this can be done in a cleaner way (but should not hold up this PR - can yo enter as an issue?)

TrackletAlgorithm/TrackletProcessor.h
```diff
@@ -984,30 +986,33 @@ TF::seed Seed, // seed layer combination (TC::L1L2, TC::L3L4, etc.)
...
+  ap_uint<1> usenext;
  istep_loop: for(unsigned istep=0;istep<N;istep++) {
 #pragma HLS pipeline II=1 rewind
 
     /*
     std::cout << "istep="<<istep<<" TEBuffer: "<<tebuffer.getIStub()<<" "<<tebuffer.getMem()<<" "
               << tebuffer.readptr()<<" "<<tebuffer.writeptr()<<std::endl;
 
-    for (unsigned k = 0; k < NTEUnits; k++){
+    for (unsigned k = 0; k < kNTEUnits[Seed]; k++){
```
aryd, line 1015: Fix formatting?

aryd (commented):
Just some minor comment about formatting and a request to create an issue. After these are addressed we should merge.

TestBenches/TrackletProcessor_test.cpp
jasonfan393: I've just replaced all the NTEUnits with NCOPY instead, since NTEUNITs is no longer a template parameter of TP there was no need to use it to determine ncopy in the testbench.

aryd (approved):
I'm happy with the code now.

aryd:
This looks fine now - I will merge.

PR #297, opened 2023-08-31
TrackBuilder disk-only and overlap seeds

aehart:
This PR just adds what we need to test the TrackBuilder for the disk-only and overlap seeds.

The additional seeds have almost perfect agreement with the emulation. The only exception is a few events that have extra entries for D1D2. For now, I just exclude this TrackBuilder from the list of modules to test in project/script_TB.tcl.


jasonfan393 (approved):
Looks fine to me -- don't know what's going on with the CI though

PR #301, opened 2023-11-01
TM download bug fix

meisonlikesicecream:
TM was crashing because the emData/PD folders were not created by the download.sh anymore. Fixed the bug, and also removed the allow failure in the CI to avoid similar things happening again (the TM tb is always passing anyway).


aehart (approved):
Seems fine to me. Approved!

PR #302, opened 2023-11-07
MP timing improvement with updated projseq comparison

jasonfan393:
PR changes the projseqs calculation in MP, doing the comparison in one stage instead of two. This greatly improves the HLS estimated timing, with a small improvement in place&route timing.

Timing/resources:
image

### DIFF
--- a/TrackletAlgorithm/MatchProcessor.h
+++ b/TrackletAlgorithm/MatchProcessor.h
@@ -1337,23 +1337,20 @@ void MatchProcessor(BXType bx,
     std::cout << std::endl;
     */
 
-     
-    //New code
+    //New code -- updated to reduce timing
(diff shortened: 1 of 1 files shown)


bryates (approved):
This is a simple change, and gives a nice improvement to the timing. Do we know if this helps the timing in the full project?

jasonfan393:
> This is a simple change, and gives a nice improvement to the timing. Do we know if this helps the timing in the full project?

unfortunately this takes many hours to run, but I am testing this now. This does reduce the maximum logic levels among MP related paths from 13->11 (among a handful I checked), so it should help.

jasonfan393:
timing  is indeed improved on combined barrel config chain, though not passing. I'll go ahead and merge this then
old:
image

new:
image

PR #303, opened 2023-11-08
Update to TP disk/overlap pt lut reading for timing

jasonfan393:
PR updates the pt LUT reading in disk/overlap TP,  using ternary operators instead of a switch. This brings timing for overlap TPs more in line with barrel TPs.

HLS timing estimate for L1D1C: 4.596ns -> 3.465ns.

### DIFF
--- a/TrackletAlgorithm/TrackletProcessor.h
+++ b/TrackletAlgorithm/TrackletProcessor.h
@@ -1129,33 +1129,22 @@ teunits[k].idle_;
 
         (ptinnerindexfirst,ptinnerindexlast) = (idphitmp,ir,innerbend);
         (ptouterindexfirst,ptouterindexlast) = (idphitmp,ir,outerbend);
-        switch(ptinnerindexfirst){ //pt LUTs split into 4 since vivado has a 1024 bit limit for csynth, FIXME better way to do this?
-          case 0:
-          lutinner = teunits[k].stubptinnerlutnew1_[ptinnerindexlast];
-          break;
-          case 1:
-          lutinner = teunits[k].stubptinnerlutnew2_[ptinnerindexlast];
-          break;
-          case 2:
-          lutinner = teunits[k].stubptinnerlutnew3_[ptinnerindexlast];
-          break;
-          case 3:
-          lutinner = teunits[k].stubptinnerlutnew4_[ptinnerindexlast];
-          break;
+
+        if (diskSeed){ //for disk/overlap seeds, pt luts are split into 2/4 parts, due to vivado_hls limit on array sizes
+          lutinner = (ptinnerindexfirst == 0) ? teunits[k].stubptinnerlutnew1_[ptinnerindexlast] :
+                     teunits[k].stubptinnerlutnew2_[ptinnerindexlast];
+          lutouter = (ptouterindexfirst == 0) ? teunits[k].stubptouterlutnew1_[ptouterindexlast] :
+                     teunits[k].stubptouterlutnew2_[ptouterindexlast];
         }
-        switch(ptouterindexfirst){
-          case 0:
-          lutouter = teunits[k].stubptouterlutnew1_[ptouterindexlast];
-          break;
-          case 1:
-          lutouter = teunits[k].stubptouterlutnew2_[ptouterindexlast];
-          break;
-          case 2:
-          lutouter = teunits[k].stubptouterlutnew3_[ptouterindexlast];
-          break;
-          case 3:
-          lutouter = teunits[k].stubptouterlutnew4_[ptouterindexlast];
-          break;
+        else{
+          lutinner = (ptinnerindexfirst == 0) ? teunits[k].stubptinnerlutnew1_[ptinnerindexlast] :
+                     (ptinnerindexfirst == 1) ? teunits[k].stubptinnerlutnew2_[ptinnerindexlast] :
+                     (ptinnerindexfirst == 2) ? teunits[k].stubptinnerlutnew3_[ptinnerindexlast] :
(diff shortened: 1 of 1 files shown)


TrackletAlgorithm/TrackletEngineUnit.h
```diff
@@ -15,8 +15,8 @@ class TrackletEngineUnit {
     kNBitsRZFine=3,
     kNBitsPhiBins=3,
     kNBitsNegDiskSize=1,
-    kNBitsPTLutInner=(Seed==TF::L5L6||TF::L1D1||TF::L2D1||TF::D1D2||TF::D3D4)?1024:(Seed==(TF::L1L2||TF::L2L3)?256:512),
-    kNBitsPTLutOuter=(Seed==TF::L5L6||TF::L1D1||TF::L2D1||TF::D1D2||TF::D3D4)?1024:(Seed==(TF::L1L2||TF::L2L3||TF::L3L4)?256:512)
+    kNBitsPTLutInner=(Seed==TF::L5L6||TF::L1D1||TF::L2D1||TF::D1D2||TF::D3D4)?1024:(Seed==(TF::L1L2||TF::L2L3||TF::L3L4)?256:512),
+    kNBitsPTLutOuter=(Seed==TF::L5L6||TF::L1D1||TF::L2D1||TF::D1D2||TF::D3D4)?1024:(Seed==(TF::L1L2||TF::L2L3)?256:512)
```
aehart, line 19: @jasonfan393 This doesn't make any sense, even before this PR. I think what these lines are trying to do is what I now have in #304 here (another argument in favor of adding the CXXFLAGS in that PR):
https://github.com/cms-L1TK/firmware-hls/blob/e785bf0b8565e4ae0a2a907a0cf1842cc3f9c620/TrackletAlgorithm/TrackletEngineUnit.h#L18-L19

The equality test has to be done for each seed value, instead of just logically ANDing the values together.

TrackletAlgorithm/TrackletEngineUnit.h
```diff
@@ -55,8 +55,18 @@ class TrackletEngineUnit {
 
     const ap_uint<1>* stubptinnertmp = getPTInnerLUT<Seed,iTC>();
     const ap_uint<1>* stubptoutertmp = getPTOuterLUT<Seed,iTC>();
-
-    if ( Seed <= TF::L5L6 ){
+    if ( Seed == TF::L3L4 ){//Split up Outer L3L4 table - reduced LUT usage for some reason 
+      for(unsigned int i=0;i<kNBitsPTLutInner;i++) {
+#pragma HLS unroll
+        stubptinnerlutnew_[i] = stubptinnertmp[i];
```
aehart, line 61: @jasonfan393 Since this part is the same for all seeds ≤ L5L6, could you please rearrange the if statements to slightly reduce duplicate code? So something like this:
```c++
    if ( Seed <= TF::L5L6 ){
      for(unsigned int i=0;i<kNBitsPTLutInner;i++) {
#pragma HLS unroll
        stubptinnerlutnew_[i] = stubptinnertmp[i];
      }
      if ( Seed == TF::L3L4){//Split up Outer L3L4 table - reduced LUT usage for some reason
        for(unsigned int i=0;i<kNBitsPTLutOuter;i++) {
#pragma HLS unroll
          stubptouterlutnew1_[i] = stubptoutertmp[i];
          stubptouterlutnew2_[i] = stubptoutertmp[i+256];
        }
      }
      else {
        for(unsigned int i=0;i<kNBitsPTLutOuter;i++) {
#pragma HLS unroll
          stubptouterlutnew_[i] = stubptoutertmp[i];
        }
      }
    }
```

TrackletAlgorithm/TrackletProcessor.h
```diff
@@ -1129,40 +1129,38 @@ teunits[k].idle_;
...
       else{ //Barrel Seeds
         auto ptinnerindex = (idphitmp, innerbend);
-        auto ptouterindex = (idphitmp, outerbend);
         lutinner = teunits[k].stubptinnerlutnew_[ptinnerindex];
-        lutouter = teunits[k].stubptouterlutnew_[ptouterindex];
+
+        auto ptouterindex = (idphitmp, outerbend);
+        if (Seed == TF::L3L4){ //hack to get HLS to not use 20000 LUTs on outerptlut array
+          ap_uint<1> ptouterindexfirst;
+          ap_uint<8> ptouterindexlast;//FIXME magic numbers
```
aehart, line 1157: @jasonfan393 I would suggest adding 8 to `TrackletEngineUnit::BitWidths`, calling it something like `kNBitsPTLutAddr`. Then you can add a typedef of `ap_uint<kNBitsPTLutAddr>` to `TrackletEngineUnit` and just use that here. Voilà, no more magic.

TrackletAlgorithm/TrackletEngineUnit.h
aehart: @jasonfan393 P.S.: It's hard to imagine these lines are currently doing the right thing in all cases, so it's probably worth rechecking the resource utilization and timing after fixing them.

jasonfan393: yeah I'll fix, I had no idea how this was not causing simulation failures before when I looked at it

jasonfan393: Looks like this is probably the root cause of the weird LUT numbers - I'm testing all modules right now, but I will probably remove the L3L4 hack if all is well(im still not sure why this worked at all)

aehart (approved):
@jasonfan393 Looks good. Approved!

PR #305, opened 2023-11-16
Removing some LUTs

bryates:
Looking into the large number of LUTs @jasonfan393 pointed out, I tried to change some of our LUT arrays to simple functions. This seems to greatly improve the post-implementation timing and reduce the number of LUTs. I wanted to run the full vivado simulation for the `ReducedCombinedConfig` and if all looks good, I can do this with other functions as well.

```
CLB:            725
LUT:           2822
FF:            3973
DSP:              2
BRAM:             0
SRL:              1
URAM:             0
#=== Final timing ===
CP required:    4.000
CP achieved post-synthesis:    3.188
CP achieved post-implementation:    3.195
Timing met
```


jasonfan393 (commented):
overall looks good, seems to improve timing in full project back to previous state a bit. Also HLS seems to be a lot less confused about which resources to use with these versions of the LUTs.

image

jasonfan393 (approved):
Latest version has (roughly) the original timing improvement. Not sure why HLS seems to do better with the two different versions of the empty/nearfull functions but we can investigate further later.
image

PR #306, opened 2023-11-28
Add alternate TP LUT reader for SW compile time/memory

jasonfan393:
PR (developed with the help of @mlarson02 ) adds an alternate way for the c simulation and future SW to read large LUT arrays in the TP, which have been causing long compile times and memory usage.

For a single L1D1 TP csim compile time goes 190 -> 76 seconds. Memory usage is reduced from ~13 Gb to ~4 Gb, with similar improvements in the future SW.

__SYNTHESIS__ macro is used to avoid this file reader for synthesis tasks


jasonfan393:
CI failing due to issue in TC - fixing now

jasonfan393:
currently has some issues related to CMSSW side, don't merge yet

TrackletAlgorithm/SWLUTReader.h
```diff
@@ -0,0 +1,31 @@
+#include <fstream>
+#include <string>
+// The following function provides a more compiler efficient method for reading
+//look up table arrays (currently used in TP), this is only used in C simulations and not synthesis
+template<class lutType, int lutsize> bool readSWLUT(lutType lut[lutsize], std::string lutpath, bool skipfirst = true, bool hex = false){
+
+  //std::cout << "I'm reading : "<< lutpath <<std::endl;
+
+  std::string emdatapath =
+#ifndef __CSIM__ //at run time the working directory differs between CSIM and future SW
```
tomalin, line 10: Using "ifndef" here is unnecessarily obscure. Use "ifdef" instead and reverse the order of the two lines inside the if else endif statement.

tomalin:
@jasonfan393 tihis is failing PR because funtion getInvtLUT is not defined inside TrackletCalculator_calculate_LXLY.h. I suspect the latter file forgot a #include for TrackletProcessor_parameters.h, where this function is defined?

TrackletAlgorithm/SWLUTReader.h
```diff
@@ -0,0 +1,29 @@
...
+// look up table arrays (currently used in TP), this is only used in C simulations and not synthesis
+template<class lutType, int lutsize> bool readSWLUT(lutType lut[lutsize], std::string lutpath, bool skipfirst = true, bool hex = false){
+  //std::cout << "I'm reading : "<< lutpath <<std::endl;
+  std::string emdatapath =
+#ifdef __CSIM__ //at run time the working directory differs between CSIM and future SW
+  "../../../../../emData/";
+#else
+  "../../../../../firmware-hls/emData/";
+#endif
+  std::string fullpath = emdatapath + lutpath;
```
aehart, line 13: @jasonfan393 Please use `const` and references when appropriate:
```suggestion
template<class lutType, int lutsize> bool readSWLUT(lutType lut[lutsize], const std::string& lutpath, bool skipfirst = true, const bool hex = false){
  //std::cout << "I'm reading : "<< lutpath <<std::endl;
  const std::string emdatapath =
#ifdef __CSIM__ //at run time the working directory differs between CSIM and future SW
  "../../../../../emData/";
#else
  "../../../../../firmware-hls/emData/";
#endif
  const std::string fullpath = emdatapath + lutpath;
```

TrackletAlgorithm/TrackletCalculator.h
```diff
@@ -8,7 +8,9 @@
 #include "AllStubMemory.h"
 #include "TrackletParameterMemory.h"
 #include "TrackletProjectionMemory.h"
-
+#ifndef __SYNTHESIS__
+#include "SWLUTReader.h"
+#endif
```
aehart, line 13: @jasonfan393 Since you also include this header file in TrackletProcessor_parameters.h, do you really need it here?

TrackletAlgorithm/SWLUTReader.h
```diff
@@ -0,0 +1,29 @@
+#include <fstream>
+#include <string>
+// The following function provides a more compiler efficient method for reading
+// look up table arrays (currently used in TP), this is only used in C simulations and not synthesis
+template<class lutType, int lutsize> bool readSWLUT(lutType lut[lutsize], std::string lutpath, bool skipfirst = true, bool hex = false){
+  //std::cout << "I'm reading : "<< lutpath <<std::endl;
+  std::string emdatapath =
+#ifdef __CSIM__ //at run time the working directory differs between CSIM and future SW
```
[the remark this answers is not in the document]
jasonfan393: No reason, I did not know about this macro I will swap this out

emData/generate_TP.py
```diff
@@ -123,6 +166,13 @@ class ProjoutIndexDisk(Enum):
       "#ifndef TopFunctions_TrackletProcessor_parameters_h\n"
       "#define TopFunctions_TrackletProcessor_parameters_h\n"
       "\n"
+      "#ifndef __SYNTHESIS__\n"
+      '#include "SWLUTReader.h"\n'
+      '#ifdef CMSSW_GIT_HASH\n'
+      '#include <mutex>\n'
+      'std::mutex getLUTMutex_;\n'
```
tomalin, line 173: Rather than declaring "std::mutex getLUTMutex_;" as a global variable in the header file, which makes it hard to restrict its scope, I'd recommend declaring it as a static variable inside the function where it is used. i.e. getLUT(). See https://stackoverflow.com/questions/14106653/are-function-local-static-mutexes-thread-safe .

TrackletAlgorithm/SWLUTReader.h
```diff
@@ -0,0 +1,34 @@
+#include <fstream>
+#include <string>
+
+// Define a global mutex for locking multi-threading
```
tomalin, line 4: This comment "Define a global mutex" seems in the wrong place. Nothing is defined here.

emData/generate_TP.py
```diff
@@ -65,6 +65,49 @@ class ProjoutIndexDisk(Enum):
     D5PHID = 19
     N_PROJOUT_DISK = 20
 
+#FIXME magic numbers for array sizes:
```
tomalin, line 68: The only thing that troubles me in this PR is the large number of hard-wired constants to specify the length of the LUT tables. 
One way to avoid this would be for https://github.com/cms-L1TK/firmware-hls/blob/SWCompileUpdate/emData/generate_TP.py#L271 , to read the .tab file and determine its length, just before line 271 etc. where it needs the hard-wired length.
(Alternatively, maybe Anders or someone knows LUT table length from first principles)

jasonfan393: added code that reads file lengths from .tab files.

these LUT sizes are pretty hard-coded in many other places in the firmware, but this will hopefully be addressed if we can figure out how to calculate them within the HLS repo.

tomalin (approved):
approved

PR #304, opened 2023-11-09
CXXFLAGS from CMSSW

aehart:
This PR adds the CXXFLAGS from CMSSW_12_6_0_pre5, the current release used for the future SW emulation, to the default flags used for C-simulation. The flags were modified in the following ways:
- Some flags are not recognized by the version of GCC that is used by Vivado HLS 2019.2, so these were removed.
- `-Wextra` is triggered by header files from Xilinx, so this was removed.
- Since HLS uses labels in its C-synthesis reports, `-Wno-unused-label` was added.

PR #309, opened 2023-12-13
SW/FW synchronization

aehart:
This PR synchronizes the FW with the SW emulation at [fw_synch_231205](https://github.com/cms-L1TK/cmssw/tree/fw_synch_231205).

The opportunity is also taken to purge the repo of all non-combined modules that are now irrelevant and to add phi regions to the seed stub indices in the TPAR and TF memory classes.

Ideally, we should merge #304 first (DONE!), so that the HEAD of the master branch can be tagged as the last working version with non-combined modules before merging this PR.


aryd (approved):
Most of this is removal of obsolete code - the other changes looked fine to me.

Issue #307, opened 2023-12-05
DR wishes an extra done signal from TB


tschuh:
A DR which selects the best duplicate to keep instead of the first requires an additional signal from TB to save latency. This signal should tell when the last track has been processed. Ideally that signal is a single bit which is always low and only high for a single clock tick, either one clock tick after or at the same tick when the last track (which might be invalid)  appears at TB output.

PR #312, opened 2023-12-21
TrackBuilder updates

aehart:
This PR makes three updates to the TrackBuilder:
- The TrackBuilder now outputs something on every iteration. Previously, if the track that was built had too few matches, nothing would be written in the C++. Now, an invalid track is written, thus simulating the gaps in valid output that are present in the FW.
- A "done" signal has been added, which goes high for one clock cycle when no more full matches are available for processing. Hopefully, this addresses #307.
- The `latency` pragma has been added to the main processing loop. This is so that the latency of each TrackBuilder is the same, which I believe simplifies integration into chains with multiple TrackBuilders.

The latency and pipelining of the L1L2 TrackBuilder is unchanged, while the minimum clock period reported by HLS improves slightly from 3.486 ns to 3.420 ns. The LUT usage decreases slightly, while the FF usage increases:
|   | L1L2 before | L1L2 after |
| ------------- | ------------- | ------------- |
| BRAM_18K | 0 | 0 |
| DSP48E | 0 | 0 |
| FF | 7841 | 10908 |
| LUT | 35359 | 34445 |
| URAM | 0 | 0 |


TrackletAlgorithm/TrackBuilder.h
```diff
@@ -292,46 +301,44 @@ void TrackBuilder(
...
+          break;
+        case 1:
+          barrelStubWords[j][nTracks] = track.template getBarrelStubWord<1>();
+          break;
+        case 2:
+          barrelStubWords[j][nTracks] = track.template getBarrelStubWord<2>();
+          break;
+        case 3:
+          barrelStubWords[j][nTracks] = track.template getBarrelStubWord<3>();
+          break;
```
jasonfan393, line 322: Just curious, and not something that needs addressing in this PR necessarily but Is there a reason why `getBarrelStubWord` (and others like it defined in TrackFitMemory) have  `Hit` as a template instead of function parameter? is it  possible to have this loop(and others like it) just do something like: 
`barrelStubWords[j][nTracks] = track.template getBarrelStubWord(j);` without the switch?

emData/project_generation_scripts
jasonfan393, line 1: Could these changes be put onto a PR in [project_generation_scripts]( https://github.com/cms-L1TK/project_generation_scripts)?

aehart: Yes, and done (cms-L1TK/project_generation_scripts#56). I was just waiting for cms-L1TK/project_generation_scripts#55 to be merged first.

TrackletAlgorithm/TrackBuilder.h
aehart: I like to be as explicit as possible about what is known at compile time and make those variables template parameters. Here, since the loop is unrolled, each of the "hit" numbers is known at compile time.

It's entirely possible that HLS is smart enough to infer that these numbers are known at compile time, so that if they were passed as regular function arguments, they would be treated the same as they are now. But that would require reworking quite a bit of TrackFitMemory. So it might result in cleaner code, but I'm not sure it's worth trying right now.

jasonfan393: Ok, that makes sense, let's not worry about this now then.

jasonfan393 (approved):

PR #313, opened 2024-01-11
Update GetBitStr Functions for CMSSW

carriganm95:
This PR adds functions to print memories as a bit string for the 

- TrackletParameter
- VMStubeTEOuter
- AllStubInner
- TrackFit

These functions are used by the CMSSW emulation to write out txt files for memories.

### DIFF
--- a/TrackletAlgorithm/AllStubInnerMemory.h
+++ b/TrackletAlgorithm/AllStubInnerMemory.h
@@ -245,6 +245,19 @@ public:
     data_.range(kASNegDiskMSB,kASNegDiskLSB) = negdisk;
   }
 
+#ifdef CMSSW_GIT_HASH
+  std::string getBitStr() const {
+    std::string str = decodeToBits(getR());
+    str += "|"+decodeToBits(getZ());
+    str += "|"+decodeToBits(getPhi());
+    str += "|"+decodeToBits(getBend());
+    if (ASType == DISKPS || ASType == DISK2S) str += "|"+decodeToBits(getNegDisk());
+    str += "|"+decodeToBits(getIndex());
+    str += "|"+decodeToBits(getFinePhi());
+    return str;
+  }
+#endif
+
 private:
 
   AllStubInnerData data_;
--- a/TrackletAlgorithm/AllStubInnerMemory.h
+++ b/TrackletAlgorithm/AllStubInnerMemory.h
@@ -3,6 +3,7 @@
 
 #include "Constants.h"
 #include "MemoryTemplate.h"
+#include "globalFunctions.h"
 
 // AllStubInnerBase is where we define the bit widths, which depend on the class template parameter
 template<int ASType> class AllStubInnerBase {};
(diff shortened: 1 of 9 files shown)


TrackletAlgorithm/AllStubInnerMemory.h
```diff
@@ -245,6 +245,19 @@ class AllStubInner : public AllStubInnerBase<ASType>
     data_.range(kASNegDiskMSB,kASNegDiskLSB) = negdisk;
   }
 
+#ifdef CMSSW_GIT_HASH
+  std::string getBitStr() const {
+    std::string str = decodeToBits(getR());
```
aehart, line 250: @carriganm95 Do you know why you don't need to include globalFunctions.h in this file? I don't understand how the compiler knows what `decodeToBits` is without such an include.

Same question for TrackFitMemory.h and VMStubTEOuterMemoryCM.h.

TrackletAlgorithm/Constants.h
```diff
@@ -19,6 +19,8 @@ constexpr int kNTEUnitsLayerDisk[] = {0, 5, 2, 5, 0, 3, 3, 3, 0, 2, 0}; // Numbe
 // The outer layerDisk of the seed type is used as index, e.g. kNTEUnits[TF::L1L2] == kNTEUnitsLayerDisk[TF::L2]. N.B.: D1 actually has two seed types (L1D1, L2D1), take the largest value for now...
 constexpr int kNbitsrzbin = 3;
 constexpr int kNbitsphibin = 3;
+constexpr unsigned kNbitsrzbinMP = kNbitsrzbin + 1;
+constexpr unsigned kNbitsphibinMP = kNbitsphibin + 1;
```
aehart, line 23: @carriganm95 I don't see these used anywhere. Are they needed?

TrackletAlgorithm/VMStubMEMemoryCM.h
```diff
@@ -111,6 +111,14 @@ class VMStubMECM : public VMStubMECMBase<VMSMEType>
     return data_.range(kVMSMEBendMSB,kVMSMEBendLSB);
   }
 
+  /*bool isPSStub() const {
+    if(VMSMEType == BARRELPS) return true;
+    else if(VMSMEType == BARREL2S) return false;
+    else if(VMSMEType == DISK) return (getBend() & (1 << (VMStubMECMBase<VMSMEType>::kVMSMEBendSize-1))) == 0 && getBend() != 0; // Check highest 1 bits regardless of template type
+    static_assert(true, "Should not be possible!");
+  }*/
```
aehart, line 119: @carriganm95 What's the purpose of adding this as a comment?

TrackletAlgorithm/AllStubInnerMemory.h
carriganm95: yeah I am not sure, TrackletParameter memory complained when I did not add it. My guess is that we have loaded other memory types in those Analyzers so that it already knows about globalFunctions? Either way I will add it to be safe.

TrackletAlgorithm/VMStubMEMemoryCM.h
carriganm95: sorry, I was removing this from a branch I copied from and forgot to delete. Should be fixed now

TrackletAlgorithm/Constants.h
carriganm95: same as below, I copied from an old branch. They should be gone now

TrackletAlgorithm/AllStubInnerMemory.h
carriganm95: I have added the include statements

aehart: > My guess is that we have loaded other memory types in those Analyzers so that it already knows about globalFunctions?

Yeah, that could be. In any case, I think it's good to include globalFunctions.h in each header file so you aren't dependent on how these are included in the analyzers.

aehart (approved):
@carriganm95 Looks good now. Approved!

PR #321, opened 2024-02-23
Update Get Bit Str Functions

carriganm95:
This PR updates two getBitStr functions for the CMSSW future SW.

This adds the stub R to the full match function and phi to TrackletParameter function.


aehart (approved):
Looks good. Approved!

PR #323, opened 2024-02-28
VMRCM - Number of AllStubInner memories fix

meisonlikesicecream:
The number of AllStubInner memories could previously not be set to 0 or the HLS synthesis would crash, so it was set to 1 even though no AllStubInner memories were present in some modules. @tomalin implicitly requested a fix for this so here it is :-)


tomalin (approved):
Thank's Mei, that does the trick for the Future SW. And seems harmless for the FW. Approved.

PR #324, opened 2024-02-28
Remove Memory copies in CMSSW

tomalin:
MemoryTemplateBinnedCM creates NCOPY identical copies of each memory. Whilst needed in FW, these are a waste of memory in CMSSW. The change to MemoryTemplateBinnedCM proposed here, (which should only affect CMSSW use), makes it pretend to the outside world that it has NCOPY copies, whilst internally it actually only has a single copy. I've checked that the Future SW is unaffected by this. (Though I believe this change is needed to allow it to be migrated to CMSSW 14).
P.S. Also made a small change to MemoryTemplate.h, to stop NBIT_BX appearing in the global space when running within CMSSW. It is now contained inside the class.

### DIFF
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -14,8 +14,7 @@
 #endif
 
 #ifdef CMSSW_GIT_HASH
-#define NBIT_BX 0
-template<class DataType, unsigned int DUMMY, unsigned int NBIT_ADDR, unsigned int NBIT_BIN, unsigned int kNBitsphibinCM, unsigned int NCOPY>
+template<class DataType, unsigned int DUMMYA, unsigned int NBIT_ADDR, unsigned int NBIT_BIN, unsigned int kNBitsphibinCM, unsigned int NCOPY>
 #else
 template<class DataType, unsigned int NBIT_BX, unsigned int NBIT_ADDR, unsigned int NBIT_BIN, unsigned int kNBitsphibinCM, unsigned int NCOPY>
 #endif
@@ -27,7 +26,22 @@ template<class DataType, unsigned int NBIT_BX, unsigned int NBIT_ADDR, unsigned 
 // (1<<NBIT_ADDR): depth of the memory for each BX
 // NBIT_BIN: number of bits used for binning; (1<<NBIT_BIN): number of bins
 // NCOPY: is the number of memory copies - same as number of TE used
+
+// N.B. The CMSSW version of this pretends to have NCOPY copies to anyone 
+// using this class, but doesn't actually instantiate them, to save memory.
+
 class MemoryTemplateBinnedCM{
+
+ private:
+
+#ifdef CMSSW_GIT_HASH
+  static constexpr bool isCMSSW = true;
+  static constexpr unsigned int NBIT_BX = 0;
+  static constexpr unsigned int NCP = 1;
+#else
+  static constexpr bool isCMSSW = false;
+  static constexpr unsigned int NCP = NCOPY;
+#endif
 
  public: 
   typedef ap_uint<NBIT_BX> BunchXingT;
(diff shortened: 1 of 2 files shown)


TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -27,8 +26,23 @@ template<class DataType, unsigned int NBIT_BX, unsigned int NBIT_ADDR, unsigned
...
 
+ private:
+
+#ifdef CMSSW_GIT_HASH
+  static constexpr bool isCMSSW = true;
+  static constexpr unsigned int NBIT_BX = 0;
+  static constexpr unsigned int NCP = 1;
+#else
+  static constexpr bool isCMSSW = false;
+  static constexpr unsigned int NCP = NCOPY;
```
aehart, line 43: @tomalin Since you define `NCP` here, I would suggest using it throughout instead of `NCOPY`. In particular, here:
https://github.com/cms-L1TK/firmware-hls/blob/2d0110f714d902afe1735422c3dd9066216da94d/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L74
and here:
https://github.com/cms-L1TK/firmware-hls/blob/2d0110f714d902afe1735422c3dd9066216da94d/TrackletAlgorithm/MemoryTemplateBinnedCM.h#L142

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -119,40 +133,42 @@ class MemoryTemplateBinnedCM{
...
+    if (isCMSSW) icopy = 0;
     return dataarray_[icopy];
   }
 
 
+#ifndef CMSSW_GIT_HASH
   const DataType (&get_mem() const)[NCOPY][1<<NBIT_BX][1<<NBIT_ADDR] {
     return dataarray_;
   }
+#endif
```
aehart, line 145: @tomalin If `NCOPY` is replaced with `NCP`, does that allow you to remove the `#ifndef` protection?
```suggestion
  const DataType (&get_mem() const)[NCP][1<<NBIT_BX][1<<NBIT_ADDR] {
    return dataarray_;
  }
```

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -14,8 +14,7 @@
 #endif
 
 #ifdef CMSSW_GIT_HASH
-#define NBIT_BX 0
-template<class DataType, unsigned int DUMMY, unsigned int NBIT_ADDR, unsigned int NBIT_BIN, unsigned int kNBitsphibinCM, unsigned int NCOPY>
+template<class DataType, unsigned int DUMMYA, unsigned int NBIT_ADDR, unsigned int NBIT_BIN, unsigned int kNBitsphibinCM, unsigned int NCOPY>
```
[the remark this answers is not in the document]
tomalin: My first attempt at this code had two dummy template parameters. But you're right. I'll change it back to DUMMY.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
tomalin: To play safe, I want the CMSSW version of this class to look identical to the outside World to the non-CMSSW version. If function getNCopy() returned NCP, which is 1 when running in CMSSW and NCOPY when running outside CMSSW, this requirement would not be met. The danger is that other HLS modules might call getNCopy when running in CMSSW, and behave strangely because they get the answer 1. e.g. If downstream modules use getNCopy to determine the number of parallel modules that can process the data from the memory, then this number might be set to 1 in CMSSW, leading to greater truncation.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
tomalin: The code would compile if I made your proposed change.. But this function would behave differently in CMSSW and non-CMSSW, which would potentially lead to different L1 tracking results inside and outside CMSSW. So I prefer to protect this function.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
aehart: Thank you for explaining. I understand now.

aehart (approved):
Looks good. Approved!

PR #328, opened 2024-03-05
Change FileRead function to inline

sarafiorendi:
The return_splitted function in FileReadUtility.h is declared inline. This small change is needed in order to make the Future Software simulation code compile, and does not change anything on the firmware side.


tomalin (approved):
approved

PR #335, opened 2024-04-12
Future SW CMSSW Fix

carriganm95:
Remove the ASType dependence from the getBitStr function in the new AllStubInner DISKPS type


aehart (approved):
Looks good. Approved!

PR #336, opened 2024-04-15
TrackBuilder that reads from MPAR memories

aehart:
This PR is an update to the TrackBuilder that reads tracklet parameters from the new MPAR memories.


aryd (approved):
Looks fine to me (I did not yet look at the changes in project_generation_scripts)

PR #338, opened 2024-04-19
First FPGA with TPAR/AS merging

jasonfan393:
PR to prepare for changes to sectorprocessor on FPGA 1 that enables the merging and streaming of TPAR and AS words to FPGA2. Interface of sector processor is changed from a read of BRAMs containing those words to a FIFO-like output stream.

Opening as a draft to check CI and work out merging.

some bugs in simulation right now, AllStubs not streaming properly and some extra truncation.


TestBenches/TrackletProcessor_test.cpp
```diff
@@ -128,7 +128,7 @@ std::cout<<module_name[MODULE_];
   std::cout << "Loaded the input files:\n";
   for (unsigned i = 0; i < nInnerStubMems; i++)
    //(i+2)%nInnerStubMems has the innerstubs match the order in the emulation
```
aehart, line 130: @jasonfan393 I think this comment is obsolete now. Or maybe it should be moved to TrackletProcessor.h?
```suggestion
```

aehart (approved):
Approved!

PR #339, opened 2024-05-10
CI tests and code style fixes

jasonfan393:
draft PR for testing some alterations to CI, like adding more modules to test csynth over.

Also code style fixes to generate_ scripts


jasonfan393:
currently increases CI run-times significantly, will look further into this some other time.

PR #334, opened 2024-04-09
FPGA2 Project

aryd:
This is a first (draft) PR for the code for the second FPGA. It includes a project that uses the VMSMER and the PC modules, and the new 'merged' MPAR and MPROJ memories.

PR #340, opened 2024-05-21
MP disk II fixes

jasonfan393:
PR fixes a few pragmas related to II issues in the MP disks. MP_D5 now meets II = 2, same as the other MPs. There is still an II violation related to the calculations of best_delta_rphi, but this will require more reworking such as is in PR [334](https://github.com/cms-L1TK/firmware-hls/pull/334)


aehart (approved):
Looks pretty straightforward. Approved!

PR #341, opened 2024-06-21
Number of events for PC and VMSMER

aehart:
Currently, only ten events are used in the test benches for the PC and VMSMER for both C-simulation and C/RTL cosimulation because the latter takes a very long time to run. This PR just updates the number for C-simulation to 100, as are used in the test benches for the other modules.


aryd (approved):
Looks OK - but ideally we should figure out why the cosim is slow so that we can run the full 100 events all the time.

PR #342, opened 2024-07-03
Various optimizations for timing

aryd:
This PR has a number of optimizations to improve timing:

1) Reducation of memories used by the TPROJ memories by changing the page size the 64 entiries

2) Replace BRAM with URAM for the TPAR and VMSME memories

3) Better floor plan

4) Adding buffering/pipelin stage in the TrackBuilder

5) Add pipelining stage in the MatchProcessor before calling the MatchCalculator. 

6) Various timing optimizations in the tf_mem_XYZ.vhd files

There is one more timing improvement in MatchEngineUnit.h in the getProjSeq method. But this code breaks (more significantly) the agreement with the emulation so I have left it commented out.

PR #343, opened 2024-07-05
EMP FPGA2

mcoshiro:
Adds EMP integration for dual FPGA project including working second FPGA project. The first FPGA project will not work until its branch is also merged into master. Includes

  * New DualFPGA folder in integration test with all 2 FPGA project specific EMP code
  * Small changes to common emp code in IntegrationTests/common
  * Small changes to EMP-related files in IntegrationTests/CombinedBarrelConfig and IntegrationTests/ReducedCombinedConfig that were accidentally in common despite being configuration specific

Associated with project_generation_scripts [#63](https://github.com/cms-L1TK/project_generation_scripts/pull/63). Associated KF branch will need to be reconciled with master after DR updates.


IntegrationTests/CombinedBarrelConfig/IRtoKF/firmware/scripts/pre_msim.tcl
aehart, line 1: @mcoshiro We had similar pre-simulation/-synthesis scripts back when the ROMs for the LUTs were included at the VHDL level. But once we embedded these ROMs in the HLS blocks themselves, these scripts were removed.

So is this script now necessary again for some reason, or is it a historical artifact that can be removed?

(Similar question for the following files:
IntegrationTests/CombinedBarrelConfig/IRtoKF/firmware/scripts/lut_copy.py
IntegrationTests/DualFPGA/firmware/scripts/pre_msim.tcl
IntegrationTests/DualFPGA/firmware/scripts/lut_copy.py
IntegrationTests/DualFPGA/firmware/scripts/lut_copy_dual.py
IntegrationTests/ReducedCombinedConfig/IRtoKF/firmware/scripts/pre_msim.tcl
IntegrationTests/ReducedCombinedConfig/IRtoKF/firmware/scripts/lut_copy.py
IntegrationTests/common/script/emp/pre_qsim.tcl
)

IntegrationTests/DualFPGA/firmware/cfg/vsim_f1.dep
```diff
@@ -0,0 +1,17 @@
...
+include -c emp-fwk:boards/testbench top.dep
+
+src tb_decl_f1.vhd
+src emp_project_decl_f1.vhd 
+src -c emp-fwk:components/links/slink emp_slink_types.vhd
+
+src -u sim ../scripts/pre_msim.tcl
+
+include -c emp-fwk:boards/apollo/cm_v2/vu13p device.dep packages_cm_v2_p1.dep
+#include -c emp-fwk:boards/apollo/cm_v2/vu13p cm_v2_p1_2.dep device.dep packages_cm_v2_p1.dep
```
aehart, line 17: @mcoshiro Based on the comment at the top of this file, this line wouldn't work due to unresolved dependencies. So should it be removed, or will it be relevant when we eventually move to TCDS2?

IntegrationTests/DualFPGA/firmware/cfg/vsim_f2.dep
```diff
@@ -0,0 +1,17 @@
...
+include -c emp-fwk:boards/testbench top.dep
+
+src tb_decl_f2.vhd
+src emp_project_decl_f2.vhd 
+src -c emp-fwk:components/links/slink emp_slink_types.vhd
+
+src -u sim ../scripts/pre_msim.tcl
+
+include -c emp-fwk:boards/apollo/cm_v2/vu13p device.dep packages_cm_v2_p2.dep
+#include -c emp-fwk:boards/apollo/cm_v2/vu13p cm_v2_p1_2.dep device.dep packages_cm_v2_p1.dep
```
aehart, line 17: @mcoshiro Based on the comment at the top of this file, this line wouldn't work due to unresolved dependencies. So should it be removed, or will it be relevant when we eventually move to TCDS2?

IntegrationTests/DualFPGA/firmware/hdl/conv_pkg_f1.vhd
aehart, line 1: @mcoshiro Would it make sense/be possible to move the `conv_single` function from this file to IntegrationTests/common/hdl/emp/pkgs/tf_interface_pkg.vhd, which seems to be where similar functions live?

IntegrationTests/CombinedBarrelConfig/IRtoKF/firmware/scripts/pre_msim.tcl
mcoshiro: Thanks for pointing this out. Indeed, I verified that they were indeed historical artifacts and that the EMP simulation runs without them, so these files are now removed

IntegrationTests/DualFPGA/firmware/cfg/vsim_f1.dep
mcoshiro: Indeed, the packages in the commented lines are in general needed for simulation, so I've removed them (in vsim_f1.dep and vsim_f2.dep)

IntegrationTests/DualFPGA/firmware/hdl/conv_pkg_f1.vhd
mcoshiro: This is possible, but I am not sure it is simpler since currently, tf_interface_pkg.vhd has a dependency on memUtil_pkg for the 2nd FPGA project since it defines numTW_104 and numBW_46. The options are thus either to separate conv_single (as is currently done), or alternatively add conv_single to this file and separate the definitions of numTW_104 and numBW_46 into a different file. If you would like, I can switch to the alternative, though to avoid conflicts between the FPGA1 and FPGA2 projects, having two files is unavoidable

aehart: Thanks for the explanation. I thought there might be a dependency reason for structuring it the way you did, and with that background, I'm happy with your current solution.

IntegrationTests/CombinedBarrelConfig/IRtoKF/firmware/scripts/pre_msim.tcl
aehart: I think there's just one last remnant here, where the `lut_inclusion` target needs to be removed:
https://github.com/cms-L1TK/firmware-hls/blob/37a58469d045426c24304b6588afa927410256f9/IntegrationTests/CombinedBarrelConfig/IRtoKF/firmware/Makefile#L53

mcoshiro: Thanks for catching this! I've fixed said line. Just let me know if you find any other issues or have any other suggestions

aehart (approved):
This looks good to me, so I'll approve it.

I don't think there will be any merge conflicts between this PR and #342, but I do want to merge that one first just in case. Hopefully, that can happen today or tomorrow, then I will merge this one as well.

PR #344, opened 2024-07-30
Removed NBIT_BX from global name space

tomalin:
In MemoryTemplateTPROJ.h, the pragma variable NBIT_BX was defined in the global namespace, which led to clashes when this header file was used in the CMSSW code. I've moved it inside the class to avoid this, as is already done in MemoryTemplate.h.


aehart (approved):
Straightforward change with no effect on the firmware. Approved!

PR #347, opened 2024-09-13
Attempt to fix Future L1 Track SW + bugs in HLS memory classes.

tomalin:
The Future L1Trk CMSSW SW has not yet migrated to the 2FPGA solution. I'm therefore trying to bodge it so it continues to work (for now) with the 1FPGA solution.

1) The three functions in TrackBuilder_parameters.h with names getMPAR*() caused the CMSSW compiler to fail because of "duplicate function definitions". I fixed this by declaring them to be "constexpr" or "inline" in generate_TB.py. I also (on style grounds) changed their template parameter from type int to TF::seed.
2) MemoryTemplateBinnedCM::clear() used constant NCOPY, which caused array-out-of-bounds errors in CMSSW, where NCOPY can be zero. Replaced NCOPY with NCP.
3) CMSSW compiler errors about "duplicate code", solved by deleting the calculation of nentries withing CMSSW_GIT_HASH that was in MemoryTemplateBinnedCM & MemoryTemplateBinned & MemoryTemplatePROJ.
4) Added constants DEPTH_BX and DEPTH_ADDR to MemoryTemplate.h, MemoryTemplateBinnedCM.h and MemoryTemplateTPROJ.h, to simplify them. In MemoryTemplateTPROJ.h, replaced magic number 128 by DEPTH_ADDR.
5) Modified MemoryTemplateTPROJ to it instantiates only one page of memory when used inside CMSSW.
6) Replaced argument addr_index of MemoryTemplate::write_mem_new() by argument "ap_uint<1> overwrite", with the same number of bits used by MatchProcessor.h [HERE](https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MatchProcessor.h#L1139) when it calls write_mem_new(). 
Howver, [this code](https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MemoryTemplate.h#L134) inside write_mem_new(), which should protect the memory against having too many entries does nothing, since addr_index is a 1 bit number. This is a BUG. I replaced it by a check that nentries_[ibx] is than than (1<<NBIT_ADDR). However, array access can be slow in FW, so one should check if this causes FW timing errors.
8) Besides the style issue that it's hard-wiring the number of pages to 4, https://github.com/cms-L1TK/firmware-hls/blob/master/TestBenches/FileReadUtility.h#L259 seems to incorrectly assume that MemoryTemplateTPROJ::getDepth() returns the total depth of the memory, whereas it can be seen from https://github.com/cms-L1TK/firmware-hls/blob/master/TrackletAlgorithm/MemoryTemplateTPROJ.h#L49 that it returns a number that is a factor NPAGE less than this. I believe this is a BUG and that compareMemWithMemPage() is only checking one quarter of the memory. Fixed.
9) SWLUTReader.h was writing by 1 unit beyond the array bounds of the input "lut" array, because it neglected that the input .tab files contain a final line with a "close bracket" but no number.  This BUG had horrible consequences, causing all LUT contents to become zero in CMSSW. Now fixed.
10) Modified download.sh, so that when running the CombinedCM chain, the TP takes its LUT tables from LUTsSplit/ instead of LUTsCM/ . This reduces the rate of memory errors, as LUTsCM/ is out-of-date. All other modules in the chain already used LUTsSplit/.
11) To improve clarity of VMRouterCM.h & MatchProcessor.h, renamed nAllInnerCopies to nAllInnerVariants, since the versions of the AllStubsInner memories that are produced are not copies of each other, but are different from each other. Similarly, renamed maxFullMatchCopies to maxFullMatchVariants.

THIS PR NEEDS DISCUSSION WITH @aehart and @aryd BEFORE MERGING.

### DIFF
--- a/TestBenches/FileReadUtility.h
+++ b/TestBenches/FileReadUtility.h
@@ -256,9 +256,9 @@ unsigned int compareMemWithMemPage(const MemType& memory_ref, const MemType& mem
   constexpr int lsb = (LSB >= 0 && MSB >= LSB) ? LSB : 0;
   constexpr int msb = (LSB >= 0 && MSB >= LSB) ? MSB : MemType::getWidth() - 1;
 
-  for (unsigned int ipage = 0; ipage < 4; ++ipage) {
+  for (unsigned int ipage = 0; ipage < memory_ref.getNPage(); ++ipage) {
     //FIXME
-    for (unsigned int i = 0; i < memory_ref.getDepth()/4; ++i) {
+    for (unsigned int i = 0; i < memory_ref.getDepth(); ++i) {
       const auto data_ref_raw = memory_ref.read_mem(ievt,i,ipage).raw();
       const auto data_com_raw = memory.read_mem(ievt,i,ipage).raw();
       const auto data_ref = data_ref_raw.range(msb,lsb);
--- a/TrackletAlgorithm/MemoryTemplateBinnedCM.h
+++ b/TrackletAlgorithm/MemoryTemplateBinnedCM.h
@@ -36,7 +36,7 @@ template<class DataType, unsigned int NBIT_BX, unsigned int NBIT_ADDR, unsigned 
 
 class MemoryTemplateBinnedCM{
 
- private:
+private:
 #ifdef CMSSW_GIT_HASH
   static constexpr bool isCMSSW = true;
   static constexpr unsigned int NBIT_BX = 0;
@@ -46,7 +46,7 @@ class MemoryTemplateBinnedCM{
   static constexpr unsigned int NCP = NCOPY;
 #endif
 
- public:
+public:
   static constexpr unsigned int DEPTH_BX = 1<<NBIT_BX;
   static constexpr unsigned int DEPTH_ADDR = 1<<NBIT_ADDR;
   static constexpr unsigned int DEPTH_BIN = 1<<NBIT_BIN;
@@ -54,7 +54,7 @@ class MemoryTemplateBinnedCM{
   typedef ap_uint<NBIT_BX> BunchXingT;
   typedef ap_uint<NBIT_ADDR-NBIT_BIN> NEntryT;
 
- protected:
+protected:
   enum BitWidths {
     kNBxBins = DEPTH_BX,
     kNSlots = DEPTH_BIN,
@@ -69,7 +69,7 @@ class MemoryTemplateBinnedCM{
   ap_uint<8> binmask8_[kNBxBins][1<<kNBitsRZBinCM];
   ap_uint<64> nentries_[slots];
 
- public:
+public:
 
   unsigned int getDepth() const {return kNMemDepth;}
   unsigned int getNBX() const {return kNBxBins;}
@@ -89,7 +89,7 @@ class MemoryTemplateBinnedCM{
   }
 
   ap_uint<8> getBinMask8(BunchXingT bx, ap_uint<kNBitsRZBinCM> ibin) const {
-    #pragma HLS ARRAY_PARTITION variable=binmask8_ complete dim=0
+#pragma HLS ARRAY_PARTITION variable=binmask8_ complete dim=0
     return binmask8_[bx][ibin];
   }
 
@@ -121,7 +121,7 @@ class MemoryTemplateBinnedCM{
   }
   
   DataType read_mem(unsigned int icopy, BunchXingT ibx, ap_uint<NBIT_BIN> slot,
-		    ap_uint<NBIT_ADDR> index) const {
+                    ap_uint<NBIT_ADDR> index) const {
 #pragma HLS ARRAY_PARTITION variable=dataarray_ dim=1
     // TODO: check if valid
     if (isCMSSW) {ibx = 0; icopy = 0;}
@@ -153,24 +153,24 @@ class MemoryTemplateBinnedCM{
 
       if (nentry == ((1 << (NBIT_ADDR-NBIT_BIN)) - 1)) return false;
 
-    	nentries_[ibx*kNBinsRZ+ibin].range(ireg*4+3,ireg*4)=nentry+1;
-    	if (ibin!=0) {
-	      nentries_[ibx*kNBinsRZ+ibin-1].range((ireg+8)*4+3,(ireg+8)*4)=nentry+1;
-    	}
-	    binmask8_[ibx][ibin].set_bit(ireg,true);
+      nentries_[ibx*kNBinsRZ+ibin].range(ireg*4+3,ireg*4)=nentry+1;
+      if (ibin!=0) {
+        nentries_[ibx*kNBinsRZ+ibin-1].range((ireg+8)*4+3,(ireg+8)*4)=nentry+1;
+      }
+      binmask8_[ibx][ibin].set_bit(ireg,true);
 
-    	//icopy comparison must be signed int or future SW fails
-      writememloop:for (signed int icopy=0;icopy< (signed) NCP;icopy++) {
+      //icopy comparison must be signed int or future SW fails
+    writememloop:for (signed int icopy=0;icopy< (signed) NCP;icopy++) {
 #pragma HLS unroll
-	      dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry] = data;
-    	}
+        dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry] = data;
+      }
 #endif      
       return true;
     }
     else {
 #ifndef __SYNTHESIS__
       if (data.raw() != 0) { // To avoid lots of prints when we're clearing the memories
-	edm::LogVerbatim("L1trackHLS") << "Warning out of range. nentry_ibx = "<<nentry_ibx<<" NBIT_ADDR-NBIT_BIN = "<<NBIT_ADDR-NBIT_BIN << std::endl;
+        edm::LogVerbatim("L1trackHLS") << "Warning out of range. nentry_ibx = "<<nentry_ibx<<" NBIT_ADDR-NBIT_BIN = "<<NBIT_ADDR-NBIT_BIN << std::endl;
       }
 #endif
       return false;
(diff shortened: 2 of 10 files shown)


tomalin:
Strange. This fails git CI because of 2FA authentification issues https://github.com/cms-L1TK/firmware-hls/actions/runs/10853460640/job/30121761689?pr=347 .

tomalin:
QUESTIONS/ISSUES:
1) The TrackletProcessor writes TrackletParameterMemory objects that inherit from MemoryTemplateTPROJ, and so have 4 pages of memory, despite the fact that the TrackletProcessor only writes to the first page. Is this wasting memory?
2) MemoryTemplateTPROJ assumes that memory txt files include the "page number". This is true of those in emData/MemPrintsSplit/, but not true of those in emData/MemPrintsCM/. It therefore crashes. Either we need to write txt files in emData/MemPrintsCM in the new format, or to support the 1FPGA solution, and option will need adding to TrackletParameterMemory and TrackletProjectionMemory so that that can be based on MemoryTemplate (as they used to be) instead of MemoryTemplateTPROJ.

tomalin:
Rebased to latest master

TestBenches/FileReadUtility.h
```diff
@@ -256,9 +256,9 @@ unsigned int compareMemWithMemPage(const MemType& memory_ref, const MemType& mem
   constexpr int lsb = (LSB >= 0 && MSB >= LSB) ? LSB : 0;
   constexpr int msb = (LSB >= 0 && MSB >= LSB) ? MSB : MemType::getWidth() - 1;
 
-  for (unsigned int ipage = 0; ipage < 4; ++ipage) {
+  for (unsigned int ipage = 0; ipage < memory_ref.getNPage(); ++ipage) {
     //FIXME
-    for (unsigned int i = 0; i < memory_ref.getDepth()/4; ++i) {
+    for (unsigned int i = 0; i < memory_ref.getDepth(); ++i) {
```
aehart, line 261: @tomalin This works for the SW, where the number of pages is one, but generally, shouldn't the upper bound of the loop still be divided by the number of pages?

```suggestion
    for (unsigned int i = 0; i < memory_ref.getDepth()/memory_ref.getNPage(); ++i) {
```

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -150,30 +153,18 @@ class MemoryTemplateBinnedCM{
...
 
-	//icopy comparison must be signed int or future SW fails
-        writememloop:for (signed int icopy=0;icopy< (signed) NCP;icopy++) {
+    	//icopy comparison must be signed int or future SW fails
+      writememloop:for (signed int icopy=0;icopy< (signed) NCP;icopy++) {
 #pragma HLS unroll
-	  dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry] = data;
-	}
+	      dataarray_[icopy][ibx][getNEntryPerBin()*slot+nentry] = data;
+    	}
```
aehart, line 166: @tomalin These changes just seem to affect the indentation, which was already a bit off, and now looks even more off, at least in my text editor. Could you clean this up, perhaps removing tabs altogether and consistently using two spaces to indent lines inside brackets?

TrackletAlgorithm/MemoryTemplateBinnedCM.h
```diff
@@ -199,10 +190,10 @@ class MemoryTemplateBinnedCM{
...
     DataType data("0",16);
     for (size_t ibx=0; ibx<(kNBxBins); ++ibx) {
-      for (size_t icopy=0; icopy < NCOPY; icopy++) {
-	for (size_t ibin=0; ibin < kNMemDepth; ibin++) {
-	  dataarray_[icopy][ibx][ibin] = data;
-	}
+      for (size_t icopy=0; icopy < NCP; icopy++) {
+       	for (size_t ibin=0; ibin < kNMemDepth; ibin++) {
+	        dataarray_[icopy][ibx][ibin] = data;
+      	}
```
aehart, line 196: @tomalin Same comment regarding indentation.

```suggestion
      for (size_t icopy=0; icopy < NCP; icopy++) {
        for (size_t ibin=0; ibin < kNMemDepth; ibin++) {
          dataarray_[icopy][ibx][ibin] = data;
        }
```

TrackletAlgorithm/MemoryTemplateTPROJ.h
```diff
@@ -68,13 +85,14 @@ class MemoryTemplateTPROJ
...
 
   DataType read_mem(BunchXingT ibx, ap_uint<NBIT_ADDR> index, unsigned int page = 0) const
   {
-	// TODO: check if valid
-	if(!NBIT_BX) ibx = 0;
-	return dataarray_[ibx][(1<<NBIT_ADDR)*page+index];
+    //assert(page < NPAGE);    
+  	// TODO: check if valid  
+	  if(!NBIT_BX) ibx = 0;
+ 	  return dataarray_[ibx][DEPTH_ADDR*page+index];
```
aehart, line 95: @tomalin Same comment regarding indentation.

```suggestion
    //assert(page < NPAGE);
    // TODO: check if valid
    if(!NBIT_BX) ibx = 0;
    return dataarray_[ibx][DEPTH_ADDR*page+index];
```

TrackletAlgorithm/MemoryTemplateTPROJ.h
```diff
@@ -127,21 +142,21 @@ class MemoryTemplateTPROJ
...
       mask_[ibx]=0;
       for (size_t page = 0; page < NPAGE; page++) {
-	nentries_[ibx*NPAGE+page] = 0;
-	for (size_t addr=0; addr<(1<<NBIT_ADDR); ++addr) {
-	  dataarray_[ibx][(1<<NBIT_ADDR)*page+addr] = data;
-	}
+      	nentries_[ibx*NPAGE+page] = 0;
+      	for (size_t addr=0; addr<DEPTH_ADDR; ++addr) {
+	        dataarray_[ibx][DEPTH_ADDR*page+addr] = data;
+      	}
```
aehart, line 151: @tomalin Same comment regarding indentation.

```suggestion
        nentries_[ibx*NPAGE+page] = 0;
        for (size_t addr=0; addr<DEPTH_ADDR; ++addr) {
          dataarray_[ibx][DEPTH_ADDR*page+addr] = data;
        }
```

TrackletAlgorithm/MemoryTemplateTPROJ.h
```diff
@@ -209,7 +210,7 @@ class MemoryTemplateTPROJ
 
   void print_entry(BunchXingT bx, NEntryT index, unsigned int page = 0) const
   {
-	print_data(dataarray_[bx][(1<<NBIT_ADDR)*page+index]);
+  	print_data(dataarray_[bx][DEPTH_ADDR*page+index]);
```
aehart, line 213: @tomalin Same comment regarding indentation.

```suggestion
    print_data(dataarray_[bx][DEPTH_ADDR*page+index]);
```

emData/generate_TB.py
```diff
@@ -122,9 +122,9 @@ class ITC(Enum):
         "template<TF::seed Seed> constexpr int getNumDiskFMMemPerStub();\n"
         "template<TF::seed Seed> constexpr int getNumBarrelStub();\n"
         "template<TF::seed Seed> constexpr int getNumDiskStub();\n"
-        "template<unsigned Seed> int getMPARNPages(const ITCType &);\n"
-        "template<unsigned Seed> int getMPARMem(const ITCType &);\n"
-        "template<unsigned Seed> int getMPARPage(const ITCType &);\n"
+        "template<TF::seed Seed> inline    int getMPARNPages(const ITCType &);\n"
+        "template<TF::seed Seed> inline    int getMPARMem(const ITCType &);\n"
+        "template<TF::seed Seed> inline    int getMPARPage(const ITCType &);\n"
```
aehart, line 127: @tomalin Probably it's better to just make these `constexpr`, as that should replace the function calls with their return values at compile time. IIRC, this would require changing the input argument to something simple like `unsigned` or `unsigned short`, but that shouldn't cause any problems on the FW side.

```suggestion
        "template<TF::seed Seed> constexpr int getMPARNPages(const unsigned short);\n"
        "template<TF::seed Seed> constexpr int getMPARMem(const unsigned short);\n"
        "template<TF::seed Seed> constexpr int getMPARPage(const unsigned short);\n"
```

TrackletAlgorithm/MemoryTemplateBinnedCM.h
tomalin: The problem is that the spaces in the HLS code are often created with "tab", and the number of spaces that "tab" corresponds to is not well-defined. It differs on different computers and in different editors/viewing tools. To fix this, we'd really have to globally replace all tabs by spaces, and then run clang to enforce a sensible indentation. And ask people to stop using tabs when editing the HLS code. But this should wait until there are no outstanding PRs.

emData/generate_TB.py
tomalin: They can't be constexpr. For example:

template<> inline int
getMPARNPages<TF::L1L2>(const ITCType &iTC) {
  return (iTC == 5 || iTC == 6) ? 1 : (
         (iTC == 3 || iTC == 4 || iTC == 7 || iTC == 8) ? 2 : 3);
}

depends on const ITC, whose value is not known at compile time.

TestBenches/FileReadUtility.h
aehart: That makes sense. Thanks for clarifying.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
aehart: I completely agree, but I also think we can clean things up as we change other things. Especially with this block of code in particular, where the only changes that have been made are to the indentation, unless I've missed something.

tomalin: I can do this, but it would mean replacing all tabs with spaces in every line of every file where you've queried my changes to the indentation. And this of course would make it harder during your PR review to spot what I've changed. I suggest I make this change at the end, after I've answered all your non-indentation related comments?

emData/generate_TB.py
aehart: The arguments to a `constexpr` function don't need to be known at compile time, the function just needs to return a constant value after argument substitution, whether that happens at compile time or run time.

But your point is valid, since the function cannot be evaluated at compile time if the arguments aren't known. So I think making the functions `inline` instead makes sense.

TrackletAlgorithm/MemoryTemplateBinnedCM.h
aehart: OK, these indentation comments are the only ones left. The block highlighted here is the most significant example, since the only changes you made, AFAICT, are to the indentation. Some of those changes also don't make sense, like lines 156 and 160 being in different columns now.

I included suggestions for the other blocks I highlighted that you can simply accept if you like, so I hope these comments are fairly trivial to address.

TrackletAlgorithm/MemoryTemplateTPROJ.h
tomalin: I've replaced all tabs by spaces in MemoryTemplate.h MemoryTemplateBinnedCM.h & MemoryTemplateTPROJ.h , and then adjusted the number of spaces, such that it increases by 2, whenever, there is an additional pair of curly brackets.

aehart (approved):
These changes look good to me. I'll go ahead and merge now, and we should discuss the remaining open questions, perhaps in the next TF FW meeting.

PR #349, opened 2024-10-10
Merge/stream synchronization

mcoshiro:
Fixes a few synchronization issues in FPGA 1 project
 - tf_merge_streamer now only read TP_bx_out when it is valid so BX 0 is synchronized with other BXs
 - tf_merge_streamer now outputs its own BX output (TP_bx_out_merged) which is synchronized with its outputs
 - tf_merge_streamer read location is now synchronized with the appropriate valid bit, which fixes occasional strange behavior

It was noticed that some of the output does not match the test vectors, but this seems to be caused by TPs upstream of the merge/stream step.

Associated with PR [#64](https://github.com/cms-L1TK/project_generation_scripts/pull/64) in project_generation_scripts.

### DIFF
--- a/IntegrationTests/common/hdl/tf_merge_streamer.vhd
+++ b/IntegrationTests/common/hdl/tf_merge_streamer.vhd
@@ -25,37 +25,37 @@ use unisim.vcomponents.all;
 use work.tf_pkg.all;
 
 entity tf_merge_streamer is
-	generic (
-		RAM_WIDTH	:	natural := 72;
-		NUM_PAGES   :   natural := 8;
-		RAM_DEPTH 	: natural := NUM_PAGES * PAGE_LENGTH;
-		NUM_INPUTS 		: natural := 4;
-		NUM_EXTRA_BITS: natural := 2;
-		ADDR_WIDTH : natural := 7
-	);
-	port (
-	bx_in : in std_logic_vector(2 downto 0 );
+  generic (
+    RAM_WIDTH:  natural := 72;
+    NUM_PAGES   :   natural := 8;
+    RAM_DEPTH   : natural := NUM_PAGES * PAGE_LENGTH;
+    NUM_INPUTS     : natural := 4;
+    NUM_EXTRA_BITS: natural := 2;
+    ADDR_WIDTH : natural := 7
+  );
+  port (
+  bx_in : in std_logic_vector(2 downto 0 );
   bx_in_vld : in std_logic;
-	rst: in std_logic;
-	clk : in std_logic;
+  rst: in std_logic;
+  clk : in std_logic;
   --output read enable to tf_mem modules
-	enb_arr: out std_logic_vector(NUM_INPUTS-1 downto 0);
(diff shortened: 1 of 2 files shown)


mcoshiro:
Expanding scope of PR: now added additional changes to help FPGA1 project to meet timing. Two main changes are included:

### Changed TPAR memories to use URAMs

This helps with high BRAM usage since currently all the TPARs must be placed near TP_L1L2A as the done signal for all TPARs comes from this single TP. This change makes the baseline project workable, but long term, we will need to change this as additional TPARs for displaced tracking will make the current solution untenable

### Modified `tf_merge_streamer`

Vivado synthesized the previous version of `tf_merge_streamer` in a way that would generate paths with 20-30 logic levels. In particular, Vivado seemed to struggle to figure out that the for loop with exit could be implemented as a small number of LUTs. I wasn't able to figure out a way that both had clean VHDL and which Vivado could figure out, so for now, I changed it to a rather ugly long if-statement, which Vivado seems to handle better. Other suggestions are welcome.

IntegrationTests/common/hdl/tf_merge_streamer.vhd
```diff
@@ -25,113 +25,202 @@ use unisim.vcomponents.all;
...
+            if (readmask(0) = '1') then
+              nextread := 0;
+            elsif (readmask(1) = '1') then
+              nextread := 1;
+            elsif (readmask(2) = '1') then
+              nextread := 2;
+            elsif (readmask(3) = '1') then
+              nextread := 3;
+            end if;
+          end if;
```
jasonfan393, line 178: I think the issue with the for loop is some combo of the variable loop size and the exit. I think something where we just do the loop backwards instead of relying on an exit statement to do the priority encoding will work fine (like below)

Tried this in standalone project, and don't get any >10 logic level paths associated with merge_streamer. The project does not pass timing but  I think this is an issue unrelated to the PR that has appeared with other recent changes. Not sure if this will react differently in EMP, but if it does then we can just leave the big if block and revisit later.
```suggestion
          for i in 0 to 3 loop
            if (readmask((toread(0) - i) mod 4) = '1') then
              nextread := (toread(0) - i) mod 4;
            end if;
          end loop;
```

jasonfan393 (commented):
@mcoshiro looks good, I'd want to simplify the big if block if possible, but if this proves to be too annoying we can always merge this and push it off for later

IntegrationTests/common/hdl/tf_merge_streamer.vhd
mcoshiro: Thanks for looking into this! I'll try building this version in EMP and see how it looks

mcoshiro: Checked in EMP and the reverse for loop doesn't generate >10 logic level paths, so I've changed the priority encoding to the suggestion (and also removed the unused line of code)

jasonfan393 (approved):
Looks good! approved

PR #350, opened 2024-11-05
Unify the VMStubMEMemoryCM and VMStubTEOuterMemoryCM in VMStubMemory class

aryd:
Some cleanup to use common class for ME and TE VM stubs.


IntegrationTests/common/script/synth.tcl
```diff
@@ -23,7 +23,7 @@ set_property -name {STEPS.SYNTH_DESIGN.ARGS.MORE OPTIONS} -value {-mode out_of_c
 
 # Synthesis
 update_compile_order -fileset sources_1
-launch_runs synth_1 -jobs 4
+launch_runs synth_1 -jobs 64
```
aehart, line 26: @aryd This may be excessive on machines with fewer CPU cores or less memory. I would recommend keeping the default at four or, better yet, getting the actual number of CPU cores from `nproc` using the `exec` command in Tcl.

```suggestion
set ncpus [exec nproc]
launch_runs synth_1 -jobs $ncpus
```

IntegrationTests/common/script/impl.tcl
```diff
@@ -20,7 +20,7 @@ reset_run impl_1
 
 # Implementation
 update_compile_order -fileset sources_1
-launch_runs impl_1 -jobs 4
+launch_runs impl_1 -jobs 64
```
aehart, line 23: @aryd Same comment as in synth.tcl.

```suggestion
set ncpus [exec nproc]
launch_runs impl_1 -jobs $ncpus
```

TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -153,7 +151,13 @@ inline void step(const VMStubMECM<VMSMEType> stubmem[4][1<<(kNbitsrzbinMP+kNbits
...
 
   }
-  ap_uint<ProjectionRouterBufferBase<VMProjType, AllProjectionType>::kPRBufferZBinSize -1 + kNBits_MemAddrBinned> slot = (iphi_ + use_[iusetmp].range(0,0)) * nbins + zbin_ + use_[iusetmp].range(1,1);
+  const int nphibins = 3; //number of bits for phi bins in VM memory not that L1 actually only uses 2 bits
+  ap_uint<ProjectionRouterBufferBase<VMProjType, AllProjectionType>::kPRBufferZBinSize -1 + kNBits_MemAddrBinned> slot = iphi_ + use_[iusetmp].range(0,0) + (zbin_ + use_[iusetmp].range(1,1)) * (1 << nphibins);
+
+  //The previous line should be like below after L1 is fixed to use four phi bins in the memories
+  //ap_uint<ProjectionRouterBufferBase<VMProjType, AllProjectionType>::kPRBufferZBinSize -1 + kNBits_MemAddrBinned> slot = iphi_ + use_[iusetmp].range(0,0) + (zbin_ + use_[iusetmp].range(1,1)) * (1 << nbits_vmmeall[LAYER]);
+
+  zbin__ = zbin_ + use_[iusetmp].range(1,1);
```
aehart, line 160: @aryd This line is duplicated.

```suggestion
```

aehart:
@aryd In addition to the inline comments above, please open a corresponding PR in project_generation_scripts. It looks like only two lines were changed in that repo, so it should be easy to merge.

IntegrationTests/common/script/impl.tcl
aryd: Fixed

TrackletAlgorithm/MatchEngineUnit.h
aryd: Fixed

IntegrationTests/common/script/synth.tcl
aryd: Fixed

TestBenches/TrackletProcessor_test.cpp
```diff
@@ -168,144 +166,6 @@ std::cout<<module_name[MODULE_];
...
-        err += compareMemWithFile<TrackletProjectionMemory<BARREL2S> >(tproj_barrel_2s[TP::L6PHIB], fout, ievt,
-                                                       "\nTrackletProjection (L6PHIB)", truncation);
-      else if (tproj_name.find("_L6PHIC") != string::npos)
-        err += compareMemWithFile<TrackletProjectionMemory<BARREL2S> >(tproj_barrel_2s[TP::L6PHIC], fout, ievt,
-                                                       "\nTrackletProjection (L6PHIC)", truncation);
-      else if (tproj_name.find("_L6PHID") != string::npos)
-        err += compareMemWithFile<TrackletProjectionMemory<BARREL2S> >(tproj_barrel_2s[TP::L6PHID], fout, ievt,
-                                                       "\nTrackletProjection (L6PHID)", truncation);
-
-    }
```
aehart, line 308: @aryd I guess you removed these comparisons because the test vectors no longer agree with the HLS. I know we don't care about this on the FW side, but will it be an issue for the future SW emulation (@tomalin)?

TrackletAlgorithm/MatchProcessor.h
```diff
@@ -912,8 +912,9 @@ void MatchCalculator(BXType bx,
                      ap_uint<MC::LUT_matchcut_r_width>& best_delta_r,
                      const AllStubMemory<ASTYPE>* allstub,
                      const AllProjection<APTYPE>& proj,
-                     ap_uint<VMStubMECMBase<VMSMEType>::kVMSMEIDSize> stubid,
-                     FullMatchMemory<FMTYPE> fullmatch[maxFullMatchCopies]
+                     ap_uint<VMStubBase<VMSType>::kVMSIDSize> stubid,
+                     FullMatchMemory<FMTYPE> fullmatch[maxFullMatchCopies],
+		     int istep
```
aehart, line 917: @aryd Didn't notice before, but `istep` is not used anywhere in this function, so I think it can be removed.
```suggestion
                     FullMatchMemory<FMTYPE> fullmatch[maxFullMatchCopies]
```

TestBenches/TrackletProcessor_test.cpp
aehart: BTW, in either case, this question relates more to the legacy emulation in CMSSW, so it need not delay merging this PR.

aehart:
@aryd Now that https://github.com/cms-L1TK/project_generation_scripts/pull/66 has been merged, please update the project_generation_scripts submodule here.

TestBenches/TrackletProcessor_test.cpp
aryd: Though we have the code still in HLS to calculate the projections in the TP modules the (default) configuration does not produce these memories as they are not used by any other module. Hence I removed them in this test...
But as I said the code is still there to calculate the projection if you have the right configuration. It is clear that it is going to be harder and harder to maintain the future SW emulation if we don't migrate to the PC and VMSMER modules...

TrackletAlgorithm/MatchProcessor.h
aryd: Removed (I keep adding this argument each time I debug this code as I like to print out on which clock step something is calculated to compare with the emulation...)

TestBenches/TrackletProcessor_test.cpp
tomalin: The Future SW is not reliant on the HLS test-benches. But I guess if you're no longer validating the content of the TrackletProjectionMemories, they will start to deviate from emulation, so it may become problematic for the Future SW in time. But hopefully, (if Future SW remains alive), it will migrate to 2-FPGA at some point.

aehart (approved):
Looks great. Approved!

PR #351, opened 2024-12-09
Optimize MP for CMSSW CPU resources

bryates:
This PR addresses an issue where CMSSW wastes CPU cycles for emptyUnit and nearFull3Unit. They are not loaded in static classes, and should only be initialized once.

### DIFF
--- a/TrackletAlgorithm/MatchEngineUnit.h
+++ b/TrackletAlgorithm/MatchEngineUnit.h
@@ -282,7 +282,7 @@ inline void step(const VMStub<VMSType> stubmem[4][1<<(kNbitsrzbinMP+kNbitsphibin
 #endif
 
  inline void set_empty() {
-   static const emptyUnitClass<MatchEngineUnitBase<VMProjType>::kNBitsBuffer> EmptyUnitClass;
+   static const EmptyUnitClass<MatchEngineUnitBase<VMProjType>::kNBitsBuffer> EmptyUnitClass;
    empty_ = EmptyUnitClass.lut[(readindex_,writeindex_)];
    //empty_ = emptyUnit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
  }
@@ -293,7 +293,7 @@ inline void step(const VMStub<VMSType> stubmem[4][1<<(kNbitsrzbinMP+kNbitsphibin
  }
  
  inline bool nearFull() {
-   static const nearFull3Class<MatchEngineUnitBase<VMProjType>::kNBitsBuffer> NearFull3Class;
+   static const NearFull3Class<MatchEngineUnitBase<VMProjType>::kNBitsBuffer> NearFull3Class;
    return NearFull3Class.lut[(readindex_,writeindex_)];
    //return nearFull3Unit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
  }
--- a/TrackletAlgorithm/MatchEngineUnit.h
+++ b/TrackletAlgorithm/MatchEngineUnit.h
@@ -282,7 +282,8 @@ inline void step(const VMStub<VMSType> stubmem[4][1<<(kNbitsrzbinMP+kNbitsphibin
 #endif
 
  inline void set_empty() {
-   empty_ = emptyUnit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
+   empty_ = EmptyUnit[(readindex_,writeindex_)];
+   //empty_ = emptyUnit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
  }
 
  inline bool empty() const {
@@ -291,7 +292,8 @@ inline void step(const VMStub<VMSType> stubmem[4][1<<(kNbitsrzbinMP+kNbitsphibin
  }
  
  inline bool nearFull() {
-   return nearFull3Unit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
+   return NearFull[(readindex_,writeindex_)];
+   //return nearFull3Unit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
  }
 
  inline bool idle() {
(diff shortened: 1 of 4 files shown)


TrackletAlgorithm/MatchEngineUnit_parameters.h
```diff
@@ -28,9 +28,28 @@ static bool nearFullUnitBool(ap_uint<kNBitsBuffer> rptr, ap_uint<kNBitsBuffer> w
   return wptr1==rptr || wptr2==rptr;
 }
 
+template<int kNBitsBuffer>
+class nearFull3Class {
```
tomalin, line 32: CMS coding rules state that class names should start with a capital letter, whereas variable and function names should start with a lower case letter.

TrackletAlgorithm/MatchEngineUnit_parameters.h
```diff
@@ -80,6 +99,22 @@ static bool nearFull4UnitBool(ap_uint<kNBitsBuffer> rptr, ap_uint<kNBitsBuffer>
   return wptr1==rptr || wptr2==rptr || wptr3==rptr || wptr4==rptr;
 }
 
+template<int kNBitsBuffer>
+class emptyUnitClass {
```
tomalin, line 103: CMS coding rules state that class names should start with a capital letter, whereas variable and function names should start with a lower case letter.

TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -282,7 +282,9 @@ inline void step(const VMStub<VMSType> stubmem[4][1<<(kNbitsrzbinMP+kNbitsphibin
 #endif
 
  inline void set_empty() {
-   empty_ = emptyUnit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
+   static const emptyUnitClass<MatchEngineUnitBase<VMProjType>::kNBitsBuffer> EmptyUnitClass;
```
tomalin, line 285: CMS coding rules state that class names should start with a capital letter, whereas variable and function names should start with a lower case letter.

tomalin:
This looks good to me, aside from the minor style comments above. We've not been able to test the CPU performance, as recent changes to the HLS L1Trk repo broke the Future SW again, but logically it should fix the problem.

bryates:
> > This looks good to me, aside from the minor style comments above. We've not been able to test the CPU performance, as recent changes to the HLS L1Trk repo broke the Future SW again, but logically it should fix the problem.
> 
> This seems like an improvement, but is there a reason for not simply declaring the existing LUTs as `static`, instead of introducing these wrapper classes? For example, this PR does what I'm suggesting for one of the existing LUTs (not sure why this is only done for one of them):
> 
> https://github.com/cms-L1TK/firmware-hls/blob/4faf3caef4bf6179becddaba2d4960a7194d4a0d/TrackletAlgorithm/MatchEngineUnit_parameters.h#L52
> 
> 
> The only thing missing would be logic to prevent reinitializing the LUT after the first call.
> As it stands, multiple objects of these new wrapper classes can be declared, each of which will initialize the exact same LUT. In fact, I guess this happens, since objects of these classes are declared in `MatchEngineUnit`, of which there are multiple objects declared in `MatchProcessor`. Declaring the existing LUTs as `static` would prevent this.

@tomalin suggested the wrapper class to prevent CMSSW from recomputing the LUTs for every event. Things were set to static before and that didn't seem to be enough.

aehart:
> > > This looks good to me, aside from the minor style comments above. We've not been able to test the CPU performance, as recent changes to the HLS L1Trk repo broke the Future SW again, but logically it should fix the problem.
> > 
> > 
> > This seems like an improvement, but is there a reason for not simply declaring the existing LUTs as `static`, instead of introducing these wrapper classes? For example, this PR does what I'm suggesting for one of the existing LUTs (not sure why this is only done for one of them):
> > https://github.com/cms-L1TK/firmware-hls/blob/4faf3caef4bf6179becddaba2d4960a7194d4a0d/TrackletAlgorithm/MatchEngineUnit_parameters.h#L52
> > 
> > The only thing missing would be logic to prevent reinitializing the LUT after the first call.
> > As it stands, multiple objects of these new wrapper classes can be declared, each of which will initialize the exact same LUT. In fact, I guess this happens, since objects of these classes are declared in `MatchEngineUnit`, of which there are multiple objects declared in `MatchProcessor`. Declaring the existing LUTs as `static` would prevent this.
> 
> @tomalin suggested the wrapper class to prevent CMSSW from recomputing the LUTs for every event. Things were set to static before and that didn't seem to be enough.

I see that the existing LUT functions have `static` return types. However, my understanding is that in C++, this is meaningless, more or less (kind of like having a `const` return type). You have to declare the actual object within the function as `static` in order for this keyword to have an effect, as you do in the line I referenced above.

bryates:
Good point. I had tried so many variations that didn't compile. Let me double check if that works

TrackletAlgorithm/MatchEngineUnit_parameters.h
bryates: Fixed

TrackletAlgorithm/MatchEngineUnit.h
```diff
@@ -291,7 +293,9 @@ inline void step(const VMStub<VMSType> stubmem[4][1<<(kNbitsrzbinMP+kNbitsphibin
  }
  
  inline bool nearFull() {
-   return nearFull3Unit<MatchEngineUnitBase<VMProjType>::kNBitsBuffer>()[(readindex_,writeindex_)];
+   static const nearFull3Class<MatchEngineUnitBase<VMProjType>::kNBitsBuffer> NearFull3Class;
```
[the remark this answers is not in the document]
bryates: Fixed

bryates:
@aehart I've pushed the latest change to remove the old LUT functions and updated some of the code where there were still use in the `MatchProcessor.h` and `ProjectionRouterBufferArray.h`. The timing did not change.

```
CLB:            580
LUT:           2250
FF:            2827
DSP:              2
BRAM:             2
SRL:             13
URAM:             0
#=== Final timing ===
CP required:    4.000
CP achieved post-synthesis:    3.227
CP achieved post-implementation:    3.202
```

aehart (approved):
Thanks, @bryates. This looks good to me.

PR #352, opened 2024-12-10
New track builder and write_mem cleanup

aryd:
This update has the combined TrackBuilder module and cleanup of the write_mem methods.


tomalin:
Function setWriteBX has been added to all three MemoryTemplate*.h classes, but it seems never to be called?

aryd:
This is part of aligning the test benches with how this implemented for the integrated project. In the integrated project we have these memories implemented in vhdl. The ‘framework’ takes care of managing which BX is being processed. So when a (HLS) processing module writes to a memory it does not specify the BX. However, the interface - for historical reasons - included the BX. So in this cleanup I removed this argument. The argument was used in the simulation implementation and I now changed this so that you set the BX that is being written through this setWriteBX method. It is called in all the test benches.

 Hope this helps.

 -Anders


Anders Ryd
***@***.******@***.***>



On Dec 11, 2024, at 8:45 PM, Ian Tomalin ***@***.******@***.***>> wrote:



Function setWriteBX has been added to all three MemoryTemplate*.h classes, but it seems never to be called?

PR #354, opened 2025-01-14
Using static for LUTs

bryates:
This is an alternative to #351 to address an issue where CMSSW wastes CPU cycles for emptyUnit and nearFull3Unit. All luts are now declared as static, so this _should_ fix the issue.

### DIFF
--- a/TrackletAlgorithm/MatchEngineUnit_parameters.h
+++ b/TrackletAlgorithm/MatchEngineUnit_parameters.h
@@ -7,7 +7,7 @@
 
 template<int kNBitsBuffer>
 static const ap_uint<(1 << (2 * kNBitsBuffer))> nearFullUnit() {
-  ap_uint<(1 << (2 * kNBitsBuffer))> lut;
+  static ap_uint<(1 << (2 * kNBitsBuffer))> lut;
   for(int i = 0; i < (1 << (2 * kNBitsBuffer)); ++i) {
 #pragma HLS unroll
     ap_uint<kNBitsBuffer> wptr, rptr;
@@ -30,7 +30,7 @@ static bool nearFullUnitBool(ap_uint<kNBitsBuffer> rptr, ap_uint<kNBitsBuffer> w
 
 template<int kNBitsBuffer>
 static const ap_uint<(1 << (2 * kNBitsBuffer))> nearFull3Unit() {
-  ap_uint<(1 << (2 * kNBitsBuffer))> lut;
+  static ap_uint<(1 << (2 * kNBitsBuffer))> lut;
   for(int i = 0; i < (1 << (2 * kNBitsBuffer)); ++i) {
 #pragma HLS unroll
     ap_uint<kNBitsBuffer> wptr, rptr;
@@ -55,7 +55,7 @@ static bool nearFull3UnitBool(ap_uint<kNBitsBuffer> rptr, ap_uint<kNBitsBuffer> 
 
 template<int kNBitsBuffer>
 static const ap_uint<(1 << (2 * kNBitsBuffer))> nearFull4Unit() {
-  ap_uint<(1 << (2 * kNBitsBuffer))> lut;
+  static ap_uint<(1 << (2 * kNBitsBuffer))> lut;
   for(int i = 0; i < (1 << (2 * kNBitsBuffer)); ++i) {
 #pragma HLS unroll
     ap_uint<kNBitsBuffer> wptr, rptr;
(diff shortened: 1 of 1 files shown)


TrackletAlgorithm/MatchEngineUnit_parameters.h
```diff
@@ -7,7 +7,7 @@
 
 template<int kNBitsBuffer>
 static const ap_uint<(1 << (2 * kNBitsBuffer))> nearFullUnit() {
-  ap_uint<(1 << (2 * kNBitsBuffer))> lut;
+  static ap_uint<(1 << (2 * kNBitsBuffer))> lut;
```
aehart, line 10: @bryates For this to work, I _think_ you also need logic to prevent the reinitialization of the LUT. For example:
```c++
template<int kNBitsBuffer>
static const ap_uint<(1 << (2 * kNBitsBuffer))> nearFullUnit() {
  static ap_uint<(1 << (2 * kNBitsBuffer))> lut;
  static bool initialized = false;
  if (!initialized) {
    initialized = true;
    for(int i = 0; i < (1 << (2 * kNBitsBuffer)); ++i) {
#pragma HLS unroll
      ap_uint<kNBitsBuffer> wptr, rptr;
      ap_uint<2 * kNBitsBuffer> address(i);
      (rptr,wptr) = address;
      auto wptr1 = wptr+1;
      auto wptr2 = wptr+2;
      bool result = wptr1==rptr || wptr2==rptr;
      lut[i] = result;
    }
  }
  return lut;
}
```
Of course, this applies to the other functions in this file.

aehart: @bryates You might also remove the `static const` on the return types of these functions, since they don't really have any effect.

bryates:
> > This seems to improve timing as well (unless I'm behind on the latest numbers)
> > ```
> > #=== Post-Implementation Resource usage ===
> > CLB:            580
> > LUT:           2250
> > FF:            2827
> > DSP:              2
> > BRAM:             2
> > SRL:             13
> > URAM:             0
> > #=== Final timing ===
> > CP required:    4.000
> > CP achieved post-synthesis:    3.227
> > CP achieved post-implementation:    3.202
> > ```
> 
> @bryates After addressing the comment above, could you rerun the out-of-context implementation and also do a comparison of the resource utilization, before your changes vs. after? Just to make sure nothing unexpected is happening during C-synthesis.

These changes reduced the resource utilization a bit. The post-implementation timing is about the same.
```
CLB:            541
LUT:           1794
FF:            2540
DSP:              2
BRAM:             2
SRL:              0
URAM:             0
#=== Final timing ===
CP required:    4.000
CP achieved post-synthesis:    2.900
CP achieved post-implementation:    3.200
Timing met
```

TrackletAlgorithm/MatchEngineUnit_parameters.h
tomalin: > @bryates For this to work, I _think_ you also need logic to prevent the reinitialization of the LUT. For example:
> 
> ```c++
> template<int kNBitsBuffer>
> static const ap_uint<(1 << (2 * kNBitsBuffer))> nearFullUnit() {
>   static ap_uint<(1 << (2 * kNBitsBuffer))> lut;
>   static bool initialized = false;
>   if (!initialized) {
>     initialized = true;
>     for(int i = 0; i < (1 << (2 * kNBitsBuffer)); ++i) {
> #pragma HLS unroll
>       ap_uint<kNBitsBuffer> wptr, rptr;
>       ap_uint<2 * kNBitsBuffer> address(i);
>       (rptr,wptr) = address;
>       auto wptr1 = wptr+1;
>       auto wptr2 = wptr+2;
>       bool result = wptr1==rptr || wptr2==rptr;
>       lut[i] = result;
>     }
>   }
>   return lut;
> }
> ```
> 
> Of course, this applies to the other functions in this file.

The one issue with this solution is that it relies on "static bool". Non const statics are not thread safe, so this would not be useable from the Future CMSSW SW (unless we used pragmas to provide a thread protection for this static variable when run inside CMSSW)

bryates: That's a very good point. @aehart what do you think? I guess I can't say for sure if this bool did anything, but the resource utilization and timing didn't change much. Maybe @tomalin can test this in CMSSW with and without the bool to see if the CPU usage still blows up?

aehart: The bool is key to this solution. Without it, every call to these functions will cause the LUTs to be reinitialized. The `static` keyword alone doesn't magically change that.

In any case, the LUT is also a non-const static. I didn't realize this was a constraint. I'll need to think about an alternative solution.

tomalin: This PR has been stuck for a while. Can't we just close this one and use Brent's other solution https://github.com/cms-L1TK/firmware-hls/pull/351 instead?

bryates: @aehart what do you think? That PR works from my end, but I wasn't sure if you had more comments about it based on the types that I used.

aehart: @bryates I think it's fine to close this PR and go with #351 instead. There are probably further optimizations possible, but #351 should already be an improvement.

I would just ask that in #351 you remove the functions that now have an equivalent implementation as a class and also show that the timing and resource utilization aren't negatively affected by these changes.

bryates:
Closing in favor of #351.

Issue #357, opened 2025-03-04
Make HLS code thread safe


tomalin:
The HLS code must be thread safe to be used from the Future L1Trk CMSSW code. All "static" variables must therefore be eliminated, unless they are declared "const". This affects the following (March 2025):

- MatchProcessor.h
- TrackletProcessor.h
- VMRouterCM_parameters.h
- TrackletProcessor_parameters.h
- DuplicateRemoval.cc

bryates:
Thanks @tomalin. PR https://github.com/cms-L1TK/firmware-hls/pull/358 removes the `static` variables from the MP.

aryd:
I tried to remove the static for the line:

static TEBuffer<Seed,iTC,innerASType,OuterRegion<Seed>()> tebuffer;

in TrackletProcessor.h.  This static should not be needed. Removing it works fine for the csim. However, it fails the cosim step. I did try a few things without success. So for now we will leave this in.

PR #359, opened 2025-03-11
Made all VMRCM static variables const

meisonlikesicecream:
VMRCM static variables are now static const, as requested: https://github.com/cms-L1TK/firmware-hls/issues/357

The timing and resource usage is the same as before this change. However, I noticed that the VMRCM Initiation interval is 2 at the moment? (not sure if it's the old Vivado HLS 2019.2 that I was using, or if something has happened since I last looked at the VMR)

### DIFF
--- a/emData/generate_VMRCM.py
+++ b/emData/generate_VMRCM.py
@@ -191,7 +191,7 @@ def writeParameterFile(vmr_list, mem_dict, output_dir):
         for i in range(numLayers):
             parameter_file.write(
                 "template<> inline const int* getPhiCorrTable<TF::L" + str(i+1) + ">(){\n"
-                "  static int lut[] = \n"
+                "  static const int lut[] = \n"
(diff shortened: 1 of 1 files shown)


aryd (commented):
I think this looks fine.

meisonlikesicecream:
> > VMRCM static variables are now static const, as requested: #357
> > The timing and resource usage is the same as before this change. However, I noticed that the VMRCM Initiation interval is 2 at the moment? (not sure if it's the old Vivado HLS 2019.2 that I was using, or if something has happened since I last looked at the VMR)
> 
> Mei - I think I have fixed the II=2 issue in PR 356. This was due to an issue in the HLS implementation of the VMStub memory.

Aha okay, perfect!

aehart (approved):
Looks good to me as well. I'll approve it, but it just needs rebasing on top of @aryd's changes before it can be merged, @meisonlikesicecream.

meisonlikesicecream:
Ok I've rebased (and indeed the II=1 now), will merge once the CI has finished :-)

PR #361, opened 2025-04-24
Remove unneeded static

aryd:
Removed "Static" that is not needed.


aryd:
We decided to not fix the static usage now and close this PR.

PR #363, opened 2025-04-30
FW/SW synchronization

aehart:
This PR updates the test vectors with those generated from the [fw_synch_250430](https://github.com/cms-L1TK/cmssw/releases/tag/fw_synch_250430) tag in cms-L1TK/cmssw.

The README has also been updated with updated recipes for generating these test vectors.


aehart (commented):
Going ahead with merge as this is a straightforward update of the test vectors.

PR #365, opened 2025-05-21
Preparation for Vitis HLS

aehart:
This PR contains several changes made while migrating to Vitis HLS that are also compatible with Vivado HLS. In particular:
- Some obsolete TQ files were removed. These should have been removed previously when the TQ was moved to another repo.
- Several unused variables have been removed, including all the `nproj` variables in the TP and PC.
- The `complete` keyword has been removed from all `array_partition` pragmas. The syntax of this keyword is slightly different in Vitis HLS, but in any case, complete partitioning is the default, so simply removing it has no effect on the results of C-synthesis.
- When absolute values are calculated, `std::abs` is explicitly used to get rid of a warning in Vitis HLS.
- The two `pipeline` pragmas in the IR have been consolidated.
- A check on the number of disk stubs was added in the TB. This check is redundant, but Vitis HLS requires it for some reason.
- An explicit `inline` pragma was added to `FullMatch::getTrackletID()`. Vivado HLS inlines this method implicitly, but Vitis HLS does not.
- A `latency` pragma was removed from `TPRegionLUT::lookup()` with no effect on the results of C-synthesis for the TP.


aehart (commented):
Going ahead with merge, since these changes should have no effect on results of C-synthesis.

PR #367, opened 2025-06-09
Memory resync

mcoshiro:
Minimal changes to allow resetting/resync'ing of memory modules and pipelining modules in the case the start signal is not initially asserted in time with BX 1. This allows for recovery via soft reset in hardware rather than having to reload firmware if modules are not synchronized. A reset port is added to pipelining modules, while the synchronization logic of memory modules is slightly modified.

Currently based on `fpga2_timing`, but will rebase on master once `fpga2_timing` is merged.

Corresponds to project_generation_scripts PR [71](https://github.com/cms-L1TK/project_generation_scripts/pull/71).


aehart (approved):
Looks good. Please update the project_generation_scripts submodule to point to https://github.com/cms-L1TK/project_generation_scripts/commit/66f76e1ca62e909a6e0fa7c00480225f78594cb8, now that https://github.com/cms-L1TK/project_generation_scripts/pull/71 has been merged. We'll merge this PR once that's done.

PR #370, opened 2025-09-01
FPGA1 debug

aryd:
This PR consists of a set of updates to both project_generation_scripts and firmware_hls to solve several disagreemetns bwtween the reduced FPGA1 project and the CMSSW C++ emulation. Many of the fixes here should be considered "patches" - e.g. the filewriting should be moved into the memory modules to make sure that we accurately print the correct debud information. Now there is a fair bit of manual adjustment to get the debug printout to allign with the memory writing.

 project_generation_scripts:

  - To avoid an overwrite of the InputLink memories I changed them to use 4 pages instead of two. Could probably have used two page after adding delays in the memory writing. This is particularly a problem here as we read multiple input memories in the VMR and the last memory to be read was over written in one of the 100 events.

 - Add delay in the FileWriterFIFO modules to allign debug writing with correct BX. Similarly allign debug printout for VMSTE, IL, and AS

 firmware_hls:

  - Add delay in the FileWriterFIFO.vhd

  - tf_merge_streamer.vhd has some significant rewrites to process data on each BX. The logic is a bit convoluted as the bx_in_vld signal is not alligned with the change in bx. A delay of three clocks has been added to fix this. This should probably be fixed in a cleaner way.

 - Fixes to CompareMemPrintsFW.py to handle MPAR and AS memories. This could be avoided if the debug printout followed the memory writing.

 - Make the IL memory use 4 pages instead of 2.

### DIFF
--- a/IntegrationTests/common/hdl/FileWriterFIFO.vhd
+++ b/IntegrationTests/common/hdl/FileWriterFIFO.vhd
@@ -20,7 +20,9 @@ use work.tf_pkg.all;
 entity FileWriterFIFO is
   generic (
     FILE_NAME  : string;   --! Name of .txt file to be written
-    FIFO_WIDTH  : natural    --! Data width
+    FIFO_WIDTH  : natural;    --! Data width
+    BX_CNT_INIT : integer := 0; --! allows to set offset to allign with BX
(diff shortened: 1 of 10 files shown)


aryd:
Replaced with new PR

PR #371, opened 2025-09-02
FPGA1 debug rebase

aryd:
This PR consists of a set of updates to both project_generation_scripts and firmware_hls to solve several disagreements between the reduced FPGA1 project and the CMSSW C++ emulation. Many of the fixes here should be considered "patches" - e.g. the file writing should be moved into the memory modules to make sure that we accurately print the correct debug information. Now there is a fair bit of manual adjustment to get the debug printout to align with the memory writing.

 project_generation_scripts:

  - To avoid an overwrite of the InputLink memories I changed them to use 4 pages instead of two. Could probably have used two page after adding delays in the memory writing. This is particularly a problem here as we read multiple input memories in the VMR and the last memory to be read was over written in one of the 100 events.

 - Add delay in the FileWriterFIFO modules to align debug writing with correct BX. Similarly align debug printout for VMSTE, IL, and AS

 firmware_hls:

  - Add delay in the FileWriterFIFO.vhd

  - tf_merge_streamer.vhd has some significant rewrites to process data on each BX. The logic is a bit convoluted as the bx_in_vld signal is not aligned with the change in bx. A delay of three clocks has been added to fix this. This should probably be fixed in a cleaner way.

 - Fixes to CompareMemPrintsFW.py to handle MPAR and AS memories. This could be avoided if the debug printout followed the memory writing.

 - Make the IL memory use 4 pages instead of 2.

### DIFF
--- a/IntegrationTests/common/script/CompareMemPrintsFW.py
+++ b/IntegrationTests/common/script/CompareMemPrintsFW.py
@@ -151,7 +151,7 @@ def compare(comparison_filename="", fail_on_error=False, file_location='./', pre
         if verbose: print(data) # Can also just do data.head()
 
         #Need to figure out how to handle the memory "overwrite" - this is a bit of a hack...
-        if (not is_binned) and ("TF_" not in comparison_filename) and ("AS_" not in comparison_filename):
+        if (not is_binned) and ("TF_" not in comparison_filename) and ("AS_" not in comparison_filename) and ("MPAR_" not in comparison_filename):
             rows = []
             for index, row in data.iterrows():
                 if row['ADDR'] == "0x01":
@@ -186,6 +186,16 @@ def compare(comparison_filename="", fail_on_error=False, file_location='./', pre
                     rows.append(index)
             data=data.drop(rows)
 
+        # Hack to remove bin address
+        if "MPAR_" in comparison_filename:
+            for index, row in data.iterrows():
+                adata = row['DATA']
+                adata=adata.replace("0x8", "0x0")
+                adata=adata.replace("0xA", "0x0")
+                adata=adata.replace("0xC", "0x0")
+                adata=adata.replace("0xE", "0x0")
+                data.loc[index, 'DATA'] = adata
+
         # Sort data by ascending address
         if is_binned:
             data.sort_values(by=['BX','ADDR','DATA'], inplace = True)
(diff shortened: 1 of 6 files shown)


TrackletAlgorithm/TrackletProcessor.h
```diff
@@ -358,7 +358,9 @@ ::rinv * const rinv, TrackletParameters::PHI0PAR * const phi0, TP::Types::z0 * c
   const ap_int<TrackletParameters::kTParPhi0Size + 2> phicrit = *phi0 - (*rinv>>8)*ifactor;
   const bool keep = (phicrit > phicritmincut) && (phicrit < phicritmaxcut);
 
-  return valid_rinv && valid_z0 && keep;
+  bool valid_tmax=std::abs(*t)<4000; //FIXME this is a protection but need to implement consistent with C++ emu.
```
aehart, line 361: @aryd Where does the 4000 come from? I'm guessing it's not a magic number in the emulation, so could it be derived from meaningfully named constants here?

I'm also not sure I understand the comment. Does the "FIXME" imply that this should be removed from the emulation and HLS at some point?

IntegrationTests/common/script/CompareMemPrintsFW.py
```diff
@@ -186,6 +186,20 @@ def compare(comparison_filename="", fail_on_error=False, file_location='./', pre
...
+                adata = row['DATA']
+                adata=adata.replace("0x8", "0x0")
+                adata=adata.replace("0x9", "0x1")
+                adata=adata.replace("0xA", "0x0")
+                adata=adata.replace("0xB", "0x1")
+                adata=adata.replace("0xC", "0x0")
+                adata=adata.replace("0xD", "0x1")
+                adata=adata.replace("0xE", "0x0")
+                adata=adata.replace("0xF", "0x1")
+                data.loc[index, 'DATA'] = adata
```
aehart, line 201: @aryd Just so I understand, could this hack be avoided if the printout was done within the memory modules themselves?

aryd: I think so, but maybe this leading valid bit should be added in the C++ emulation instead?

aehart (approved):
I understand the magic 4000 in the TP will eventually be updated when the equivalent selection in the emulation is understood, so this looks good to me for now.

For some reason, the CI tests were canceled on the GitHub side, though they all passed in Gitlab. I am restarting them in GitHub, and will set this PR is automatically merge once they show up as passing.

PR #372, opened 2025-09-11
Synchronization with emulation after 15_1_0_pre4 rebase

aehart:
This PR updates the test vectors to those from the [fw_synch_250903](https://github.com/cms-L1TK/cmssw/releases/tag/fw_synch_250903) tag. The default test vectors use the D98 detector geometry, but URLs for those that use the D110 geometry are also available in emData/download.sh:
https://github.com/cms-L1TK/firmware-hls/blob/f4a3f5499a127127cd93f0d95cf765952ff62e4f/emData/download.sh#L16-L24
However, the C-simulation for PC_L2L3ABCD currently fails when using the D110 test vectors.


TrackletAlgorithm/ProjectionCalculator.h
```diff
@@ -207,7 +207,8 @@ template<
   constexpr int izproj3 = zmean[9];
   const ap_int<14> zproj3 = izproj3;
   projToDisk(zproj3, itinv, rinv, phi0, t, z0, ir_D4, iphi_D4);
-  const bool valid_D4 = ir_D4 >= irmindisk && ir_D4 < irmaxdisk && ((t > itcut) || (t<-itcut));
+  constexpr int iphi_max = (1<<14); //Should have propper constant for 14
```
aehart, line 210: @aryd I think we may already have a constant for the magic 14. Assuming this cut on `iphi_D4` corresponds to the following cut in the emulation:
https://github.com/cms-L1TK/cmssw/blob/e9a64590c22c0f03f55d3894c1f745acc6d103ab/L1Trigger/TrackFindingTracklet/src/ProjectionCalculator.cc#L297
then I think the 14 we need is already here in Constants.h:
https://github.com/cms-L1TK/firmware-hls/blob/fc59320c951a9eacf22c5244595dcfe4187d03d6/TrackletAlgorithm/Constants.h#L32

Also, if I'm interpreting the cuts correctly, I think this would need to be `(1<<14)-1` to exactly match the behavior of the emulation:
```suggestion
  constexpr int iphi_max = (1<<n_phidisk_)-1;
```

Finally, should this cut not be applied to the other disk projections? I don't see anything in the emulation that would make this specific to D4.

aryd:
Andrew,

 thanks for looking into this. I must admit that I just tracked down the problem, but did not look carefully into how to fix this. See comments below.


Anders Ryd
***@***.******@***.***>



On Sep 16, 2025, at 11:10 PM, Andrew Hart ***@***.******@***.***>> wrote:



@aehart commented on this pull request.

________________________________

In TrackletAlgorithm/ProjectionCalculator.h<https://github.com/cms-L1TK/firmware-hls/pull/372#discussion_r2353664915>:

> @@ -207,7 +207,8 @@ template<
   constexpr int izproj3 = zmean[9];
   const ap_int<14> zproj3 = izproj3;
   projToDisk(zproj3, itinv, rinv, phi0, t, z0, ir_D4, iphi_D4);
-  const bool valid_D4 = ir_D4 >= irmindisk && ir_D4 < irmaxdisk && ((t > itcut) || (t<-itcut));
+  constexpr int iphi_max = (1<<14); //Should have propper constant for 14


@aryd<https://github.com/aryd> I think we may already have a constant for the magic 14. Assuming this cut on iphi_D4 corresponds to the following cut in the emulation:
https://github.com/cms-L1TK/cmssw/blob/e9a64590c22c0f03f55d3894c1f745acc6d103ab/L1Trigger/TrackFindingTracklet/src/ProjectionCalculator.cc#L297
then I think the 14 we need is already here in Constants.h:
https://github.com/cms-L1TK/firmware-hls/blob/fc59320c951a9eacf22c5244595dcfe4187d03d6/TrackletAlgorithm/Constants.h#L32

Also, if I'm interpreting the cuts correctly, I think this would need to be (1<<14)-1 to exactly match the behavior of the emulation:

⬇️ Suggested change

-  constexpr int iphi_max = (1<<14); //Should have propper constant for 14
+  constexpr int iphi_max = (1<<n_phidisk_)-1;


I think what you say above is correct. However, when looking at the code in cmssw (https://github.com/cms-L1TK/cmssw/blame/e9a64590c22c0f03f55d3894c1f745acc6d103ab/L1Trigger/TrackFindingTracklet/src/ProjectionCalculator.cc#L297) I’m not sure why it would reject phi values of (1<<n_phidisk_)-1. But if we do what you suggest we will have agreement with the CMSSW simulation.


Finally, should this cut not be applied to the other disk projections? I don't see anything in the emulation that would make this specific to D4.

I was wondering about this also. Clearly we can apply it; but it may be that it does not get triggered in other disks? Adding it will not use much resources; but it seems it was only happening for L2L3 seeds. Again, I have not tried to understand this.

 -Anders

aryd:
Andrew,

 I just run the CMSSW simulation to look into this. I see that we have phi values out of range fairly often for other seeding layers, e.g. L1L2. But these gets rejected as we don’t combine all L1L2 seed into one projection calculator like we do for L2L3. This has the (accidental) effect of rejecting these projections as there is no matching memory to write the projection to if it either over or underflows. Hence, we only have very few cases where we actually get a mismatch between the emulation and simulation.

 But I guess we should add this projection for all disks to avoid any future confusion. Do you want me to make this update?

 -Anders



Anders Ryd
***@***.******@***.***>



On Sep 17, 2025, at 8:54 AM, Anders Ryd ***@***.******@***.***>> wrote:


 Andrew,

 thanks for looking into this. I must admit that I just tracked down the problem, but did not look carefully into how to fix this. See comments below.


Anders Ryd
***@***.******@***.***>



On Sep 16, 2025, at 11:10 PM, Andrew Hart ***@***.******@***.***>> wrote:



@aehart commented on this pull request.

________________________________

In TrackletAlgorithm/ProjectionCalculator.h<https://github.com/cms-L1TK/firmware-hls/pull/372#discussion_r2353664915>:

> @@ -207,7 +207,8 @@ template<
   constexpr int izproj3 = zmean[9];
   const ap_int<14> zproj3 = izproj3;
   projToDisk(zproj3, itinv, rinv, phi0, t, z0, ir_D4, iphi_D4);
-  const bool valid_D4 = ir_D4 >= irmindisk && ir_D4 < irmaxdisk && ((t > itcut) || (t<-itcut));
+  constexpr int iphi_max = (1<<14); //Should have propper constant for 14


@aryd<https://github.com/aryd> I think we may already have a constant for the magic 14. Assuming this cut on iphi_D4 corresponds to the following cut in the emulation:
https://github.com/cms-L1TK/cmssw/blob/e9a64590c22c0f03f55d3894c1f745acc6d103ab/L1Trigger/TrackFindingTracklet/src/ProjectionCalculator.cc#L297
then I think the 14 we need is already here in Constants.h:
https://github.com/cms-L1TK/firmware-hls/blob/fc59320c951a9eacf22c5244595dcfe4187d03d6/TrackletAlgorithm/Constants.h#L32

Also, if I'm interpreting the cuts correctly, I think this would need to be (1<<14)-1 to exactly match the behavior of the emulation:

⬇️ Suggested change

-  constexpr int iphi_max = (1<<14); //Should have propper constant for 14
+  constexpr int iphi_max = (1<<n_phidisk_)-1;


I think what you say above is correct. However, when looking at the code in cmssw (https://github.com/cms-L1TK/cmssw/blame/e9a64590c22c0f03f55d3894c1f745acc6d103ab/L1Trigger/TrackFindingTracklet/src/ProjectionCalculator.cc#L297) I’m not sure why it would reject phi values of (1<<n_phidisk_)-1. But if we do what you suggest we will have agreement with the CMSSW simulation.


Finally, should this cut not be applied to the other disk projections? I don't see anything in the emulation that would make this specific to D4.

I was wondering about this also. Clearly we can apply it; but it may be that it does not get triggered in other disks? Adding it will not use much resources; but it seems it was only happening for L2L3 seeds. Again, I have not tried to understand this.

 -Anders

PR #374, opened 2025-10-09
Preliminary CDC support

aehart:
This PR includes many changes to add preliminary support for the CDCs that are being developed, in particular for FPGA1:

### HLS

- In the HLS, `kMaxProc` has been made into a function that takes in a clock multiplier. The default multiplier is 6, corresponding to a frequency of 240 MHz and 108 cycles per TMUX period. In the future, it will be set to 9 for the InputRouter, corresponding to a frequency of 360 MHz and 162 cycles per TMUX period.

### Core VHDL

- Extra registers have been added in the `clkb` domain for `nent_o`, in order to stabilize this output across a CDC. These are not enabled by default.
- In the pipelining modules, `dont_touch` has been replaced with `keep_hierarchy` so that Vivado is free to optimize away certain modules during implementation, while maintaining the hierarchy for the remaining modules. Similarly, `keep` attributes have been added to internal signals so that they are kept during synthesis, but can be optimized away during implementation.
- A port for BX valid signals has been added to the pipelining modules.
- Parameters have been added to the pipelining modules so that start signals are latched only once, at the beginning of a pipeline.
- The now-obsolete `regceb` ports have been removed from memory modules.

### EMP wrappers

- Extra ports have been added to `payload_f1`, `tf1_wrapper`, and `tf2_wrapper` so that both the 240-MHz and 360-MHz clocks can be plugged into the `SectorProcessor`. Only the 240-MHz clock is currently plugged in though.

### `SectorProcessor` test benches

- A clocking wizard IP has been added, which is used only the test benches to generate the 240-MHz and 360-MHz in such a way that they are guaranteed to be in phase.
- The `DELAY` parameter has been removed from `FileReader` and `FileReaderFIFO`, and replaced with a `LOCKED` port. This is done so that reading starts when the two, in-phase clocks generated in the test benches are locked.
- A `MAX_ENTRIES` parameter has been added to `FileReaderFIFO` and `tf_mem`, so that this can be configured differently for 240 MHz (`MAX_ENTRIES` = 108, the default) and 360 MHz (`MAX_ENTRIES` = 162).

### Constraints for OOC builds

- Constraints for the 360-MHz clock and for the 360→240 MHz CDC have been added for the OOC builds of `SectorProcessor`.
- The floorplan for the OOC build of FPGA1 has been updated.

### Other changes

- The delay argument has been removed from calls to generator_hdl.py in emData/download.sh.
- The project_generation_scripts submodule has been updated to point to the head of the source branch of https://github.com/cms-L1TK/project_generation_scripts/pull/77.


mcoshiro:
I unfortunately didn't get a chance to review the code this week, but I checked that the FPGA2 output in EMP looks okay, and 1/4 FPGA2 EMP implementations met timing, so I think it is okay from the FPGA2 EMP side.

PR #375, opened 2025-10-20
Fix binmask not being reset when sync_nent goes high

lnestor:
This fixes a bug in the `tf_mem_bin` module that can occur when a value is written on the same clock cycle that `sync_nent` goes high. This first occurred when testing with `in_fpga1_2.txt`, where event 19 (the 2nd event in that file) had `TPAR_L1L2F` and `TPAR_L1L2E` signals stored at incorrect memory addresses.

The root cause was the valid bin-mask register (`validbinmasktmp`) was not being completely reset on the first BX transition. Instead, it is possible for one bit to remain set. On a normal BX transition, this is prevented by setting `new_bx=true`. However, this was not done for the first BX transition, when `sync_nent` goes high. This fix addresses that issue.

Also included is a script to run the EMP simulation for FPGA1 over all input files and an update to the latency of the FPGA1 payload. This new latency properly aligns the outgoing metadata and data.

### DIFF
--- a/IntegrationTests/DualFPGA/firmware/scripts/run_sim.tcl
+++ b/IntegrationTests/DualFPGA/firmware/scripts/run_sim.tcl
@@ -1,0 +1,114 @@
+# The directory where the input files to run are found
+set input_dir "../mem"
+
+# The directory to save the simulation output to
+set output_dir "./dataOut"
+
+# The path to the simulation project
+set project_path "../../../../../../proj/vsim_f1/vsim_f1/vsim_f1.xpr"
+
+# The length of the simulation to run (6us is enough to get through 9 events)
+set sim_time 6us
+
+# The list of input files to run over
+set mem_files {
+     "in_fpga1_0.txt"
+     "in_fpga1_1.txt"
+     "in_fpga1_2.txt"
+     "in_fpga1_3.txt"
+     "in_fpga1_4.txt"
+     "in_fpga1_5.txt"
+     "in_fpga1_6.txt"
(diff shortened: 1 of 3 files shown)


aryd (commented):
The bug fix in tf_mem_bin was a case I had missed earlier. I don't have much input on the other changes.

mcoshiro:
Looks good to me. I guess we might eventually want to update the sim script to similarly be able to run 9 events at a time for FPGA2 simulation as well, but we can add this functionality later

IntegrationTests/DualFPGA/firmware/scripts/run_sim.tcl
aehart, line 1: @lnestor Can you say a few words about the use case for this script? I guess it's useful because you've seen different behavior when the simulation begins with certain events?

lnestor: It is the EMP equivalent of the `runSim.tcl` script that exists in the CombinedConfig_FPGAX directories. Those scripts test the SectorProcessor, while this one tests the EMP payload. I used it when testing hardware/simulation agreement.

Additionally, EMP has restrictions on input size, so the 100 input events is split up between 12 files. Running on all those individually can be time consuming because you have to relaunch the simulation every time. This script gets around that by replacing the underlying file the EMP framework reads from, basically making it quick to test all 12 files.

aehart:
Closing, as this PR is contained in #376.

PR #376, opened 2025-11-03
Fix sp2memwriter

mcoshiro:
Fixes some bugs in `sp2_mem_writer.vhd` that caused discrepancies between EMP and standalone simulation/emulation, specifically
 * write addresses made to be only page, as expected by memory modules
 * delay between start, BX, and data signals slightly adjusted

Also added logic to propagate EMP reset to SectorProcessor

Remaining simulation discrepancies seem to just result from extra truncation caused by EMP input (this will be changed when links go to 360 MHz anyway).

These changes were added on top of PR #375.


aehart (approved):
Thanks, @mcoshiro and @lnestor. Looks good. Approved!

